PR #105, opened 2014-07-20
Add "subTypes" support in swagger models, model signature now contains all inherited properties

garryyao:
The patch implements simple singular inheritance with prototype, where model will simply pull in all inherited properties from ancestor model, which seems to be enough to support the spec:
https://github.com/wordnik/swagger-spec/blob/master/versions/1.2.md#modelSubTypes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ capybara-*.html
 /spec/tmp/*
 **.orig
 *.idea
+node_modules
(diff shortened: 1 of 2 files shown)


whitlockjc:
Why not just write unit tests?  There are tests for all things that swagger-ui uses.

whitlockjc:
The pull request is not mergeable and there are no tests for this.  Is there any way to get this updated?

webron:
@garryyao - I realize it's been a long time, but would you be willing to update the PR to the current code base?

fehguy:
I have to close this out because of code drift.  Sorry @garryyao for lagging on getting it merged.

PR #110, opened 2014-08-13
Fix: treat child models as referenced

Larry0ua:
This commit shows model subtypes as referenced models in the swagger ui. Things missing: discriminator fields should be added here also, discriminator values should be shown for each child model. This is not possible yet as swagger spec does not have values information.

Does not break existing behavior if there are no subtypes in the model list.


fehguy:
subtypes will be in the next milestone.

whitlockjc:
The pull request is not mergeable and there are no tests for this. Is there any way to get this updated?

webron:
@Larry0ua - I realize it's been a long time, but would you be willing to update the PR to the current code base?

fehguy:
There is too much drift to get this in.  Sorry @Larry0ua for lagging on it.

PR #121, opened 2014-09-07
Better @Path annotation regex handling

s4nch3z:
Fixed hadling {} regex quantifiers

Signed-off-by: Alexander Prokopyev xana4ok@gmail.com


fehguy:
please add tests to the test directory with any regression.  I want as much coverage as possible.

PR #128, opened 2014-09-17
fix authorizations

aurelienlt:
Pull request for the issue https://github.com/wordnik/swagger-js/issues/127

* Calls to the authorizations are only made when requesting an operation, and not when requesting the resources list, or any resource (according to the RFC)
* Operations without any `authorizations` field inherits from the `authorizations` field of its resource
* Empty object `{}` as `authorizations` field means no authorization required, and is the default value for a resource
* All the authorizations of the `authorizations` field are required

### DIFF
--- a/lib/swagger.js
+++ b/lib/swagger.js
@@ -161,8 +161,6 @@ SwaggerApi.prototype.build = function() {
       }
     }
   };
-  var e = (typeof window !== 'undefined' ? window : exports);
-  e.authorizations.apply(obj);
   new SwaggerHttp().execute(obj);
   return this;
 };
@@ -340,6 +338,7 @@ var SwaggerResource = function(resourceObj, api) {
   produces = (this.produces | []);
   this.path = this.api.resourcePath != null ? this.api.resourcePath : resourceObj.path;
   this.description = resourceObj.description;
+  this.authorizations = (resourceObj.authorizations || {});
 
   var parts = this.path.split("/");
   this.name = parts[parts.length - 1].replace('.{format}', '');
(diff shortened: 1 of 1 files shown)


lib/swagger.js
```diff
@@ -384,8 +383,6 @@ var SwaggerResource = function(resourceObj, api) {
         }
       }
     };
-    var e = typeof window !== 'undefined' ? window : exports;
-    e.authorizations.apply(obj);
```
fehguy: I believe that if you remove this line, the authorizations won't be applied when fetching an individual api declaration

lib/swagger.js
```diff
@@ -162,8 +162,6 @@ SwaggerApi.prototype.build = function() {
       }
     }
   };
-  var e = (typeof window !== 'undefined' ? window : exports);
-  e.authorizations.apply(obj);
```
fehguy: I believe that if you remove this line, the authorizations won't be applied when fetching an individual api declaration

lib/swagger.js
```diff
@@ -1549,29 +1546,18 @@ SwaggerAuthorizations.prototype.remove = function(name) {
...
+
+  for(name in authorizations) {
+    for(key in this.authz) {
+      if(key == name) {
+        value = this.authz[key];
+        result = value.apply(obj, authorizations);
+        if (result === false)
+          status = false;
+      }
+    }
```
fehguy: This change will introduce a different behavior, namely, _all_ authorizations should be applied if the object does not specify an authorization requirement (line 1556).  If we remove lines 1556 to 1563, we will lose this logic, and only apply authorizations where they're matching by name

fehguy:
I'd like to get this merged, but will need to spend some more time on it based on the comments I made.  In addition, the testing framework for swagger-js has been massively upgraded, and if you can please help provide tests to demonstrate the fixes, it'll help out a ton.

fehguy:
this requires removing globals, which needs more work in swagger-ui.  moving to m2...

Issue #168, opened 2014-11-27
body param called body will error if swagger definition is not body


brodyd:
Prior to the fix for #166 the body parameter had to be called body regardless of the name in the swagger definition.  After the fix for #166 the name of the body param has to match the name given to it in the swagger UI,

This is a breaking change.  We could make a very small change to allow both methods to work properly.

PR #169, opened 2014-11-27
Allow a body alias of body

brodyd:
Allow the body of a swagger call to be aliased to body if it is not using the correct swagger defined name.  This allows a default of body as well as the swagger name for the body. Resolves swagger-api/swagger-js#168


brodyd:
This has been fixed with another change sop this request is not long valid

Issue #181, opened 2014-12-24
[develop 2.0] replace cake with modern build


fehguy:
The library is using cake for building, but may be better off using grunt or gulp.  Not trying to make upgrades just for the same of upgrading, but cake isn't used as often and being coffeescript based, moving away from coffeescript (from #180) it makes sense to move on here as well.

fehguy:
replaced with `gulp build`

Issue #178, opened 2014-12-16
Why use globals?


MattiSG:
This library seems to be relying on a large amount of globals to function. This is surprising, as leaking state-holding variables in the global namespace is most often considered [bad practice](http://www.w3.org/wiki/JavaScript_best_practices#Avoid_globals), for it leads to potential conflicts with other libraries and decreased maintainability.

However, this practice seems absolutely conscious and voluntary in this project, as these variables are prefixed with `window.` to expose them globally on the browser (potentially through an intermediate variable for Node compatibility).

As this practice makes me extremely wary of including the library as it is, I would like to know what is the rationale for exposing so many globals (and requesting clients to do so too).

MattiSG:
@fehguy Does adding the “Feature” tag mean removing globals is an expected feature?

fehguy:
Yes, we are looking at moving them out in the 2.1 release

whitlockjc:
https://github.com/swagger-api/swagger-js/pull/279 should remove all globals other than `SwaggerClient` in the browser.  `authorizations` is now an instance variable for `SwaggerClient` named `clientAuthorizations`.  `models` is now an instance variable for `SwaggerClient` named `models`.  The `addModel` function was removed as it was only for testing and the following constructors were removed because they too were only for testing: `Model`, `Operation` and `Resolver`.

The following constructors are exposed as properties on the `SwaggerClient` constructor and can be used as they were before: `ApiKeyAuthorization`, `CookieAuthorization` and `PasswordAuthorization`.  `SwaggerApi` and `SwaggerClient` are also exposed as properties on the `SwaggerClient` constructor for backward compatibility.  Usage of these constructors will result in a deprecation warning.

MattiSG:
Great!! Will test ASAP.

whitlockjc:
I just merged #279 so the only global now is `SwaggerClient`.

PR #196, opened 2015-01-01
[develop 2.0] Use Gulp as build tool

noirbizarre:
This pull-request add gulp as a build tool (should address #181).

It's a start, but to be ideal, I think the following is needed:
- [ ] split each class in its own file
- [ ] make use of require for dependencies
- [ ] no more globals (see #178 )
- [ ] use npm to manage external dependencies (shred, jquery..)
- [ ] use browerify or webpack for portable browser build
- [ ] produce a bower version in the build (https://www.npmjs.com/package/gulp-bower)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,5 @@ capybara-*.html
 *.idea
 *.iml
 npm-debug.log
+/node_modules/
+/dist/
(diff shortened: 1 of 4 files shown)


fehguy:
Thanks for this--I'll review today

MattiSG:
How's this going?  :)

PR #211, opened 2015-01-24
Make compatible with browserify

### LINKED ISSUE #208: Empty object when browserified
I have a simple wrapper module that starts with:

``` javascript
var EventEmitter = require('events').EventEmitter;
var util = require('util');
var swagger = require('swagger-client');

console.log(swagger) // { } in browser, { ... normal stuff ... } in repl

// Bunch of stuff below
```

As mentioned by the log comment, I get different values in REPL and the browser. I can use it in the REPL but not in the browser. Should this client not work with browserify?

nhagen:
Attempts to resolve #208. Pull request still needs some work (and lots of cleaning), but I want to get feedback on how much this breaks.

This makes significant changes to the export/globals pattern, and includes new standalone files for the library

How can I test this on legacy IE from a mac?


fehguy:
Hi sorry for the delay getting to this--I'm trying to get M1 released and will not be able to verify these changes by that time.

One thing I'll mention is that we have added mocha tests for lots of this, and they are fairly comprehensive.  All in the develop_2.0 branch

So thanks again for the contrib, will be getting back to this PR after the weekend.

jackphel:
This looks really sweet, @nhagen

whitlockjc:
Any progress with updating this to work with the `develop_2.0` branch so we could potentially get the merge passing again?

whitlockjc:
I am going to close this but I will leave #208 open.  My plan is to use #208 as a way to clean the project up some and to get more familiar with things.  My reason for closing this is more about timing, since I think other issues that are in my queue will be easier once this is done, and is not about the quality of what you've done.

PR #216, opened 2015-01-28
Fix jQuery HTTP

noirbizarre:
This PR fix useJQuery parameter broken (SwaggerJS try to use Shred sometimes even if useJQuery is specified).
The ``useJquery`` parameter wasn't given to operations and SwaggerHTTP wasn't able to find it (always undefined) in execute (https://github.com/swagger-api/swagger-js/blob/develop_2.0/src/js/swagger.js#L673)


fehguy:
thank you

noirbizarre:
Waow ! Faster than lightning, Travis doesn't event finish to build !!
Thanks for the merge !

PR #220, opened 2015-02-02
Merged from develop_2.0

fehguy:


fehguy:
this isn't needed anymore.

PR #227, opened 2015-02-05
Fix/dont encode header values

duncanhall:


duncanhall:
Whoops, sorry, should've branched from here, will resubmit this (again!)

PR #226, opened 2015-02-05
Don't encode header values

duncanhall:
Fix for https://github.com/swagger-api/swagger-js/issues/225

PR #228, opened 2015-02-05
Avoid URL encoding header values

duncanhall:
Fix for https://github.com/swagger-api/swagger-js/pull/226

Resubmitted after branching from the wrong place


fehguy:
If you look at the src structure and gulp tasks, you'll see that swagger-client is built from files under src/js. When running gulp build, your changes will be lost. Please modify the src contents

duncanhall:
@fehguy I'm not sure I follow? The changes were made to `src/js/swagger-a.js`. The `swagger-client` file has changes because I built and tested and then committed - and the lib folder is not `.gitignored`.

fehguy:
My bad, scrolled past that part of the commit. Sorry!

duncanhall:
Ha, no worries. Thanks!

PR #236, opened 2015-02-11
regex pattern matching fix

statueofmike:
Expanding on issue #121.

This supports regular expressions in path parameters.

For example: `@Path( "/{first : (?i)[\\w]{2}[\\d]{2}}5/{second : (?i)[\\d]{4}}" )`

will allow calls to `basepath/ab125/1234`

but will reject calls to `basepath/abcd5/1234`, `basepath/ab12/1234`, `basepath/ab125/12345`, or others that don't fit the regex pattern.

### DIFF
--- a/src/js/swagger-compat.js
+++ b/src/js/swagger-compat.js
@@ -792,7 +792,7 @@ SwaggerOperation.prototype.urlify = function (args) {
     if (param.paramType === 'path') {
       if (typeof args[param.name] !== 'undefined') {
         // apply path params and remove from args
-        var reg = new RegExp('\\{\\s*?' + param.name + '.*?\\}(?=\\s*?(\\/?|$))', 'gi');
+        var reg = new RegExp('\\{\\s*?' + param.name + '[^\\{\\}\\/]*(?:\\{.*?\\}[^\\{\\}\\/]*)*\\}(?=(\\/?|$))', 'gi');
         url = url.replace(reg, this.encodePathParam(args[param.name]));
         delete args[param.name];
       }


statueofmike:
Sorry I didn't add tests. I just don't understand the test mock input well enough.

fehguy:
Hi, I'm trying to write some tests around this.  Can you please share a couple things:

* In your swagger JSON, what will you see in the path value?  What you're showing is coming across as such:

```java
@Path( "/{first : (?i)[\\w]{2}[\\d]{2}}5/{second : (?i)[\\d]{4}}" )
```
```
"path": "/pet/{first : (?i)[\w]{2}[\d]{2}}5/{second : (?i)[\d]{4}}"
```

But what are the substitution values?  What are you try ing to replace?

statueofmike:
> In your swagger JSON, what will you see in the path value?

You will see something like this:
```
{"apiVersion":"1.0.0","swaggerVersion":"1.2","basePath":"/api/v1","resourcePath":"/devices","apis":[{"path":"/pet","operations": ....
```

Then further down in the swagger JSON, there will be a 
```
"parameters":[
{"name":"first","required":true,"type":"string","paramType":"path","allowMultiple":false},
{"name":"second","required":true,"type":"string","paramType":"path","allowMultiple":false}
]
```

> What are the substitution values? What are you trying to replace?

My fix is attempting to match path patterns that have `{curly braces}` inside them, which are used in regex quantifiers. The bug I'm addressing is that the current matcher will stop when it finds the first `}` ending curly brace. 

The example `@Path( "/{first : (?i)[\\w]{2}[\\d]{2}}5/{second : (?i)[\\d]{4}}" )` was contrived to be complicated, so I knew the matcher was flexible. Here's a simpler example that I'm using right now:

My basepath is `/api/v1`. My resource has annotation `@Path("/object")`. My handler has the annotation `@Path( "/{uuid : [a-fA-F\\d}{32}}" )`.

So I could substitute in any 32-character hex string as a path parameter here.
For example: if my resource handler has signature `handle( @PathParam( "uuid" ) String uuid )`, a call to "hostname/api/v1/object/abcdef12345678900000000000000000" will replace the "uuid" Java parameter with "abcdef12345678900000000000000000".

PR #250, opened 2015-02-20
Fixed check for absolute URL

chkal:
Comparing the `indexOf()` result against `-1` just tells you whether `http` is somewhere in the URL. So an basepath like `/foo/http/bar` is currently treated as absolute.


fehguy:
Thanks--if you could, please send the PR to the corresponding src folder in the develop_2.0 branch and I'll merge it.

chkal:
Sure, see #253

fehguy:
Thanks, closing this pr

PR #260, opened 2015-02-26
Fix '&' adding to url if param is empty

kojoru:
This is a fix to https://github.com/swagger-api/swagger-ui/issues/959


kojoru:
I'm not sure how to go about testing the commit. There are no apis in petstore which offer 2+ optional query params. Should I add some new operations to sample API?

fehguy:
Take a look at the test/operation.js and you should see how to mock your own operations and test the generated query string

kojoru:
Can't do that since that's a 1.x-only bug.

kojoru:
I can't access SwaggerOperation from swagger-compat.js, can I?

kojoru:
I did eventually write tests with adding new operation to the 1.2 spec. Please let me know if there's a better way.

fehguy:
Yes I didn't notice this was for 1.2. For now yes, adding an operation is the best way

kojoru:
Ok, let me know if there's anything I need to do for it to be merged.

PR #261, opened 2015-02-26
Support inline model properties (properties of type 'object')

### LINKED ISSUE #230: enable map support
for the 2.0 spec support, need to add map support for model samples, data types, etc.

whitlockjc:
This commit ensures `Model.createJSONSample` and `Model.getMockSignature`, and due to the location of the fix `Property.toString`, all support inline models.  _(An inline model is a model property whose `type` is `object`)_

Fixes #230


whitlockjc:
Wow...let me fix this.  Somehow the commit history got jacked up.

Issue #126, opened 2014-09-15
Feature request : registering response interceptors


laurentperez:
Hi
I'd like to feature request the ability to register a response interceptor, much like you can register request "interceptors" by populating window.authorizations with custom parameters.
I know this can somehow break the stateless nature of a call, given that I could reuse parts of the response to alter future requests. But it'd be the sole caller responsibility.
My use case is : I call a login service, it gives me a security token, and I have to use the token for other api requests.
I hacked swagger.js to retain the response object in localStorage, but it's ugly.
Objections ?

fehguy:
hi, i think this is a good feature but we need to be careful about both the signature and nature of the calls.  For instance, it's fine if it's a synchronous interceptor but doing something that chains the response in a callback will certainly break the clients.

I'd like to push this into a future release but keep the feature request open so we can work through first the ideal behavior, and next the signature for the interceptor.

whitlockjc:
I think we should work on this when we [make SwaggerHttp pluggable](https://github.com/swagger-api/swagger-js/issues/239).

fehguy:
now that [SwaggerHttp is pluggable](https://github.com/swagger-api/swagger-js/pull/428), this poses another challenge.  Should the interceptor be handled in the pluggable client?  Looking into this more.

PR #263, opened 2015-02-27
add preResolve callback

wodka:
I had the case where I needed to modify the data of the server response -> to do that I added the config var preResolve that would be called.

(I use it to replace the .host value of the response)


fehguy:
Thanks, I want to support the notion of request and response interceptors but am not sure if `options` is the right place for it.  Off the top of my head, we should follow a pattern like the authorizations, where we expose a mechanism to add response interceptors in the library and add to them as needed.

fehguy:
I will get this implemented in https://github.com/swagger-api/swagger-js/issues/126

Issue #188, opened 2014-12-29
[develop 2.0] support allOf in models


fehguy:
Currently the 2.0 spec supports `allOf` constructs but the js library does not.

https://github.com/swagger-api/swagger-ui/issues/723

webron:
Does this include the `discriminator` property for inheritance or just the `allOf` for composition?

fehguy:
the intent of this issue is to support the spec, for purposes of composition which means adding the discriminator.

webron:
Technically, you can use `allOf` without a discriminator just for composition, which implies a model that includes the properties of another, but there's no direct hierarchy between them. 

Though if this issue includes both, that's great.

DavidBiesack:
yes, we'd like to use it for composition, to reuse schema instead of copy/paste (and to model property inheritance).

wodka:
I just gave it a try in https://github.com/swagger-api/swagger-js/pull/264 - this seems to be working:

```yaml
definitions:
    AbstractObject:
        properties:
            id:
                type: string
    MyObject:
        allOf:
            -
                $ref: "#/definitions/AbstractObject"
            -
                type: object
                required:
                    - name
                properties:
                    name:
                        type: string
    ListOfMyObject:
        type: array
        items:
            $ref: "#/definitions/MyObject"
```

Only errors I see so far is that the Model is not shown in ListOfMyObject - but the ModelSchema is shown correctly.
Also the datatype only reads "undefined"...

webron:
I wonder how we would end up displaying it, as technically this is not a single model but a set of models to be checked against.

whitlockjc:
How do you think this should be displayed?  A lot of work recently has been done to properly support arrays, inline schemas and references but to fix this we need to know just how to display ancestors/included models.

m8rge:
@whitlockjc, I think, we just need to get all schemas from `allOf` property and merge them all together into result model.

webron:
@m8rge - unfortunately, `allOf` is not really a merge, and the problem will come when there are conflicting schema.

The solution should probably be closer to an array of schema possibilities.

m8rge:
@webron This is spec-related question: https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#composition-and-inheritance-polymorphism

webron:
What's the question?

m8rge:
@webron 
> the problem will come when there are conflicting schema
[...]

PR #264, opened 2015-03-03
add ability to parse allOf

wodka:
add ability to parse allOf in definitions
https://github.com/swagger-api/swagger-js/issues/188 -> I do not know if this fixes all cases (my usecase works this way)


lib/swagger-client.js
```diff
@@ -2689,10 +2708,10 @@ SwaggerOperation.prototype.urlify = function (args) {
           addition += encodeURIComponent(param.name) + '=' + encodeURIComponent(args[param.name]);
         } else {
           if (param.required)
-            throw "" + param.name + " is a required query param.";
+            throw '' + param.name + ' is a required query param.';
```
mohsen1: Please `throw new Error('string')`

lib/swagger-client.js
```diff
@@ -2674,8 +2694,7 @@ SwaggerOperation.prototype.urlify = function (args) {
     param = params[i];
     if(param.paramType === 'query') {
       var addition = '';
-      if (Array.isArray(param)) {
-        var j;   
+      if (Array.isArray(param)) {
```
mohsen1: Please use `_.isArray`. Array.isArray is not supported in all browsers we support

src/js/resolver.js
```diff
@@ -14,9 +14,28 @@ Resolver.prototype.resolve = function(spec, callback, scope) {
...
   for(name in spec.definitions) {
     var model = spec.definitions[name];
-    for(propertyName in model.properties) {
-      property = model.properties[propertyName];
-      this.resolveTo(property, resolutionTable);
+    if (!model.properties) {
+      model.properties = {};
+    }
+    if (model.allOf) {
+      for(mergeName in model.allOf){
```
mohsen1: Instead of `for(mergeName in model.allOf){  merge = model.allOf[mergeName];` use underscore: `_.values(model.allOf)`

`for..in` is not safe to use. It might iterate in prototype of the object too. 
in line 15 we use `for..in` again. I wish we could fix that too.

src/js/resolver.js
```diff
@@ -14,9 +14,28 @@ Resolver.prototype.resolve = function(spec, callback, scope) {
...
+        }else{
+          for(propertyName in merge.properties) {
+            model.properties[propertyName] = property = merge.properties[propertyName];
+            console.log(property);
+            this.resolveTo(property, resolutionTable);
+          }
+        }
+      }
+    }else {
+      for (propertyName in model.properties) {
```
mohsen1: another `for..in`

lib/swagger-client.js
wodka: it was introduced in 209019f1390f1652926a9446398244ccef2f7976, and not part of my change, if you want I can change it

lib/swagger-client.js
wodka: this was introduced in bc8f0f68b5de91b8fc26d2433909be010d2b61a7, again not part of my change but I can change it if you want.

src/js/resolver.js
wodka: this would introduce a dependency on underscorejs - it is not yet a requirement for swagger-js and I'm not sure if I should really add it

mohsen1: @wodka I misread your diff. Sorry about that. You don't have to change the code you didn't write but if you refactor those parts it would be nice.

wodka:
I just changed all for..in loops to the underscorejs equivalent

-> the test will fail on those since I cannot figure out right now how to add the underscorejs dependency. @mohsen1 can you help me with this?

src/js/resolver.js
```diff
@@ -5,61 +5,70 @@ var Resolver = function (){};
...
-  }
+    if (model.allOf) {
+      _.each(model.allOf, function(merge){
+        if(merge.$ref){
+          model.$ref = merge.$ref;
+          this.resolveInline(spec, model, resolutionTable, unresolvedRefs);
+        }else{
+          _.each(merge.properties, function(property, propertyName){
+            model.properties[propertyName] = property = merge.properties[propertyName];
+            console.log(property);
```
wodka: will remove that in the next commit...

mohsen1:
@wodka @whitlockjc is adding lodash to this project. Please coordinate with him

whitlockjc:
This needs a rebase.

fehguy:
Hi, this has drifted enough that I'm going to close it.  Apologies.

PR #269, opened 2015-03-04
Fixed exceptions when spec has null http operations

csuich2:
Fix for issue #268

PR #271, opened 2015-03-04
Fixed exceptions when spec has null http operations

csuich2:
See https://github.com/swagger-api/swagger-js/pull/269


whitlockjc:
It seems there are test failures.  If you can get these resolved, we should be in good shape.

csuich2:
Yep - they are passing locally so I am trying to figure out what the difference is. =)

csuich2:
@whitlockjc I'm quite new to a lot of the tools here and seem to be stumped. Both of those tests are passing on my machine and the errors seem like the test is being executed without the first of the commits being in place (since the first commit fixed the issue and the second added tests).

Am I setting up the pull request wrong or is Travis-CI only testing with one of my commits and not both?

csuich2:
Or could it be that I am not checking in swagger-client.js? Should I be committing that after I build?

whitlockjc:
I'm looking into how the test suite works and it is currently brittle.  Can you update your commit to include the changes to `lib/swagger-client*.js` and see if that fixes things?

csuich2:
Done. Looks like Travis-CI hasn't kicked off a new test yet.

csuich2:
There we go. I initially tried to figure out whether I should commit the lib/* changes, but it looked like some commits didn't include that and I wanted to err on the side of caution.

fehguy:
this should be covered already in the refactoring.

PR #275, opened 2015-03-05
Made test task depend on build

aurelian:
Since tests are executed against the distribution build, it makes sense to have the gulp test task depend on build. Now ```gulp watch``` is executing the tests without a rebuild.


whitlockjc:
I've done this locally already and I'd rather not have conflicts.

PR #279, opened 2015-03-07
Support Browserify and related refactoring

### LINKED ISSUE #187: Timeout when submitting causing js errors
See https://github.com/strongloop/loopback-explorer/issues/76#issuecomment-64902913

### LINKED ISSUE #208: Empty object when browserified

### LINKED ISSUE #267: switch from shred to super agent?
Seems like the shred version we're on is broken for some important use cases, and their API has changed considerably.  Can we move to super agent instead?  @whitlockjc might be able to assess.

I would simply add it to the `swaggerHttp.js` class and should be 100% transparent

whitlockjc:
Unfortunately, supporting Browserify required a lot of refactoring.
Below is a list of the changes required to support Browserify:

* Source restructuring so that sources now are Node.js modules and
  Browserify use used to build the browser binaries
* NPM was cleaned up so that only the required files are packaged for
  NPM
* Gulp file was restructured to build with Browserify
* Removed all globals and unnecessary objects/functions from the exports
* Removed shred and replaced with Superagent

Fixes #187
Fixes #208
Fixes #267

Others are possibly fixed but I'd rather add tests to ensure they are.


whitlockjc:
@fehguy gave me the +1 to merge since all tests pass and preliminary testing with swagger-ui was fine.  Since we did remove Shred and switched to Superagent I am aware that some things could have changed that will require some follow-up work.

PR #280, opened 2015-03-09
Properly encode slashes in path parameters

gwicke:
Swagger spec version 2.0 only supports level 1 URI templating as described in
http://tools.ietf.org/html/rfc6570#section-3.2.2. This means that the entire
path component (including slashes) needs to be encoded with
encodeURIComponent.

### DIFF
--- a/lib/swagger-client.js
+++ b/lib/swagger-client.js
@@ -9,7 +9,7 @@ var ArrayModel = function(definition) {
   this.name = "arrayModel";
   this.definition = definition || {};
   this.properties = [];
-  
+
   var requiredFields = definition.enum || [];
   var innerType = definition.items;
   if(innerType) {
@@ -301,7 +301,7 @@ PrimitiveModel.prototype.getMockSignature = function(modelsToIgnore) {
   }
   return returnVal;
 };
-/** 
+/**
  * Resolves a spec's remote references
  */
 var Resolver = function (){};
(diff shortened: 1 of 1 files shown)


lib/swagger-client.js
```diff
@@ -1460,22 +1460,8 @@ Operation.prototype.encodeQueryParam = function(arg) {
   return encodeURIComponent(arg);
 };
 
-/**
- * TODO revisit, might not want to leave '/'
- **/
 Operation.prototype.encodePathParam = function(pathParam) {
-  var encParts, part, parts, i, len;
```
mohsen1: Replying [this](https://github.com/swagger-api/swagger-ui/pull/1015#discussion_r26013542). 

```js
return pathParam.toString().split('/').map(encodeURIComponent).join('/')
```

The above code does exactly the same thing

@fehguy should we merge this change?

gwicke: It does the same thing as the old code, but not the same thing as

```javascript
encodeURIComponent(pathParam);
```

mohsen1: That's true. 

```
encodeURIComponent('a/b') === "a%2Fb"
'a/b'.split('/').map(encodeURIComponent).join('/') === 'a/b'
```
if we decided to have an option to allow slash(`/`), we would need to skip encoding it. Your code skip encoding slash. My suggestion is to make that code shorter and less complex.

Based on @fehguy's comment we need to decide if we want to skip slash in encoding and if the answer to that is yes, we should make this behavior configurable.

gwicke: @mohsen1, according to the [URI template spec](http://tools.ietf.org/html/rfc6570):

- slashes need to be encoded for `{foo}` (the only variant supported in Swagger 2.0), and
- slashes don't need to encoded for `{+foo}`.

So both can be supported easily and within the same URL once the spec gains support for level 2 URI templating.

fehguy:
I believe the spec does not support nor prohibit encoding slashes in path params.  Let me get input from @webron on this and if it is supported, I'll clean and merge this.

webron:
We don't support non-encoded slashes in path parameter values. However, I don't see a problem if they are indeed encoded.

Please note that this PR as against `master ` and should not be merged.

gaillard:
The UI still seems to have this not encoding, this would seem like a pretty silly thing to not encode values the user enters in the boxes for path parameter. Is it planned to change this?

webron:
Which branch have you tested, @gaillard?

gaillard:
Hey @webron, I tried the develop_2.0 branch for swagger-ui with the same effect.

webron:
@gaillard - please open two separate issues on the above then, we won't be able to keep track of those as comments on a closed PR.

gaillard:
will do thanks

PR #288, opened 2015-03-13
Fixed memory leak

sobolk:
This code is causing memory leak in long running app on nodejs. We can't append logs to static array ...


webron:
@sobolk - PRs should be submitted against the develop_2.0 branch.

webron:
Closed in favor of #289.

PR #302, opened 2015-03-19
Bug fix to display comments for array of primitives for properties or fields.

aphu:


whitlockjc:
This pull request is made against an old version of the `develop_2.0` branch.  Any way to get this updated?

lib/swagger-client.js
```diff
@@ -1555,7 +1555,7 @@ var Property = function(name, obj, required) {
   if(obj.$ref)
     this.$ref = simpleRef(obj.$ref);
   else if (obj.type === 'array' && obj.items) {
-    if(obj.items.$ref)
+    if(obj.items)
       this.$ref = simpleRef(obj.items.$ref);
     else
       obj = obj.items;
```
Smolations: This seems like a bad change to me. This conditional will _never_ hit the else statement because it's already wrapped in a conditional that checks for the existence of `obj.items`. The original code seems appropriate, because then the entire block reads:

> Does `obj.items` contain a $ref? If it does, then retain the resolved $ref to `this`. Otherwise, the schema attached to `obj.items` must contain the information `this` needs.

whitlockjc:
As part of getting the M2 release out the door, this work was done already and should be available in the 2.1/M2 release.  Please feel free to re-open, or file a new bug, if things aren't working like you'd expect.

Issue #293, opened 2015-03-18
Expose response headers in the model


mohsen1:
From https://github.com/swagger-api/swagger-ui/issues/930 SwaggerUi needs response headers in model

whitlockjc:
It's already done on develop_2.0.  Check again.l.please and if you can repeat, give me a recipe.

whitlockjc:
This has been taken care of.

PR #305, opened 2015-03-20
httpClient: Fix response headers. This looks like a typo, but the curren...

liviosoares:
This looks like a typo, but the current response headers are being inhereted from the *request*, not from the response.

This bug prevents swagger-ui from rendering responses appropriately, since the content-type from the server's response is ignored.

### DIFF
--- a/lib/http.js
+++ b/lib/http.js
@@ -178,7 +178,7 @@ SuperagentHttpClient.prototype.execute = function (obj) {
     var response = {
       url: obj.url,
       method: obj.method,
-      headers: headers
+      headers: res.headers
     };
     var cb;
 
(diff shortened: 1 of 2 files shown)


liviosoares:
This also fixes issue  #293 for me.

whitlockjc:
Do you mind adding a test case for this to show it's working?  The code looks fine but policy is to require unit tests.

liviosoares:
Hi @whitlockjc. Sure, I added a test. This was my first time using Mocha (and expect), so please review.

For me, the unit test is broken on current develop_2.0, but fixed with my one liner (@7817330).

whitlockjc:
Adding to M2 milestone since this was a regression.

whitlockjc:
I went ahead and merged this since it was a regression.

PR #286, opened 2015-03-11
Add support for bower.io

eburi:
Added a bower.json File such that it can be installed using bower.

Note: A few things are required to eventually be able to just "bower install swagger-js"

- Add a tag "v2.1.11"

When this is done, swagger-js can be installed with: **bower install swagger-js#v2.1.11-M1**

bower by default downloads the latest not-prerelease version by looking at all tags (currently 2.0.49) and install this version. As there is no bower.json in this version, it will not work properly.

Therefore when doing the next "real"  (not a prerelease according to semver):
- Set the release version in bower.json (e.g. 2.1.20)
- Add a tag with this release (v2.1.20)
- register swagger-js with **bower register swagger-js git@github.com:swagger-api/swagger-js.git**

PR #334, opened 2015-03-30
Add Bower support

ehmicky:
I am following up on #286 as the modifications are not done yet, but I would need Bower support quite soon.

Also #286 included the wrong mail file. This bower.json downloads the minimum set of files required for browser.


fehguy:
Master is the release branch and we don't accept patches directly to it--please send your PR into develop_2.0.

PR #336, opened 2015-03-30
Add Bower support

ehmicky:
I am following up on #286 as the modifications are not done yet, but I would need Bower support quite soon.

Also #286 included the wrong mail file. This bower.json downloads the minimum set of files required for browser.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,0 +1,29 @@
+{
+  "name": "swagger-js",
+  "version": "2.1.0-M2",
+  "main": "browser/swagger-client.min.js",
+  "ignore": [
+    "gulpfile.js",
+    "index.js", 
+    "lib/", 
+    "package.json", 
+    "README.md",
+    "test/",
+    ".*"
+  ],
+  "description": "swagger-client is a javascript client for use with swaggering APIs.",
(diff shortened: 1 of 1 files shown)


whitlockjc:
This commit as-is is useful but when if/when we pull this, we need to create a follow-up issue to go beyond just supporting Bower and build Bower-specific binaries that do not bundle the dependencies within the binary itself.  A good example of how this could be done is [json-refs](https://github.com/whitlockjc/json-refs).

The way json-refs does things is it builds standalone, single binaries just the way swagger-js does now.  But it also builds Bower-specific binaries that are as small/lightweight as possible by avoiding bundling its dependencies in the Bower-specific binaries and instead uses the `dependencies` section of `bower.json`.

whitlockjc:
This is a duplicate of #286.  I'm going to close this and add my feedback/suggestion there.

whitlockjc:
Reopened here since #286 was old and made against the old master.  _(Shouldn't matter but just in case.)_

ehmicky:
Also #286 uses the Node.js library in the `main` field, as opposed to the browser library.

PR #338, opened 2015-03-30
bower.json

eburi:
#286


eburi:
Adds a bower.json

Tagging and versioning needs to be setup properly. Don't have time to digg into that.

fehguy:
handled in https://github.com/swagger-api/swagger-js/pull/336#issuecomment-87767646

PR #344, opened 2015-04-01
Update models.js

hamx0r:
UI should show Object name unless _title_ is part of the definition object, in which case it should display the title value.


webron:
@hamx0r - I believe you know this by now, but PRs should be made against the develop_2.0 branch ;)

hamx0r:
Oops, my mistake.   First time writing a JS test and lost sight of protocol =)

PR #345, opened 2015-04-01
Update models.js

hamx0r:
UI should show Object name unless title is part of the definition object, in which case it should display the title value.


fehguy:
@whitlockjc can you look at this?

whitlockjc:
The pull request is failing merge and it appears it only contains the unit test and not the actual work to make the unit test pass.

hamx0r:
pull #352 supersedes this PR as it has both the change and a test for the change.

Issue #346, opened 2015-04-02
basePath not taken into account for v1 spec (resource listing)


sberryman:
`SwaggerSpecConverter.prototype.convert()` is not taking into account the basePath for a resourceListing.

in `/lib/client.js` (https://github.com/swagger-api/swagger-js/blob/master/lib/client.js#L182-L183) you are passing in self.url which I'm sure is fine for almost everyone but not for hapi-swagger's output.

I did a simple test of setting the docLocation by adding the following AFTER line 34:
https://github.com/swagger-api/swagger-js/blob/master/lib/spec-converter.js#L34

```js
// basePath
if (obj.basePath) {
  this.setDocumentationLocation(obj.basePath);
}
```
Once this was added, everything worked like a charm.

fehguy:
hmm could you send a PR and we can test?

sberryman:
done

PR #347, opened 2015-04-02
Update spec-converter.js to take into account obj.basePath

sberryman:
issue #346


webron:
@sberryman - PRs need to be submitted against the develop_2.0 branch. Please resubmit.

Issue #239, opened 2015-02-13
make SwaggerHttp pluggable


fehguy:
It'd be easy to register different transports for executing http requests, as we do for swagger authorizations.  Then proxies, websockets, etc., can be added

whitlockjc:
We moved from shred to superagent as a part of https://github.com/swagger-api/swagger-js/pull/279, which should resolve https://github.com/swagger-api/swagger-js/issues/267.  When we work on this pluggable feature, I'd like to remove jquery as superagent should make it no longer necessary.  Superagent works on the browser and in Node.js, while jquery does not, and it supports all of the features that require using jquery (form data, cors, ...).

hekike:
+1

aurelian:
:+1:

andrewdavidmackenzie:
Any indication of where this lies on your priority stack Tony?
It's a must have for one of the features our customers appreciate when using our Swagger based (1.2) ActiveDocs.
Thanks

fehguy:
Yes, I'd like to get this in the next release, @andrewdavidmackenzie It's needed also to address a regression in IE.

fehguy:
added by https://github.com/swagger-api/swagger-js/pull/428

aurelian:
Unfortunately it means you have to write your own HTTP/s js client instead of re-using (extending) existing ones. Nice progress thou.

PR #348, opened 2015-04-02
Remove dependency on jquery

aurelian:
This one removes the dependency on jquery as the heavy-lifting of doing requests is handled by superagent both in browser and in nodejs.


webron:
kill-jquery! (yes, my job is to add important information)
screenshot 2015-04-02 17 48 35

whitlockjc:
I wanted to do this in M2 but didn't think the time was right.  I think this should be fine and will be a good start to working on making SwaggerHttp pluggable _(#239)_.  I will let @fehguy comment on this but the work looks fine.

PR #352, opened 2015-04-02
Use title in models when available per swagger-ui #1104

hamx0r:
Modified source and created test to make use of "title" in models when avaialble


dmyers:
Is there a way to add support for referenced schemas? I have title attribute on all of my schemas, but ones that are referenced are not using the title though.

ss

ss2

PR #354, opened 2015-04-03
Default value attribute as an array for array type parameters.

valdemon:
Hi,

this is related and required by the https://github.com/swagger-api/swagger-ui/pull/1121 PR in a context of default values for the array type parameters.

Additionally - enum definition at array items level as an option. This approach conforms either to Swagger Spec or to JSON Schema, so probably should be handled.

Best,
Waldek


webron:
@valdemon - thank you for the PR, but all PRs should be submitted against the develop_2.0 branch.

PR #357, opened 2015-04-07
Support csv strings for array parameters

ddm:
This will allow swagger UI to support array params with any collectionFormat


fehguy:
this shouldn't be necessary--you can pass an array as the input rather than a CSV.

ddm:
I understand this should be fixed at the swagger-ui level but how do you input an array in a swagger-ui text field?

PR #362, opened 2015-04-13
fix the parse error from the json generated by springfox.

nurinamu:
Hi

When I'm using sprintfox, swagger-ui failed to display my json. So I traced the code and found the difference in the data structre using SwaggerSpecConverter.prototype.toJsonSchema().

I added the fixing code. please review the code and accept this.


webron:
- What was the actual problem? They fix may need to come from springfox itself.
- PRs should be made against the develop_2.0 branch, not master.

nurinamu:
detectedType.toLowerCase() fired the exception when detectedType has a object. It includes the actual value in one more lower level - absoluteType. Anyway I'll generate the issue to springfox and close this issue. Thanks.

PR #370, opened 2015-04-16
parseBody handle whitespace in responses

BjRo:
Hi there,

we're currently investigating using swagger as the primary means to document our APIs at XING. We really like swagger so far and see a lot of potential in it for us. You guys do an awesome job. 

During our tests with swagger we bumped into little things here and there. I just wanted to contribute a little bit back :-)

Our APIs are mostly Rails based and Rails seems to return responses  body containing whitespace on 201s, which results in a parser error. This little patch makes the client a little more robust against that.

I tried to write a test for that as well, but unfortunately I'm not a javascript crack and failed to succeed. I hope that's alright ^^

### DIFF
--- a/browser/swagger-client.js
+++ b/browser/swagger-client.js
@@ -18484,7 +18484,7 @@ Response.prototype.setHeaderProperties = function(header){
 
 Response.prototype.parseBody = function(str){
   var parse = request.parse[this.type];
-  return parse && str && jQuery.trim(str).length
+  return parse && str && str.trim().length
     ? parse(str)
     : null;
 };
--- a/browser/swagger-client.js
+++ b/browser/swagger-client.js
@@ -18484,7 +18484,7 @@ Response.prototype.setHeaderProperties = function(header){
 
 Response.prototype.parseBody = function(str){
   var parse = request.parse[this.type];
-  return parse && str && str.trim().length
+  return parse && str && jQuery.trim(str).length
     ? parse(str)
     : null;
 };
--- a/browser/swagger-client.js
+++ b/browser/swagger-client.js
@@ -18484,7 +18484,7 @@ Response.prototype.setHeaderProperties = function(header){
 
 Response.prototype.parseBody = function(str){
   var parse = request.parse[this.type];
-  return parse && str && str.length
+  return parse && str && str.trim().length
     ? parse(str)
     : null;
 };


browser/swagger-client.js
```diff
@@ -18484,7 +18484,7 @@ Response.prototype.setHeaderProperties = function(header){
 
 Response.prototype.parseBody = function(str){
   var parse = request.parse[this.type];
-  return parse && str && str.length
+  return parse && str && str.trim().length
```
mohsen1: IE issue here again!
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim#Browser_compatibility

BjRo:
Updated to use jQuery for trimming.

fehguy:
@mohsen1 swagger-client doesn't depend on jquery, so we can't add any reference to it.

@BjRo the `browser/*` files are generated from the source files in `lib` during the `gulp` task, so upon rebuild your change will get wiped out.

Seems like the proper solution is to manually do the trim as to avoid the query requirement?

mohsen1:
Oh I missed that SwaggerJS on its own doesn't depend on jQuery. Maybe `str..replace(/^\s+|\s+$/g, '')` ?

BjRo:
@fehguy @mohsen1 Now looking at that, I don't think `parseBody` is coming from `swagger-js` itself. Looks more like it's coming from the `superagent` package https://github.com/visionmedia/superagent/blob/abc5a452778cd6c7a03bd8a8c0c6067835990f8e/lib/client.js#L360 and is only merged into the `swagger-client.js` file. Can you confirm?

fehguy:
I see.  Two thoughts.

1) We do have the response object in-hand here:

https://github.com/swagger-api/swagger-js/blob/master/lib/http.js#L204

So it is possible to take the statusText and trim it.  However, I don't know if that's a general need across all swagger-js users and specific to your service.  So the right solution is probably this enhancement:

https://github.com/swagger-api/swagger-js/issues/126

So you can plug in some special response processing.

BjRo:
So, I've tried to remove the patch from `parseBody` and apply it to `SwaggerHttpClient.execute` but unfortunately that doesn't work. 

When Rails sends a 201 with whitespace in the response body and `application/json` as the content type 
already the creation of the `Response` object [fails](https://github.com/swagger-api/swagger-ui/blob/master/dist/swagger-ui.js#L19636) because in the `Response` constructor it already tries [to parse the body](https://github.com/swagger-api/swagger-ui/blob/master/dist/swagger-ui.js#L19484).  

This is how the response ends up being represented in the UI

screen shot 2015-04-17 at 09 06 21

I'm not sure whether the Response interceptor would solve this problem because in that particular case the Response can't even be created :-/

How do you think should we proceed with this? My fix clearly doesn't work (at least the way I proposed it) Should I open a bug and close this PR?

fehguy:
@BjRo can we just use str.trim() to avoid the jquery dependency?  Aside from that, yes this is fine.

BjRo:
@fehguy I reverted the jQuery commit

fehguy:
merging to develop_2.0

PR #371, opened 2015-04-16
Expose http client via transport option

### LINKED ISSUE #239: make SwaggerHttp pluggable

aurelian:
This pull request does 2 things:

 * Removes dependency on jquery - closes #348 (obsolete)
 * Expose http client via an option that can be passed on client - closes #239

```js
var swagger = new client({
  url: 'http://petstore.swagger.io/v2/swagger.json',
  success: function() {
    swagger.apis.pet.getPetById({petId:1});
  },
  transport: function(httpClient, obj) {
      // you have access to obj before the request is made

      // eventually, httpClient has to make the request
      var result = httpClient.execute(obj);
      
      // you can do customizations after the request was done

      //  will be nice to return the result here
      return result;
  }
});
```


fehguy:
Thank you for this.  I'm going to close this out, and follow your pattern but with some changes.  Please watch issue #239 for updates.

PR #380, opened 2015-04-20
ui#1133: add more support for inline schema's

ponelat:
Fixes swagger-api/swagger-ui#1133.

operation#getType returns a string, which was fine for everything but
inline models, which until now couldn't be referenced.

opertation#getType now returns 'inline#123' for inline models, where 123
is the index of the inline model.
I've put in an array, operation#inlineModels, that keeps track of
inline models added via operation#addInlineModel.
To de-reference inline models, use operation#getInlineModel, passing it
the 'inline#123' string

operation#getSampleModelJson, uses the above.


fehguy:
@whitlockjc can you please review this?

Issue #343, opened 2015-04-01
Parameter display does not display arrays of models properly


whitlockjc:
Whenever you have an array of models and the schemas items are a reference, instead of `Array[ReferencedType]` you get `Array[Reference]` in the parameter display.  So for example, the following parameter schema:

```js
// ...
{
  name: 'testing',
  in: 'body',
  schema: {
    type: 'array',
    items: {
      $ref: '#/definitions/Person'
    }
  }
}
// ...
```

results in a display of `Array[#/definitions/Person]` instead of `Array[Person]`.

This was original reported from [domaindrivendev/Swashbuckle/issues/200](https://github.com/domaindrivendev/Swashbuckle/issues/200).

beryozad:
As noted elsewhere this behavior can also be observed in http://petstore.swagger.io/#!/user/createUsersWithArrayInput - parameter "body" is shown as "Array[#/definitions/User]"

Once it is fixed, will it only be shown as "Array[User]", or will there be a way to expand it to see the definition of User? 

Thanks!

whitlockjc:
@ponelat So is this fixed after merging #381?

ponelat:
Yes it works in the latest develop_2.0 now.
@beryozad Arrays of models will render the JSON as expandable, something like....

```javascript
[
  {
     "your": "model here",
      "...": "..."
  }
]

whitlockjc:
What about parameters?  When you see a parameter that is an array of models, inline or referenced, does it show `Array[SomeDefinition]` or `Array[#/definitions/SomeDefinition]`?

ponelat:
It wouldn't show a string signature but an expandable widget(lack of a better word?)
Like so.... (Note on the right)


![screenshot from 2015-04-23 17 58 43](https://cloud.githubusercontent.com/assets/8438485/7301356/8619b67e-e9e2-11e4-9a97-7405dd4a55d9.png)

whitlockjc:
That's correct.  I think it showed the string version because it wasn't typed right but it seems you've fixed this.  Thanks a lot.

PR #381, opened 2015-04-20
Fix/swagger ui 1037

ponelat:
fixes swagger-api/swagger-ui#1037
fixes swagger-api/swagger-js#343 

Based on pending PR... https://github.com/swagger-api/swagger-js/pull/380

Minor tweaks in this PR. Just a typo and a mis-step. With a test.
Remember not to confuse the amount of code in this PR, with the previous PR. Have a look at the commit rather... https://github.com/swagger-api/swagger-js/commit/1ecfe04ee4b6d912e3721c5331ebe45f8d5974d0
PS: I removed an old test (of mine) which kinda screws up the diff a little.

### DIFF
--- a/lib/types/operation.js
+++ b/lib/types/operation.js
@@ -246,7 +246,7 @@ Operation.prototype.getType = function (param) {
     } else {
       // If inline schema, we add it our interal hash -> which gives us it's ID (int)
       if(schema.type === 'object') {
-        return 'inline#'+ this.addInlineModel(schema);
+        return this.addInlineModel(schema);
       }
       return this.getType(schema);
     }
@@ -268,7 +268,7 @@ Operation.prototype.addInlineModel = function (schema) {
   var model = this.resolveModel(schema, {});
   if(model) {
     this.inlineModels.push(model);
-    return len; // to be added to the ID...eg: 'inline#123', where 123 is the ID
+    return 'Inline Model '+len; // return string ref of the inline model (used with #getInlineModel)
   }
   return null; // report errors?
 };
@@ -279,8 +279,8 @@ Operation.prototype.addInlineModel = function (schema) {
  * @return {Model} the model being referenced. Or null
  **/
 Operation.prototype.getInlineModel = function(inline_str) {
-  if(/inline#/.test(inline_str)) {
-    var id = parseInt(inline_str.split('#')[1],10); //
+  if(/^Inline Model \d+$/.test(inline_str)) {
+    var id = parseInt(inline_str.substr('Inline Model'.length).trim(),10); //
     var model = this.inlineModels[id];
     return model;
   }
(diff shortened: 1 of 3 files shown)


fehguy:
@ponelat looks good to me.  @whitlockjc could you please review and merge if good from your POV?

lib/types/operation.js
```diff
@@ -242,6 +244,10 @@ Operation.prototype.getType = function (param) {
         return ref;
       }
     } else {
+      // If inline schema, we add it our interal hash -> which gives us it's ID (int)
+      if(schema.type === 'object') {
+        return 'inline#'+ this.addInlineModel(schema);
```
whitlockjc: I really think this should be modeled after how we handle inline schemas in responses, for consistency of course..  So instead of `inline#{num}` it would be `Inline Model {num}`.

lib/types/operation.js
```diff
@@ -252,6 +258,36 @@ Operation.prototype.getType = function (param) {
   }
 };
 
+/**
+ * adds an inline schema (model) to a hash, where we can ref it later
+ * @param {object} schema a schema
+ * @return {number} the ID of the schema being added, or null
+ **/
+Operation.prototype.addInlineModel = function (schema) {
```
whitlockjc: Why do we need this?

lib/types/operation.js
```diff
@@ -252,6 +258,36 @@ Operation.prototype.getType = function (param) {
...
+  }
+  return null; // report errors?
+};
+
+/**
+ * gets the internal ref to an inline model
+ * @param {string} inline_str a string reference to an inline model
+ * @return {Model} the model being referenced. Or null
+ **/
+Operation.prototype.getInlineModel = function(inline_str) {
```
whitlockjc: Why do we need this?

lib/types/operation.js
ponelat: As far as I can tell (looking at lib/type/model.js ), "Inline Model" is just decoration? Which is fine...consistency is king.

whitlockjc: Nah, we actually print that when you look at the Model Schema in the response.  Here is a test showing this: https://github.com/swagger-api/swagger-js/blob/master/test/models.js#L313

lib/types/operation.js
ponelat: Abstraction, this is where we cache the model for de-referencing. I wouldn't want that logic inside of operation#getType... And in the commit below it's used to return the full string ID, further abstracting things.

whitlockjc: Gotcha.  I was just wondering why it was on the prototype.  No big deal.

lib/types/operation.js
```diff
@@ -252,6 +258,36 @@ Operation.prototype.getType = function (param) {
...
+  if(/Inline Model/.test(inline_str)) {
```
whitlockjc: I just saw this.  This regex will match `Inline Model*` but what we really want is `Inline Model {num}`.  So this regex should be `^Inline Model \d+$` so that we only match what we want so we have less of a chance of false positives.

whitlockjc:
Thanks.

PR #387, opened 2015-04-23
Fix/swagger ui 1199

ponelat:
An oversight regarding the models rendering.
fixes swagger-api/swagger-ui#1199
Also minor tweak to pass lint

PR #393, opened 2015-04-27
add String#trim pollyfill and tests

### LINKED ISSUE #386: Normalize instances of String#trim
I noticed a comment on using String#trim and IE8 Support (talking about you @mohsen1 ).
I made the mistake, and it looks like this code has a few instances...
https://github.com/swagger-api/swagger-js/blob/master/lib/http.js#L101-L117

The only reason that I'm creating an issue and not a direct PR, is because I see two solutions...
Make sure we use jQuery.trim (enforce it with a comment somewhere in contributions?)
Or better yet, pollyfill it with something like this...

Taken from CSSTricks...https://css-tricks.com/snippets/javascript/strip-whitespace-from-string/
```javascript
if (!String.prototype.trim) {
  String.prototype.trim = function () {
    return this.replace(/^\s+|\s+$/g, '');
  };
}
```

ponelat:
fixes #386 
Add a polyfill for String#trim.
Also include tests for the two other polyfills I found (String#endsWith and Array#indexOf)


webron:
Can you rebase please?

whitlockjc:
#387 has some of these changes in it so it looks like this PR needs to be rebased, committed and then #387 needs to be rebased after this commit to remove the duplicate stuff from it.

PR #394, opened 2015-04-28
Fix/swagger ui 1152 - update rendering of 'required' props

ponelat:
fixes swagger-api/swagger-ui#1152

Three things:
1) Convert old v1.2 'required' array, into v2.0... someProperty.required. = true
2) Updated the check on required properties in model#schemaToHTML (aka, model#getMockSignature)
3) Strengthened some fragile tests, which relied on faulty (pre-fix) values. Instead of checking the generated HTML, we just take a peek with some regexes.

Number 2,3 are stop-gap measures, until we can remove HTML generation from swagger-js. Or at least make that functionality a plugin.


cancan101:
Does this change just mark all properties as required though?

cancan101:
NM. Looks like some issue with DRF-swagger.

lib/spec-converter.js
```diff
@@ -139,6 +139,17 @@ SwaggerSpecConverter.prototype.models = function(obj, swagger) {
     if(_enum.length > 0) {
       schema['enum'] = _enum;
     }
+
+    // Convert required array into required props on .property
+    if(existingModel.required instanceof Array){
```
whitlockjc: Might be safer to use `_.isArray` but this is not a requirement.

test/models.js
```diff
@@ -77,7 +77,12 @@ describe('models', function () {
     };
     var model = new Model('Sample', definition);
 
-    expect(model.getMockSignature()).toEqual('<span class="strong">Sample {</span><div><span class="propName false">id</span> (<span class="propType">integer</span>, <span class="propOptKey">optional</span>),</div><div><span class="propName false">name</span> (<span class="propType">string</span>, <span class="propOptKey">optional</span>),</div><div><span class="propName false">photos</span> (<span class="propType">Array[object]</span>, <span class="propOptKey">optional</span>)</div><span class="strong">}</span>');
```
whitlockjc: Testing the generated HTML is a pain in the butt but I don't think removing them makes much sense here.  The tests you've replaced them with are quite lax and do not ensure the structure created is what is required.  While I do think we need some utilities for easing the pain of this, I'm not sure making things more lax makes sense.

This comment applies to all similar changes below so I will not be marking the other places.

whitlockjc:
This PR needs rebasing as well.  The comment about testing might be enough to keep this PR from getting my +1 because while it's a pain to write good tests for the generated HTML, it's required.

ponelat:
@whitlockjc I've rebased and put in the original (albeit updated) html tests.

PR #397, opened 2015-04-30
Use primitives in param.enum for booleans

ponelat:
Linked to https://github.com/swagger-api/swagger-ui/issues/1191#issuecomment-97450141

Use primitives in the param.enum list for booleans. Not strings of booleans.
Allow default values to be either a primitive, or a string of that primitive.
Make sure v1.2 default values are carried across correctly (they allow for default = false)


lib/types/operation.js
```diff
@@ -90,7 +90,7 @@ var Operation = module.exports = function (parent, scheme, operationId, httpMeth
     if (innerType && innerType.toString().toLowerCase() === 'boolean') {
       param.allowableValues = {};
       param.isList = true;
-      param['enum'] = ['true', 'false'];
+      param['enum'] = [true, false]; // use actual primitives
```
whitlockjc: We probably need to handle both or update the converter to do type coercion.  In Swagger 1.x, boolean values were their string counterpart so this change works great for Swagger 2.0 but will likely fail for Swagger 1.x.

webron: They were? :) But I'm ok with keeping both.

whitlockjc: Maybe there is code elsewhere in the converter that already handles this.  If you're sure it's fine, I've no reason to push back.  :)

webron: No, I'm just saying that technically, boolean in 1.X should have been boolean values and not strings, but for the sake of robustness there's probably no harm supporting both.

whitlockjc: Ah yes, that's right.  I'm going to delete the comments now.

Issue #409, opened 2015-05-07
Top level example does not end up in model


Hes-Siemelink:
In Swagger 2.0 spec you can set an example on a definition. For example:

    {
        "definitions": {
            "Release": {
                "description": "The release type.",
                "example": "{ Handcrafted JSON example for RELEASE type }",

This is not being picked up in the model (`swagger-client.js`), so swagger-ui won't display it.

fehguy:
fixed in #410

PR #410, opened 2015-05-07
Fix for #409: Top level example is included in model

Hes-Siemelink:
Examples on top-level definitions are now included in the model.

See issue #409


Hes-Siemelink:
Thanks Tony! :-)

Issue #323, opened 2015-03-26
Support YAML


mohsen1:
https://github.com/swagger-api/swagger-ui/issues/1047

I've created a [branch in SwaggerUI](https://github.com/swagger-api/swagger-ui/tree/yaml) which has js-yaml dependency. Just make swagger-js depend on js-yaml and use `jsyaml.load` instead of `JSON.parse`

fehguy:
@ponelat can you take this?

PR #415, opened 2015-05-08
add js-yaml as drop in replacement for JSON.parse

ponelat:
fixes https://github.com/swagger-api/swagger-js/issues/323

Just replaced JSON.parse with jsyaml.load.
Added js-yaml to package.json.
Built the files.

PS: @mohsen1 's branch uses `"js-yaml": "^3.2.7"` while this one's on the latest at `"js-yaml": "^3.3.0"`


fehguy:
thanks @ponelat can you add some tests?

ponelat:
@fehguy I was going to, then realized they'd be slightly intense. I can get at it on Monday.
I've added comments to those that need tests, as forcing errors at those locations didn't break our test runner :C.
I'll be writing unit tests for these, on the invoking methods.

fehguy:
Sure thing.  It's not a rush as I just published 2.1.3-M2 last night.  I can help get some sample YAML in the mock servers.

mohsen1:
Please feel free to delete my yaml branch after this is merged

ponelat:
I'm still working on the tests for this, I need clarity on two things...
- **One** do we have browser tests? The only thing I can see is this.. https://github.com/swagger-api/swagger-js/blob/develop_2.0/browser/index.html

- **Two** Why do we have fallbacks for response.obj (namely we try to parse response.data and response.statusText if response.obj isn't an object) ? 

I'll be researching into this, but thought I'd let you know where my head is at.

Re browser tests, if we don't have, perhaps we could use karma + browserify transform on the existing tests.

On another note, swagger-client is called in the mock server (which is called before each test) causing my initial test to fail in the beforeEach hook. As far as I can tell, swagger-client is only called for some sort of macro functionality @whitlockjc is that needed in the mock? Over here ... https://github.com/swagger-api/swagger-js/blob/develop_2.0/test/mock.js#L90

ponelat:
Superagent has tests. Now to test for jquery...
Also I cleaned up a little around the expected response of from SwaggerHttp.

We were redundantly trying to parse data from the response AFTER SwaggerHttp had tried. 
There should be no cause for this, if SwaggerHttp failed... what other info is there to parse the files?

I see that Travis is failing on timeout's in the browser tests.... I'll have to look into that :/

mrhanlon:
I wonder if the browser tests timeout is related to this: https://github.com/SeleniumHQ/selenium/issues/437

fehguy:
I've manually merged this code into the `feature/relative-ref` branch.

PR #454, opened 2015-06-03
Don't warn when produces is "*/*".

joeheyming:
I'm sure this is a very naive fix, but I just wanted to throw that out there unless someone has any better ideas


webron:
Sorry, had to revert it. Please resubmit against develop_2.0 and we'll merge again.

joeheyming:
Thanks for the heads up.  Sorry for the mistake

Issue #455, opened 2015-06-03
allOf composition support


fehguy:
We need to support composition and inheritance that can be modeled with `allOf`.

webron:
Umm... #188?

fehguy:
as @webron so helpfully pointed out, we have an existing issue.

PR #456, opened 2015-06-03
Feature/composition

fehguy:
Addresses #455.  Notes on this integration

* The primary use cases are for composing multiple objects and inheritence
* To track the "origin" of properties, we are using a vendor extension in the resolver, `x-resolved-from`.
* To track the "parents" of a composed object, we are using a vendor extension, `x-resolved-from` which contains an array of _locations_ from which the object was created

Limitations:
* multiple compositions are not supported.  They can be, but currently are not.
* The rendering support is not complete.  The objects will contain the necessary reference information, but the json sample & mock signatures are not added
* Others?


fehguy:
This should address #188

webron:
When you say multiple compositions are not supported, does that mean more than two entries in the `allOf` array, or nested `allOf` definitions?

fehguy:
@webron nested are not supported.  It's not much work, but if we can push that to a dot release, I'd appreciate it.

webron:
Sure, was just looking for a clarification. :+1:

Issue #445, opened 2015-06-02
Authorization for the inital fetching of swagger spec file.


ponelat:
Short of recreating a rough implementation of SwaggerAuthorization (yuck), there isn't much one can do to include authorization headers/tokens for the fetching the spec file.

> Authorizations only get applied after we fetch the spec file. Is it required that spec files are public?

I think the correct functionality should be so....
- The [authorization](https://github.com/swagger-api/swagger-js/blob/develop_2.0/lib/client.js#L131) option in SwaggerUi/SwaggerJS can contain a hash of `{ authName: authFunc(){} }`, 
- which will get appended to a fresh `SwaggerAuthorization` instance on initializing.

This happens before any HTTP executes are done, and so will apply to all. 

The only question is whether we should have a special authName for fetching the spec file (so that spec specific headers/tokens don't get sent with other requests). Or we rely on the security arrays inside the spec file to _NOT_ include our arbitrarily named auth (for fetching the spec file).

fehguy:
@ponelat I don't think that's true--we should be able to set the authorization headers when initializing swagger-client.  I think that's a requirement for it to work correctly.

If it is the case, that's a regression in the client.

ponelat:
@fehguy the only possible way to do so is via the `authorization` field. But we require it to be an instance of SwaggerAuthorization. Which the user has no way of creating, as it isn't exposed. Much simpler would be to create the instance from a hash, no?
ApiAuthorizationKey and such _are_ exposed by-the-way.

fehguy:
I believe @whitlockjc changed it so you can call the constructor then the `initialize(args)` function, which allows you to add the authorizations at that point, which is _before_ calling to get the spec.

ponelat:
You can, so long as you don't provide a `url` which would trigger `initialize` from within the constructor.
However the crux is still within `initialize` which is where SwaggerAuthorization is instantiated, _unless_ we define it as an option.... but we _must_ define it as an instance of SwaggerAuthorization which is the original argument.

fehguy:
OK got it.  So the proper flow should be something like  this:

```
var client = new SwaggerClient(); // initializes swagger authorizations
client.authorizations.add(...);
client.initialize({url: url, opts: opts});
```

Or something like that...

ponelat:
That would be much better than ....
```javascript
var client = new SwaggerClient({
   authorization: new window.SwaggerAuthorization(), // assuming we expose it
   url: false
});
[...]

PR #458, opened 2015-06-04
Fix/auth for initial spec

ponelat:
fixes https://github.com/swagger-api/swagger-js/issues/445
Three different types of workflows are supported when adding auth.

#### Pre-initialize
```javascript
var client = new SwaggerClient();
client.clientAuthorizations.add('someAuth', function(){ });
client.clientAuthorizations.add('moreAuth', new SwaggerClient.ApiKeyAuthorization('moreAuth', 'key', 'value'); // Or using one of the predefined manipulates, exposed on SwaggerClient
client.initialize({ .. });
```

#### One-step
```javascript
var client = new SwaggerClient({
  authorizations: {
    someAuth: function() {}
  },
  url: '...',
  success: function() {}
});
```
#### Post initialize + options
```javascript
var client = new SwaggerClient({
  authorizations: { // our options
    someAuth: function() {}
  },
  url: '...',
  success: function() {
    client.clientAuthorizations.add('anotherAuth', function(){}); // Post init'd
  }
});
```

As a bonus, you can add a whole hash to `clientAuthorizations`
```javascript
client.clientAuthorizations.add({
  someAuth: function() {},
  moreAuth: function() {}
  // ...
});

### CODE REVIEW

**Comment by ponelat** [comment]
On another note, 
I'm using http mocks instead of a dummy server.
Which I feel is much simpler to maintain... I first used nock (quite nice, but only works in node.js) then migrated to faux-jax which works in both node.js and the browser.

What happens is faux-jax will replace the core methods used to make requests (http.request for node.js and XMLHttpRequest for the browser) thus allowing us capture the requests and return what we like... including status code, headers, etc.

It also allows us to check the actual request sent.

**Comment by ponelat** [comment]
@whitlockjc I haven't replaced clientAuthorizations as I'm not sure if it'll break compatibility. I felt the above was a good compromise.

File: lib/auth.js [hunk 1]
```diff
@@ -6,12 +6,18 @@ var CookieJar = require('cookiejar');
 /**
  * SwaggerAuthorizations applys the correct authorization to an operation being executed
  */
-var SwaggerAuthorizations = module.exports.SwaggerAuthorizations = function () {
-  this.authz = {};
+var SwaggerAuthorizations = module.exports.SwaggerAuthorizations = function (authz) {
```
whitlockjc: I really don't think we need to keep this unless it's for backward compatibility.  `SwaggerAuthorizations` is really just a JavaScript object from an implementation perspective so wrapping it doesn't make much sense.

Of course, how the review of the rest of this PR goes depends on whether we agree on this or not.

ponelat: I agree, I'm not much a fan of SwaggerAuthorization... but I do need something to do the following.. https://github.com/swagger-api/swagger-js/blob/master/lib/auth.js#L23-L74
And as a namespace, its fine. We won't be exposing it.

whitlockjc:
I think what you have is fine but as a JavaScript author, it feels weird to me to have a `SwaggerAuthorizations` object just to expose an `apply` method.  Not only that but with the `SwaggerAuthorizations` container hiding the JavaScript object used for key/value storage, we now have to maintain the code required to provide access to the container.  Another thing to think about is JavaScript in general.  JavaScript developers are use to dot/bracket notation for key/value storage and what we have now is more like a Java API.

All of this being said, this is just style so I will give my +1 for the overall fix.  But I still think we could end up with a better JavaScript experience if we blew away the `SwaggerAuthorizations` object itself and instead just exposed a function that would would be used to replace what `SwaggerAuthorizations#apply` does.

Issue #462, opened 2015-06-05
Remove old references to wordnik, make agnostic


ponelat:
As per email request, make swagger-js agnostic about it's parent maintainer.

webron:
This is done, thanks!

PR #463, opened 2015-06-05
change email address in spec files to apiteam@swagger.io

ponelat:
fixes https://github.com/swagger-api/swagger-js/issues/462
Changed the email addresses in the spec file

@webron before we merge this, I want to know whether we are keeping the oauth urls.
https://github.com/swagger-api/swagger-js/blob/master/test/spec/v2/petstore.json#L807


ponelat:
Never mind last remark, I found the correct oAuth URL

PR #481, opened 2015-06-09
Remove the 'name' argument of PasswordAuthorization

### LINKED ISSUE #480: SwaggerClient.PasswordAuthorization takes an unused argument
The prototype of PasswordAuthorization constructor is `function (name, username, password)`. The argument `name` is superfluous (and it's not used), so it should be removed.

zub2:
fixes #480

Three arguments are still supported in order to not break old apps, but
a message is logged when three arguments are used.


fehguy:
eww.  I didn't see this was in the original code

Issue #219, opened 2015-01-31
add extended formats


fehguy:
It'd be nice to support more extended formats:
- [ ] password
- [ ] uuid, uuid3, uuid4, uuid5
- [ ] email
- [ ] uri (absolute)
- [ ] hostname
- [ ] ipv4
- [ ] ipv6
- [ ] credit card
- [ ] isbn, isbn10, isbn13
- [ ] social security number
- [ ] hexcolor
- [ ] rgbcolor
- [ ] date
- [ ] date-time
- [ ] duration

webron:
Closing as an old ticket.

PR #485, opened 2015-06-11
Extract format field from property models and parameters when it is a…

joepadmiraal:
This checks whether the models and parameters in the Swagger definition have a format field. If it is present it will be placed on the generated model.

I think this solves issue #219.


webron:
It doesn't resolve #219, but it's a step in that direction. Thanks.

PR #496, opened 2015-06-24
specify host in options

pm100:
allow swagger-ui to specify host url. This is useful when the rest api target host name is a) not known in advance and b) is not the machine hosting swagger


webron:
@pm100 - I understand this relates to the issue you opened on swagger-ui (would help if you link to it from the PR). I can't say whether this will be merged in general, but you also need to submit the PR against the develop_2.0 branch. master is our stable branch.

PR #497, opened 2015-06-24
Use a SPDX-compatible identifier for the 'license' value

ankon:


webron:
Same here, please resubmit against the develop_2.0 branch.

PR #498, opened 2015-06-24
Set operation.schemes to the value of 'schemes' if set

ankon:
This is based on my reading of the 2.0 spec: https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#operationObject says that the `schemes` in the operation override any `schemes` in the parent.


webron:
Looks like you're on a PR spree (which is great!) but really, we only accept PRs against the develop_2.0 branch.

ankon:
Sorry, missed that part - reopening them after rebasing. Sorry for the noise :(

PR #501, opened 2015-06-24
Apply global security requirements

ankon:
Based on https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#securityRequirementObject and the global `security` element I read this as that I should be able to specify global security requirements via the global `security` element, and that a single operation can (but doesn't have to) modify the requirements by having a `security` element.

There is a bit of confusion (for me ...) related to the "AND" (local) vs "OR" (global) interpretation of multiple security requirements, so I might be missing something here.

The idea in any case would be to define APIs that always require authentication/authorization, without having to repeat the same block of `security` requirements for each and every operation.

Please note that I think 02a799f and 1fddfcd should be applied in any case :)

### DIFF
--- a/lib/types/operation.js
+++ b/lib/types/operation.js
@@ -19,7 +19,7 @@ var Operation = module.exports = function (parent, scheme, operationId, httpMeth
     this.client = parent.options.client || null;
     this.responseInterceptor = parent.options.responseInterceptor || null;
   }
-  this.authorizations = args.security;
+  this.authorizations = args.security || parent.security;
   this.basePath = parent.basePath || '/';
   this.clientAuthorizations = clientAuthorizations;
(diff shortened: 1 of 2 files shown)


lib/types/operation.js
```diff
@@ -694,7 +694,7 @@ Operation.prototype.execute = function (arg1, arg2, arg3, arg4, parent) {
     }
   };
 
-  this.clientAuthorizations.apply(obj, this.operation.security);
+  this.clientAuthorizations.apply(obj, this.security);
```
fehguy: This is going to need some testing.  If you can add anything around this to ensure we're not introducing a regression on operation-based security, we can move forward.

fehguy:
closing due to inactivity.  If you can reopen with some tests, we'll get this in.

PR #506, opened 2015-06-24
fix allOf support

ramonsnir:
Fix `allOf` to work as intended.

The example JSON we have:

```js
"definitions": {
  "Parent": {
    "properties": {
      "hey": {
        "type": "string"
      }
    }
  },
  "Child": {
    "allOf": [
      {
        "$ref": "#/definitions/Parent"
      },
      {
        "properties": {
          "there": {
            "type": "string"
          }
        }
      }
    ]
  },
  "Grandchild": {
    "allOf": [
      {
        "$ref": "#/definitions/Child"
      },
      {
        "properties": {
          "bob": {
            "type": "string"
          }
        }
      }
    ]
  }
}
```

And in the Swagger UI now the model looks like (as should be):

```js
{
  "grandchild": {
    "hey": "string",
    "there": "string",
    "bob": "string"
  },
  "child": {
    "hey": "string",
    "there": "string"
  },
  "parent": {
    "hey": "string"
  }
}
```

instead of the previous (incorrect) result:

```js
{
  "grandchild": {
    "bob": "string"
  },
  "child": {
    "hey": "string",
    "there": "string"
  },
  "parent": {
    "hey": "string",
    "there": "string"
  }
}
```

### DIFF
--- a/lib/resolver.js
+++ b/lib/resolver.js
@@ -31,6 +31,7 @@ Resolver.prototype.resolve = function (spec, arg1, arg2, arg3) {
     }
 
     if(definition.allOf) {
+      definition['x-resolved-from'] = [ '#/definitions/' + name ];
       var allOf = definition.allOf;
       // the refs go first
       allOf.sort(function(a, b) {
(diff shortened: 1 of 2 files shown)


tkatkov:
Try adding to your test:

test.undefined(spec.definitions.Animal.properties.firstName);

ramonsnir:
@tkatkov passes.

PR #509, opened 2015-06-25
Fix inability to apply multiple apiKey authentications

PASSTor:
If you have several apiKey authentications defined (like key+secret),
they were not applied both to the Request, because of
`status = status || !!auth.apply(obj);`

Once status becomes `true` after applying the first security rule, the
seconds (and all next ones) are ignored, as logic expression was
errorneous.

Fixing it, by changing OR into AND, as we want to return `false` if ANY
of the auths were failing, second of all making `auth.apply(obj)` be
executed always, by extracting the statement into separate line.

### DIFF
--- a/lib/auth.js
+++ b/lib/auth.js
@@ -38,7 +38,7 @@ SwaggerAuthorizations.prototype.remove = function (name) {
 };
 
 SwaggerAuthorizations.prototype.apply = function (obj, securities) {
-  var status = null;
+  var status = true;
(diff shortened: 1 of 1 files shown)


webron:
@PASSTor - thanks for the PR. Can you please resubmit against the develop_2.0 branch?

PASSTor:
@webron Just did – PR #515

webron:
Thanks. Closing this one.

PR #510, opened 2015-06-26
allow host to be specified

pm100:


Gillingham:
+1 To this, needed when using Swagger-UI against an API with a dynamic API host.

webron:
@Gillingham - just want to make sure you know that you don't have to specify the `host` and as long as the Swagger definition is part of the application and is affected by the dynamic API host as well, it should work as expected.

fehguy:
Hi, this doesn't really solve the problem.  You'll need to iterate over each operation to set the host since they are copied to each operation.  See #532

PR #519, opened 2015-07-08
Use traditional for instead of for-in in Resolver.countUnresolvedRefs().

morvans:
for-in syntax breaks by "finding" unwanted indices such as Array methods ("equals", in my case).

In my setup I was not able to use swagger-js at all in any browser (tested under FF & Chrome).


webron:
Thanks for the PR. Can you please resubmit it against the develop_2.0 branch? Master is stable and we don't accept PRs against it.

PR #520, opened 2015-07-08
 Use traditional for instead of for-in in Resolver.countUnresolvedRefs(). 

morvans:
Follow up of #519.

for-in syntax breaks by "finding" unwanted indices such as Array methods ("equals", in my case).

In my setup I was not able to use swagger-js at all in any browser (tested under FF & Chrome).


morvans:
Nevermind...

PR #523, opened 2015-07-08
Fix swagger-api/swagger-ui#1435

sublimeye:
Fix for an issue reported in swagger-ui project: (swagger DELETE requests changing to incorrect content type) https://github.com/swagger-api/swagger-ui/issues/1435

The code contains a skipped test for DELETE method with a body ( I'm not sure what is the best way to handle this... I'd need to research the structure of json/schemas and update mocks, I'm afraid it might take too long without a hint)

// This pull request is a part of a sample task by @webron.

### DIFF
--- a/test/request.js
+++ b/test/request.js
@@ -158,15 +158,14 @@ describe('swagger request functions', function () {
 
     expect(req.method).toBe('DELETE');
     expect(req.headers.Accept).toBe('application/json');
-    expect(req.headers['Content-Type']).toBe('application/json');
+    expect(req.headers['Content-Type']).toBe(undefined);
     expect(req.url).toBe('http://localhost:8000/v2/api/pet/100');
     expect(req.body).toBe(undefined);
   });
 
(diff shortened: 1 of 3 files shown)


test/request.js
```diff
@@ -158,11 +158,26 @@ describe('swagger request functions', function () {
 
     expect(req.method).toBe('DELETE');
     expect(req.headers.Accept).toBe('application/json');
-    expect(req.headers['Content-Type']).toBe(undefined);
+    expect(req.headers['Content-Type']).toBe('application/json');
```
fehguy: You should not set `content-type` to `application/json` if there is no body.  That will cause many parsers--including express--to give a JSON parse error

sublimeye:
Hey Tony,

I've updated code with another dirty check for "body" together with DELETE method.
Also I updated petstore.json with a body param. We need it for delete requests with body, but also this will show body param for DELETE request in swagger-ui demo, and I'm not sure is it ok :)

Issue #525, opened 2015-07-16
Doesn't support loading YAML from URL source.


NicholasRBowers:
When a request and subsequent response are created (line 29,710), we assume that the response is coming is formatted in JSON, even though we're now supporting YAML.

`err = new Error('Parser is unable to parse the response');` is triggered when creating the Response object.

If the URL parameter points to a YAML file, it's contents are likely to be transferred with the header `application/json`, because YAML MIME types are not very common.

```
request.parse = {
  'application/x-www-form-urlencoded': parseString,
  'application/json': JSON.parse
};
```

Here, we only allow for processing of two MIME types, so we're kind of stuck.  I would add cases for YAML MIME types and use the already implemented `js-yaml` plugin to process these contents.

This is important because currently our `url` parameter can only consume JSON, while the `spec` contents can ONLY be sent in JSON.  Without this change, I'm unsure of how SwaggerUI can support loading and parsing YAML files?

ponelat:
@NicholasRBowers 
Thanks for looking into the mime types.
What we ended up doing is to try and parse the contents of the response, regardless of mime type (as @fehguy says... servers lie).
While this is fine for client side, there might be implications when requesting the file (if we ask for a yaml file, but have Accept: json). Haven't bumped into it yet, but prevention is better than the cure :wink:

mohsen1:
@ponelat Since JSON is YAML you don't need to do any detection for YAML vs. JSON. Always use YAML for parsing, it will work with JSON content also.

ponelat:
@mohsen1 thanks, that's what we do (use js-yaml for everything that is)
I believe that @NicholasRBowers is referring to the Accept header sent? While this won't affect the client (we parse whatever, see what sticks) I'm not sure if there are any implications for the server.
For example, if I send `Accept: application/json` to the server, but I want a yaml file ... the server has the right to send me json, no? Probably won't happen (custom mime types on a swagger spec? too much effort ;) )

mohsen1:
I think you should provide a better Accept header. For example Chrome send  this Accept header:

```
Accept:text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8
```

It prefers `text/html` but if server can't do that it can also work with `application/xhtml`.

So for Swagger UI `application/yaml,application/json` would be a fair Accept header. Although [nobody knows](http://stackoverflow.com/questions/332129/yaml-mime-type) what is the correct mime type for YAML!

[...]

PR #526, opened 2015-07-16
Enabled parsing YAML from URL parameter.

NicholasRBowers:
This commit addresses [this issue](https://github.com/swagger-api/swagger-js/issues/525).


NicholasRBowers:
Sorry, I didn't realize that the file I edited was a compiled file -_-

Which source file compiles to this bit of code?

```
request.parse = {
  'application/x-www-form-urlencoded': parseString,
  'application/json': JSON.parse
};
```

I can't seem to locate it.

ponelat:
I'll see if I can't find it, but you can try grepping for it (text search) on the files under src/js

NicholasRBowers:
Tried that. I wasn't getting ANY hits :/

ponelat:
I think I know why.... it might not be part of swagger at all, it looks like superagent (http lib)
We don't use JSON.parse in swagger, we've replaced it with js-yaml (which handles both json and yaml)

here is your line...https://github.com/visionmedia/superagent/blob/360ca37c0ffa775f3e788a39a109584f1accc286/lib/client.js#L182

fehguy:
since this affects dependent libraries, i'm closing the issue.

PR #537, opened 2015-07-23
Fix: Make collectionFormat work for array parameters when post form data

birbird:
The test need to be improved.
I don't know how to form a 'multi/part' post operation.
The test I added is 'application/x-www-form-urlencoded' actually.


birbird:
Fix test break

cbornet:
:+1:  need this

jeevanrd:
Facing the same issue here.  Any update on this???

dakisa91:
Im using swagger-ui and it seems to me that this bug is present there too.

fehguy:
This should be supported in master now.

PR #539, opened 2015-07-27
Ability to choose scheme on client init..

tommymessbauer:
When an API supports multiple schemes like below, this client arbitrarily picks the first entry.  This PR adds the ability for the client to be initialized with a specific scheme.

```
"schemes": [
"http",
"https"
]
```

To initialize with a specific scheme, do this:

```
var client = require('swagger-client');

var swagger = new client({
  url: 'http://petstore.swagger.io/v2/swagger.json',
  scheme: 'https',
  success: function() {
    // do something
  }
});```


fehguy:
oops, this was supposed to merge into develop_2.0.  Undoing, and will re-apply in a few

tommymessbauer:
Is this PR rejected or just delayed?

PR #548, opened 2015-08-12
Added allOf support to schemaToJSON

nickpeirson:
Currently Swagger UI outputs `{}` for a model defined with `allOf`. I dug into it and identified the underlying issue as it not being supported by the `schemaToJSON` method in Swagger JS.

This PR adds support for `allOf` so the sample JSON rendered by Swagger UI is a composite of the objects embedded under `allOf`. The Swagger Editor, however, appears to display the components still embedded in the `allOf` container. While my current implementation is best suited for my needs, I'm happy to adapt it if I've misinterpreted it.

I've seen there are a few different issues open against `allOf`support, but I wasn't sure which was the best one to relate this PR to.

### DIFF
--- a/lib/schema-markup.js
+++ b/lib/schema-markup.js
@@ -156,14 +156,27 @@ function schemaToJSON(schema, models, modelsToIgnore, modelPropertyMacro) {
     } else if (type === 'object') {
       output = {};
 
-      _.forEach(schema.properties, function (property, name) {
-        var cProperty = _.cloneDeep(property);
+      if (schema.properties) {
+        _.forEach(schema.properties, function (property, name) {
+          var cProperty = _.cloneDeep(property);
 
-        // Allow macro to set the default value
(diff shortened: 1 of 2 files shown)


fehguy:
`allOf` support should be covered by `resolver.js`.  Can you please share a test case that exhibits the behavior without your fix?

nickpeirson:
If I revert `lib/schema-markup.js ` to `master` then the test I've added demonstrates the problem, giving:
```
  1) response types should return a composite JSON sample for a definition including allOf:

      AssertionError: {} deepEqual { id: 0, name: 'string', code: 0, message: 'string' }
      + expected - actual

      -{}
      +{
      +  "code": 0
      +  "id": 0
      +  "message": "string"
      +  "name": "string"
      +}
```
I'm not sure if you meant something else?

fehguy:
Thank you for this and sorry for the delay--we did add generalized support inside the resolver.js

PR #550, opened 2015-08-17
Added back optional path specs

Pchelolo:
This PR is identical to the code we've had in swagger-ui: optional path elements. 

Only `operation.js` is changed manually, other changes reflect grunt build changes
Original PR: https://github.com/wikimedia/swagger-ui/pull/1
Bug: https://phabricator.wikimedia.org/T109311


webron:
I'm afraid we won't be able to accept this PR as optional path parameters are not supported by the spec.

PR #551, opened 2015-08-19
responses containing references to definitions were not …

aaronb-cg:
…being fully resolved when the spec was pulled in over AJAX but was working locally. helpers.simpleRef()'s parsing of the responses was just checking if the ref _started_ with '#/definitions/', not whether it contained it, and when pulled in over AJAX the refs have the URL prepended to them.

This relates to [issue #1423](https://github.com/swagger-api/swagger-ui/issues/1423) in swagger-ui but also affects swagger-js.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -39,8 +39,8 @@ var simpleRef = module.exports.simpleRef = function (name) {
     return null;
   }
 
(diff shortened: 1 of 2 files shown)


fehguy:
this has been resovled in... the resolver.

PR #556, opened 2015-08-25
download non json content

crazySocket:
when you 'GET' resource you know is not of type "application/json" or any text format the resource is passed to browser to handle. For instance if u want to download resource with mime type "application/octet-stream" u want browser to ask user where to save. This commit make both that happen and handles error in case of failure.

### DIFF
--- a/lib/types/operation.js
+++ b/lib/types/operation.js
@@ -699,6 +699,7 @@ Operation.prototype.execute = function (arg1, arg2, arg3, arg4, parent) {
   if (opts.mock === true) {
     return obj;
   } else {
+    var request = obj;
     /*
      * swagger deals nicely with json response. However, in case of other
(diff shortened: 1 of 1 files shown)


fehguy:
Hi, you can't depend on jQuery inside the swagger-client.  This same code is used on node.js in headless deployments.

I suggest looking at the [http.js](https://github.com/swagger-api/swagger-js/blob/master/lib/http.js) code for the `JQueryHttpClient`.  You can add jquery-specific code into that class, but not the general `operation.js` class.

fehguy:
looking at this more, we definitely cannot merge the PR.  We can, though, move the logic into the jquery client.  I'll keep it open for that and try to refactor it into

fehguy:
closing per above comments.

PR #560, opened 2015-09-02
add hasOwnProperty check in resolver

akaravashkin:
I have tried to use swagger-js in my Ember.js project. Swagger fails on reading json and raises "Maximum call stack size exceeded" error. The reason is [Ember's prototype extension](http://guides.emberjs.com/v2.0.0/configuring-ember/disabling-prototype-extensions/). I have added `.hasOwnProperty` check, it solves the problem.


fehguy:
thanks

PR #566, opened 2015-09-06
Removed redundant HTTP calls when resolving locally defined `$ref`s.

### LINKED ISSUE #489: Regression for Top-Level Objects Resolution
Originally posted in [this swagger-ui issue thread](https://github.com/swagger-api/swagger-ui/issues/1358), I have discovered an issue (I _hope_ it isn't a feature, at least) around global parameter resolution.

After pulling in recent builds from each projects' `develop_2.0` branches several days ago, I noticed that the page was taking longer to load than usual. I inspected the console/network panels and found that an initial call to fetch the api was made, followed by 5 more additional calls. It was tolerable and I assumed it was getting fixed, so I let it be. After adding several more paths to my api, I noticed the number grew to 11, significantly slowing the initial page load. None of the UI was rendering until _all_ calls were complete.  I believe I have finally pinned down the issue.

The resolver seems to be making new requests for the entire api each time it encounters a global parameter ref. I was able to pin this problem down to `Resolver.prototype.resolve`. When these global params are being resolved, a custom object is composed for the resolver's internal processing. This method checks for a property named `root` on that object. If it has a URL, a request for that resource is made for the resolver to process. Every global parameter ref made in the body schema points to the URL of the API passed to swagger-ui upon initialization, resulting in multiple calls to the same URL for resolution. With only 10 defined paths in my api response, the page load has slipped to 35 seconds or more. 

Here is some console output for reference:
screen shot 2015-06-13 at 2 23 58 pm

fehguy:
Resolves #489

### DIFF
--- a/lib/resolver.js
+++ b/lib/resolver.js
@@ -188,7 +188,7 @@ Resolver.prototype.resolve = function (spec, arg1, arg2, arg3) {
   // resolve anything that is local
   for(var ii = 0; ii < toResolve.length; ii++) {
     (function(item, self) {
-      if(item.root === null) {
+      if(item.root === null || item.root === root) {
         // local resolve
         self.resolveItem(spec, _root, resolutionTable, resolvedRefs, unresolvedRefs, item);
         processedCalls += 1;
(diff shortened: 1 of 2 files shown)


Smolations:
@fehguy You are tehguy to go to to get things done!  Thank you sir!  :laughing:

PR #568, opened 2015-09-08
Fix for #1542

Cwallice:
Allows different types of request parameters having the same name.

OperationParams is intriduced to solve this issue.
While you can still use old notation to send arguments into operation following format is preferred:
var params = new OperationParams( { 
                                                               security: {yoursecutitysettings}, 
                                                               consumes: {yourconsumessetting},
                                                               header:{
                                                                    "language": "en"
                                                               }
                                                               etc...
                                                           }  );

later you can set and get params by using following methods:
set( parametertype:string, name:string, value:object );
get( parametertype:string, name:string );

and check whether param exists:
exists( paramtype/s:[ string or array of strings ], paramname );


fehguy:
Hi @Cwallice thanks for the PR.  Some pointers and feedback.

First, please use GFM syntax for your comments.  You can make it much easier to read the code examples, etc like such:

```js
var params = new OperationParams( { 
  security: {yoursecutitysettings}, 
  consumes: {yourconsumessetting},
  header:{
    "language": "en"
  }
  // etc...
} );
```

It's also easier to copy/paste the code when it's in a code block.

Next, it's not clear how you're approaching this.  It seems like you're adding a new Object ( `OperationParams`) but it seems like this is changing the library signature.  How does it work for existing implementations?  Is it back compatible?

Cwallice:
@fehguy 
It supports both old and new arguments type. Every method that accepts operation parameters has now argument type conversion to OperationParams prior any other code. That's why tests work as expected without any changes and it's safe to use old signature.

PR #581, opened 2015-09-22
Fix support for "title" tag

bajtos:
Add "title" to the list `reservedClientTags` to prevent overriding of `self.title` string with an operation group object.

Fix https://github.com/swagger-api/swagger-ui/issues/1615

@webron could you please review?

I did not include the result of `gulp build` as the change-set is quite large, but I am ok to do that if you like.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -61,6 +61,7 @@ var reservedClientTags = [
   'supportedSubmitMethods',
   'swaggerRequestHeaders',
   'tagFromLabel',
+  'title',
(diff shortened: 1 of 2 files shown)


bajtos:
Ping, could one of the maintainers review this patch please?

/cc @fehguy

fehguy:
all good, was off the grid last week so apologies for the delay!

bajtos:
Thank you for landing the patch, no need to apologise :)

PR #596, opened 2015-10-21
Use window location for scheme before defaulting to schemes

FoxxMD:
If the window is using https and `this.schemes = ['http','https']` then the browser will fail to load the request because it is insecure. This change addresses that bug and uses the location schema before falling back to the schemes index.

Fixes swagger-api/swagger-ui#1382


FoxxMD:
@ponelat Thanks for the update, I hadn't seen that issue. Should I modify my PR to reflect that behavior or is this being worked on elsewhere (and I should close the PR)?

ponelat:
As far as I know no-one has actually written any code... ping @dalbani ?

dalbani:
No, I haven't written any code yet.
I think the difficulty of solving this bug lies on finding a nice browser-agnostic solution (as `swagger-js` should not be "tied" to such a runtime environment).

@FoxxMD: you wrote in the title of this PR "use window location". But the `location` variable in `client.js` is *not* the browser's `window.location`.
It's the location of the JSON spec file.

There are in fact 3 protocol schemes in this topic:
1. the scheme of the Swagger UI page
2. the scheme of the JSON spec file URL (processed by `swagger-js`)
3. the scheme(s) defined for the API endpoints

In a browser, due the "mixed content" security enforcement, we know that if scheme 1 is equal to `https`, then scheme 2 and 3 *have* to be `https` as well. Otherwise the browser will block the XHRs.

In a browser environment, your fix does work when all the 3 schemes are `https`.
But, at least in theory, one may want to server Swagger UI over `http`, pointing to a spec file hosted over `https` (using CORS)... but with endpoints only supporting `http`. A bit convoluted, I agree :-)
Basically, @FoxxMD, I'm not sure that the modification you made is 100% correct in all cases.

By the way, why shouldn't the choice of protocol be left to the user in Swagger UI?

FoxxMD:
@dalbani I appreciate the insight! 

Yeah my change isn't going to cover everything but it still seems better than choosing whatever schema is first in the list...Doesn't the url check at [on this line](https://github.com/FoxxMD/swagger-js/blob/master/lib/client.js#L269) imply that swagger-ui is operating in a window though?

Your right though the choice should be left to the user, however [the issue](https://github.com/swagger-api/swagger-ui/issues/1247) doesn't seem to be getting much love :(

dalbani:
> Doesn't the url check on this line imply that swagger-ui is operating in a window though?

Nope, I think there's no direct relation. This `url` variable simply contains the URL where the spec file is located / has been downloaded from.
It is then used as a kind of "base URL" -- for instance in the case where no `host` parameter is specified in the spec.

FoxxMD:
Gotcha. Then it looks like a PR may be more relevant on the ui side to fix whatever it preventing the protocol dropdown from appearing when there is more than one scheme present.

fehguy:
I'm going to close this out--we need to address this in the swagger-ui client.

PR #597, opened 2015-10-21
Display model property descriptions in Swagger 1.2 mock signatures

### LINKED ISSUE #533: Nested schemas do not show description in swagger-ui
When working with swagger-ui, I've found that nested schemas do not show the description. While I found the issue in swagger ui, I believe the actual fix is in swagger-js. I was able to track this down in swagger-ui.js to the resolveSchema helper function.

This function is called within the processModel function (which is within the schemaToHTML function in swagger-client.js). The problem occurs at line 2432 in swagger-client.js:

```javascript
// Resolve the schema (Handle nested schemas)
cProperty = helpers.resolveSchema(cProperty);
```

I temporarily put in a couple of console.log statements to print the before and after values of cProperty:

```javascript

console.log("Property Before: ", cProperty);

// Resolve the schema (Handle nested schemas)
cProperty = helpers.resolveSchema(cProperty);

console.log("Property After: ", cProperty);

```

The first console.log gives the following:

```
description: "Some description here"
schema: Object
  $ref: "#/definitions/ModelB"
  __proto__: Object
__proto__: Object
```

where the second log after the resolveSchema gives this:

```
Object {$ref: "#/definitions/ModelB"}
```

Later on in processModel, the following is what shows the description:

```javascript
if (!_.isUndefined(cProperty.description)) {
  html += ': ' + '<span class="propDesc">' + cProperty.description + '</span>';
}
```

Since the description property is stripped off, it sees undefined, and as a result does not show the description. Not sure if there was a reason to strip all properties off here or not, or if only schema needed to be stripped off, but I managed to fix this by making resolveSchema reappend any properties in the old object that isn't named 'schema':

```javascript
var resolveSchema = module.exports.resolveSchema = function (schema) {
  var resolvedSchema = schema;

  if (_.isPlainObject(schema.schema)) {
    resolvedSchema = resolveSchema(schema.schema);

    for(var key in schema) {
      if(key != "schema") {
        resolvedSchema[key] = schema[key];
      }
    }
  }

  return resolvedSchema;
};
```

I did notice that resolveSchema is used in a few other places, so not sure if this is the proper fix, but if it is, or close to anyway, I would be happy to contribute a pull request.

I should also mention that the project is using the 1.2 swagger spec. An example json being returned by the api is as follows (only including what I believe is relavent for brevity):

```
{
	"swaggerVersion": "1.2",
	"models" : {
	  "ModelA" : {
	    "description" : "",
	    "id" : "ModelA",
	    "properties" : {
	      "modelB" : {
		"description" : "Some description here...",
		"required" : true,
		"type" : "ModelB"
	      }
	    }
	  },
	  "ModelB" : {
	    //omitted for brevity...
	  }
	}
}
```

I am new to swagger, so if I am misunderstanding something or missing anything in my post, please let me know.

whitlockjc:
Whenever you had a model property in Swagger 1.2 where the property
itself was a reference, the description was not being rendered when
generating the model's mock signature.

Fixes #533


fehguy:
Thanks @whitlockjc

PR #559, opened 2015-08-28
Fix spec conversion with anonymous array/object types.

STRML:
The || logic here was faulty and always resolved to true. If a
spec came through with a type of 'object', the definition inside
spec-converter was returned as '#/definitions/undefined'.

This manifested in a *lot* of extra calls to the `swagger.json` endpoint when the client booted. For example, before this patch:

```
GET/explorer/resources/chat                2015.08.28 13:51:56.602
GET/explorer/resources/apiKey              2015.08.28 13:51:56.606
GET/explorer/resources/apiKeyConstraints   2015.08.28 13:51:56.608
GET/explorer/resources/execution           2015.08.28 13:51:56.609
GET/explorer/resources/instrument          2015.08.28 13:51:56.611
GET/explorer/resources/order               2015.08.28 13:51:56.612
GET/explorer/resources/orderBook           2015.08.28 13:51:56.617
GET/explorer/resources/position            2015.08.28 13:51:56.619
GET/explorer/resources/quote               2015.08.28 13:51:56.633
GET/explorer/resources/schema              2015.08.28 13:51:56.640
GET/explorer/resources/settlement          2015.08.28 13:51:56.646
GET/explorer/resources/stats               2015.08.28 13:51:56.651
GET/explorer/resources/user                2015.08.28 13:51:56.653
GET/explorer/resources/trade               2015.08.28 13:51:56.655
GET/explorer/resources                     2015.08.28 13:51:56.763
GET/explorer/resources                     2015.08.28 13:51:56.764
GET/explorer/resources                     2015.08.28 13:51:56.766
GET/explorer/resources                     2015.08.28 13:51:56.767
GET/explorer/resources                     2015.08.28 13:51:56.768
GET/explorer/resources                     2015.08.28 13:51:56.840
GET/explorer/resources                     2015.08.28 13:51:56.841
GET/explorer/resources                     2015.08.28 13:51:56.844
GET/explorer/resources                     2015.08.28 13:51:56.845
GET/explorer/resources                     2015.08.28 13:51:56.846
GET/explorer/resources                     2015.08.28 13:51:57.045
GET/explorer/resources                     2015.08.28 13:51:57.045
GET/explorer/resources                     2015.08.28 13:51:57.047
GET/explorer/resources                     2015.08.28 13:51:57.049
GET/explorer/resources                     2015.08.28 13:51:57.050
GET/explorer/resources                     2015.08.28 13:51:57.122
GET/explorer/resources                     2015.08.28 13:51:57.123
[...]

PR #601, opened 2015-10-25
Add support for an req/res Auditor logging

marcellodesales:
This commit adds the support for the Auditor function as specified
on issue #559. This way, every HTTP request/response from the Swagger
client to the requested api-docs can be logged using the user-provided
logger and audit functions.

*	modified:   lib/client.js
- Capturing the auditor function in the set of options for the client.
- Propagating the swagger options to the SwaggerSpecConverter, as it
  will make HTTP Calls calling the SuperagentHttpClient.

*	modified:   lib/http.js
- Propagating the options to the SuperagentHttpClient class so that it
  can be used from the SuperagentHttpClient calls after the spec is
  downloaded and built.
- Creating the properties "auditorEnabled" and "auditor" to reflect the
  client API options used. The log object must be provided.
- Starting a timer for the latency calculation before the request is
  submitted by the SuperagentHttpClient.
- Collecting the end time and creating the headers for latency.
- Adding the logger to the request object as it is done in other
  frameworks.
- Calling the auditor with the parameters.

*	modified:   lib/spec-converter.js
- Adding the options to the constructor as it is propagated from the
  client.
- Calling the SwaggerHTTP with the propagated options.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -150,6 +150,12 @@ SwaggerClient.prototype.initialize = function (url, options) {
     this.ready = true;
     return this.build();
   }
+
+  // Logger for the req/res
+  this.options.audit = {
+      func: options.auditor || undefined,
+      defaultEnabled: options.audit || false
+  };
 };
 
 SwaggerClient.prototype.build = function (mock) {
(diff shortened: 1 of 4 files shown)


fehguy:
Hi @marcellodesales have you seen the responseInterceptor pattern in the client?  I believe that--plus the request signing technique--does this already.  Can you please review those functions and report back where we're lacking in functionality (or documentation)?

fehguy:
We've discussed this and it's my understanding that we have the support as-is.

PR #608, opened 2015-11-01
Remove jQuery to reduce library file size

jacobrask:
This pull request removes the dependency on jQuery, because I don't see any references to it in the documentation and it adds a considerable amount of data to the download size.


ponelat:
There are some legacy ties to jQuery's ajax (which you've seen in the PR). [superagent](https://github.com/visionmedia/superagent) is the primary one, but it can be lacking at times. There wasn't a way to make an OPTIONS call, nor to add `withCredentials=true`.

But this is a good time to raise the issue of removing jQuery entirely. What we can try to do so over the next week (or so), and collect any outstanding issues against superagent, and then deprecate jQuery 

Thanks for the PR 
@webron will know how to make this deprecation work

quintonparker:
yes. library size is enormous. definitely not ready for the web

fehguy:
This has been addressed here:

https://github.com/swagger-api/swagger-js/commit/4ff3c50bcc6f46dc3113f2ec4adc1626142570b3

Issue #580, opened 2015-09-21
useJQuery flag not working (jQuery not properly resolved)


vvanmol:
Hello, 

I'm updating my current application based on Swagger JS so that it uses jQuery.  I added the boolean useJQuery to my SwaggerClient constructor : 

var client = new SwaggerClient ({
    spec: ... , 
    success: function() { ... }, 
    failure: function() { ... }, 
    useJQuery: true
});

By executing this I get the following error in my JS console : TypeError: Cannot set property 'cors' of undefined.

This error is thrown by the http.js file at line 153 : 

jQuery.support.cors = true;

Basically jQuery is not resolved properly.  The only way I've found to fix this is by modifying the require statement for jQuery in the http.js file : 

require('jquery')  does not work.
require('jQuery') fixes the issue.

I'm using swagger-client 2.1.2-M2 but I've seen the same error in 2.1.5.

Could you please fix this ?  Or is there another way for me to fix this without changing the swagger-client source code ?

Thank you !

vvanmol:
Hello, 

Could someone please tell me if I'm doing something wrong when trying to use jQuery or if this is an issue with swagger-js ?

Thanks !

fehguy:
If you want to use jQuery, you'll have to have it loaded when swagger-js loads.  Swagger-UI, for example, does this automatically.  We don't include it in the library because it's not great for node.js environments.

So if you're using a browser, just make sure you load jquery before swagger-js.  And if you're using it headless inside node, well, don't use jQuery :)

vvanmol:
Hello, 

Thank you for your answer.

However you say swagger-js does not include the jquery library, but it is part of the swagger-js package.json file and therefore it is part of the node_modules directory.  The package.json file refers to "jquery": "^2.1.3".

Now I have jquery available in my app, I'm able to use it but it's only resolved with the following require statement : require('jQuery').  The actual require statement used by swagger-js is : require('jquery').  And if I modify the require statement from the swagger-js http.js file it fixes my issue.

How can you explain that I need a different require statement than swagger-js initially define ?  Is there something I'm doing wrong ?

fehguy:
OK I stand corrected--jquery _is_ loaded by the library.  I wish it wasn't.  I will look into this more to figure out why we're having this issue.

fehguy:
OK so I have a pull request to rip out jquery from the library-if you want to use the `useJQuery` flag, it's your job to add it.

[...]

PR #618, opened 2015-11-04
BYOJQ: bring your own jquery.  Not bundling jQuery per #580

fehguy:
This commit removes the packing of jquery inside swagger-js.

When running headless, super-agent is the proper library to use.  When running in a browser, if one wants to use jquery, they can provide their own in the browser scope.  So let's get it out of here.

* Size before jQuery: 2,971,464 / 361,042
* Size after jQuery: 2,290,204 / 276,702

I see no reason why we wouldn't do this.  @ponelat can you review?


ponelat:
Regarding '$' https://api.jquery.com/jquery.noconflict/

ponelat:
So instead of enforcing '$' to be an object, we deprecate (and dummi-ify) the 
`SuperagentHttpClient.prototype.execute = function (obj) {`

something like...
```js

SuperagentHttpClient.prototype.execute = function (obj) {
// normal code...
}

if (!window || !window.jQuery) {
  SuperagentHttpClient.prototype.execute = function (obj) {
     log('Warning: You need to load jQuery onto the page, before you can use it in Swagger-JS')
     return
  }
}
```

PR #627, opened 2015-11-09
Output curl --data in single quotes (instead of double quotes)

kwizzn:
We have a scenario where we have a "!" character inside our JSON request data:
```
{
  "username": "alice",
  "password": "cooper!00"
}
```

Swagger's `asCurl` method renders this as follows:

```
curl -X POST -d "{
  \"username\": \"alice\",
  \"password\": \"cooper!00\"
}" https://example.com/api
```

bash now tries to interpret the "!" here, which results in this error: `-bash: !00: event not found`.

There's no way to escape "!" in double quotes so we suggest the following output:

```
curl -X POST -d '{
  "username": "alice",
  "password": "cooper!00"
}' https://example.com/api
```

Does the trick and is also much easier to read.

Would you accept this PR? It fixes the problem and updates the corresponding 2 tests.

### DIFF
--- a/lib/types/operation.js
+++ b/lib/types/operation.js
@@ -861,9 +861,9 @@ Operation.prototype.asCurl = function (args1, args2) {
     for (key in obj.headers) {
       var value = obj.headers[key];
       if(typeof value === 'string'){
-        value = value.replace(/"/g, '\\"');
+        value = value.replace(/\'/g, '\\u0027');
       }
-      results.push('--header "' + key + ': ' + value + '"');
+      results.push('--header \'' + key + ': ' + value + '\'');
     }
   }
 
--- a/lib/types/operation.js
+++ b/lib/types/operation.js
@@ -876,7 +876,7 @@ Operation.prototype.asCurl = function (args1, args2) {
       body = obj.body;
     }
 
-    results.push('-d "' + body.replace(/"/g, '\\"') + '"');
+    results.push('-d \'' + body.replace(/"/g, '\\"') + '\'');
   }
 
   return 'curl ' + (results.join(' ')) + ' "' + obj.url + '"';
--- a/lib/types/operation.js
+++ b/lib/types/operation.js
@@ -876,7 +876,7 @@ Operation.prototype.asCurl = function (args1, args2) {
       body = obj.body;
     }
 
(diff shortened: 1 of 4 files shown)


ponelat:
Thanks, that does look much better... I think this should be merged..
Do you think we should consider escaping headers as well? https://github.com/kwizzn/swagger-js/blob/single-quotes-in-curl-output/lib/types/operation.js#L864

kwizzn:
@ponelat, you may be right and I will gladly make that change if it helps getting this issue merged.

kwizzn:
OK done, @ponelat. I am confident my changes are ok. I've done lots of manual testing and ```npm test``` passes locally but Travis now seems to fail for a different reason. Looking at the Travis build logs, I see that the same tests are failing in the master branch due to some weird markup order difference involving an opening ```<div>```.

Is that a known issue?

ponelat:
Ugh... html diff'ing is no man's idea of fun... that needs to be resolved, so that we don't hinder PRs.
I must just figure out, which is the correct one... `expected` or `actual` :smile:

ponelat:
This is where failing test comes from... https://github.com/swagger-api/swagger-js/compare/7f0f8c500812...cbc18795a981
@fehguy, do you know what's happening there?

we need to drop raw html diffing, or at least add a clearer diff.

fehguy:
yes, the html should be updated or just ignored for now.  I need to get all this rendering out of the client asap.

ponelat:
Seeing as the its just the html tests, failing... I'm happy to merge, @fehguy ?

fehguy:
go for it

kwizzn:
Thanks guys!

ponelat:
@kwizzn thanks for the PR :smile:

fehguy:
@ponelat @kwizzn is there a reason why we would leave the URL in double quotes?

kwizzn:
I thought about that too and I guess not. In fact, double quotes do not necessarily have to be escaped in a  URL so we might be better off with single quote in this case, too. What do you guys think?

ponelat:
we should escape all of 'em... no reason not to. Mind making the change, @kwizzn ?

kwizzn:
Happy to do it as soon as possible.

Issue #629, opened 2015-11-13
When sending a GET request with a body parameter the content type is not set correctly


mgttlinger:
When sending a GET request with a body parameter the chosen (in swagger-ui) content type is not used. Similar to https://github.com/swagger-api/swagger-ui/issues/1435 the fix should be relatively easy.

I will try to fix this.

mgttlinger:
Fixed by merge

PR #630, opened 2015-11-13
Fix for #629 content type for get with body

mgttlinger:
This should fix issue #629


mgttlinger:
The failing test does not seem to be an issue introduced by me

mgttlinger:
@fehguy Would it generally be better to check if there is a body in the request and then use the selected content type regardless of the http method in use, instead of whitelisting every method individually (which has so far introduces 2 bugs)?

fehguy:
can you explain more?

mgttlinger:
Currently there is an if clause which checks the request method. If it is "post", "put", "patch" it sets the content type from the opts object (if there is one specified). If it is a "delete" it also checks if there is a body which was introduces due to https://github.com/swagger-api/swagger-ui/issues/1435. From my commit it now also does the same if it is a "get" request with a body. My idea was to not check the request method at all but check if there is a body instead.

Issue #599, opened 2015-10-25
Add support to Req/Res Auditors / Loggers


marcellodesales:
Hi there,

The Intuit Services Platform Team uses auditors for every HTTP request/response in the Servers and now HTTP clients. This is a feature we need to track outbound connections from Node.js services. I'm coming from supporting other HTTP clients such as Restify, which offers an "auditor" plugin along with an asynchronous Logger library such as `Winston` or `Bunyan`.

Swagger Client API change
-------

Adding the following properties to the client:

* Auditor: a function that takes the `err`, `req`, `res` objects from the HTTP package, which is the base of the `superagent` module used by `swagger-js`.
* Logger: an instance of Bunyan with a formatted output.

```js
var clientOpts = {
  auditor: auditor,
  log: logger
};
var apis = new SwaggerClient(clientOpts);
```

Swagger Sample
------

This is a sample for the up-coming 2.1.6 version with Promises. The patch can be applied to that branch. I have defined:

```js
"use strict";

var SwaggerClient = require("swagger-client");
var auditor = require("../lib/auditor/");

var ISPLogging = require("sp-logging");
var logger = new ISPLogging({
  console: {
    level: "debug",
  }
}).getLogger("SWAGGER-HTTP-CLIENT");

var clientOpts = {
  usePromise = true;
  url: "http://petstore.swagger.io/api/api-docs",
  auditor: auditor,
  log: logger
};
var apis = new SwaggerClient(clientOpts).then(function(swagger) {
  apis = swagger;
  return apis.pet.getPetById({petId: 7}, {responseContentType: "application/json"});

}).then(function(pet) {
  console.log("Swagger promise resolved pet", pet);
  return apis.store.getOrderById({orderId: 2});

}).then(function(order) {
  console.log("Swagger promise resolved order", order);

}).catch(function(error) {
  console.error("Swagger promise rejected", error);
});
```

Auditor  Function
-------

Although not included in this pull request, the audit function is anything that takes the err, req, and res objects. I'm using the auditor function is based on the Restify's Audit plugin for the server https://github.com/restify/node-restify/blob/5.x/lib/plugins/audit.js and the Restify HTTP Client version https://github.com/restify/clients/issues/14.

The auditor will do the following:

* Properly adjust the Log level according to the HTTP Response Status Code.
* Display the HTTP Request Headers from the HTTP Request/Response.
* The BODY is **not** included in the auditor.
[...]

Issue #591, opened 2015-10-16
Problems adding arbitrary headers through Request Signing


richardkerr:
I'm trying to add arbitrary headers to a request and have been looking at the 'Custom Request Signing' example in the documentation, but I can't see how to register the Signer so that it is including in the processing chain?  
The only option, to me, seems to be to modify SwaggerClient so that the clientAuthorizations are replaced by the new signer, but it feels very wrong.  Help?

I'm coming from swagger-ui, if that makes any difference.

Thanks!

fehguy:
No  need to modify code, you can add the custom signing logic either before initializing swagger-ui:

```
window.swaggerUi = new SwaggerUi({
   url: url,
   dom_id: "swagger-ui-container",
   authorizations: {
     api_key : new SwaggerClient.ApiKeyAuthorization("api_key", "foo", "query")
  },
```

Or after swagger-ui has been initialized (note! This will not help if the initial call requires signing):

https://github.com/swagger-api/swagger-ui#custom-header-parameters---for-basic-auth-etc

richardkerr:
This requires the "api_key" security attribute to be available for the documented operation.  I want to be able to apply an 'authorization' (which is not actually a security related operation) to the object regardless of the documentation.

richardkerr:
Some more info, it seems like the behaviour of the 'authorizations' option has changed a bit between versions.  
In 2.1.6-M2, it behaves like so:
```javascript
  if (options.authorizations) {
    this.clientAuthorizations = options.authorizations;
  } else {
    this.clientAuthorizations = new auth.SwaggerAuthorizations();
  }
```
This allows me to override the default implementation of SwaggerAuthorizations to have the custom signing as is documented.

In 2.1.6 and later the SwaggerClient constructor & initialization does:
```javascript
this.clientAuthorizations = new auth.SwaggerAuthorizations();
...
this.clientAuthorizations.add(options.authorizations);
```
This explains why the security attribute must be present, it's using SwaggerAuthorizations which does all this checking.  But it also means that we can't add an 'authorization' that is not documented.

This change was made in commit a769bc211d849337c5dba56534b97157c6534eb6, possibly an unintended side effect?

richardkerr:
In 2.1.6+ it also, due to the behaviour of SwaggerAuthorizations, is applied when there are _no_ 'security' definitions in the documented operations.

fehguy:
fixed in #632

PR #632, opened 2015-11-16
Added request interceptor pattern

fehguy:
This addresses #599 as well as #591.  With this PR, you can now intercept *all* requests and responses with the same pattern, allowing auditing request flows by creating timers, adding arbitrary headers, etc.  Note unlike the signing mechanisms, this does _not_ require a `security` tag to match a custom signing algorithm.  You should still use the security tags for selective signing of calls, and the interceptors for handling *all* calls.

@marcellodesales please review to ensure this addresses your needs, this implementation is more consistent with the swagger-js project.


marcellodesales:
@fehguy I will take a look at it and make sure it works in our environment! Thanks a lot for considering it!

marcellodesales:
@fehguy Can you consolidate this branch?

fehguy:
sure @marcellodesales just rebased

marcellodesales:
@fehguy perfect! thanks!

lib/http.js
```diff
@@ -48,6 +49,13 @@ SwaggerHttp.prototype.execute = function (obj, opts) {
...
+  var requestInterceptor = function(data) {
+    if(opts && opts.requestInterceptor) {
+      data = opts.requestInterceptor.apply(data);
+    }
+    return data;
+  }
+
   var responseInterceptor = function(data) {
     if(opts && opts.responseInterceptor) {
       data = opts.responseInterceptor.apply(data);
```
marcellodesales: How would I get access to the `req`, `res`, and `err` objects to support #599 ???

test/request.js
```diff
@@ -405,6 +405,69 @@ describe('swagger basePath override functions', function () {
...
+      done();
+    }, function(err){
+      expect(err.obj).toBeAn('object');
+      expect(err.obj.code).toBe(400);
+      expect(err.obj.type).toBe('bad input');
+      expect(err.obj.message).toBe('sorry!');
+      done();
+    });
+  });
+
```
marcellodesales: Can you please add a supporting test case for #599 as you claim this patch solves for it??? I still cannot see how the API has access to the lower-level `req`, `res`, and `err` objects.

marcellodesales:
@fehguy I cannot understand how this would solve #599 ... Could you please address the questions above and add a test case to support your PR? Thanks!

PR #643, opened 2015-11-25
Fix https://github.com/swagger-api/swagger-ui/issues/1780

Atry:


Atry:
@fehguy Could you review this PR?

fehguy:
Thank you. Since the relative path code is fairly complex, can you please help get a test in that fails without your fix?

PR #652, opened 2015-12-04
Remove Q in favor of promise-native-deferred

wbinnssmith:
This stops Q from being bundled with swagger-js, particularly in the browser where page weight is very important.

Currently the browserify bundle weighs about 850k without the sourcemap (huge!), and Q accounts for more than 150k of that. 

before:

screenshot 2015-12-04 00 14 00
screenshot 2015-12-04 00 13 56

Removing Q means that:

1) We don't bundle a Promise implementation with a ton of bells and whistles, none of which are used
2) We don't burden users with a Promise implementation if they're not using Promises
3) Users can polyfill Promises on their own (or use the browser's native implementation) and choose the implementation they prefer without using Q's promises.
4) We save more than 150K in the unminified bundle size!

This is technically a breaking change since users need to be aware of what the environment requires for Promises (they'll have to manually add an implementation), and that `Q` promises aren't returned, meaning the API on the returned promises is smaller, and users may be depending on bells and whistles inadvertently.

[Here are some reasons why this is a better approach](https://github.com/wbinnssmith/awesome-promises/blob/master/README.md#convenience-utilities). If you're happy with the direction of this, I'll gladly document the additional step for promises users in the README.

Thanks for swagger! 

after:

screenshot 2015-12-04 00 19 22

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -20,7 +20,7 @@ var OperationGroup = require('./types/operationGroup');
 var Resolver = require('./resolver');
 var SwaggerHttp = require('./http');
 var SwaggerSpecConverter = require('./spec-converter');
-var Q = require('q');
+var defer = require('promise-native-deferred');
 
 // We have to keep track of the function/property names to avoid collisions for tag names which are used to allow the
 // following usage: 'client.{tagName}'
@@ -98,7 +98,7 @@ var SwaggerClient = module.exports = function (url, options) {
   this.url = null;
   this.useJQuery = false;
   this.swaggerObject = {};
-  this.deferredClient = Q.defer();
+  this.deferredClient = defer();
 
   this.clientAuthorizations = new auth.SwaggerAuthorizations();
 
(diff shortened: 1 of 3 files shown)


fehguy:
hi @wbinnssmith thanks for the PR and description of why this is better.  Since this would be a breaking change, I'm going to flag it for a future release.  While I'd love to save on the file size, I don't want to incur the headache from broken clients right now, since people are expecting promises to be included directly in the library.

cadesalaberry:
@wbinnssmith definitely agree with you. Btw, what did you use to generate the graphs ?

richardcrichardc:
This also has the advantage that unlike Q, native Promises do not swallow exceptions.

jvivs:
This would be a nice update for the next major release 👍 

@wbinnssmith I think that would be `disc`: http://hughsk.io/disc/

fehguy:
we have a giant update coming to swagger-js that will address this.  Closing...

PR #665, opened 2015-12-15
Https resolution based on url

jonocodes:
Updated scheme resolution to use the incoming url even if schemes are defined


fehguy:
Seems like it should check `this.schemes[0]` before `location.scheme` as that is the intent of the API

jonocodes:
Sure, the order can be switched. Is it safe to assume that when you add two schemes in the spec, that the order they are inserted is preserved? This way schemes[0] is deterministic. Just making sure.

fehguy:
yes

jonocodes:
Pull request updated

PR #671, opened 2015-12-29
an additional null check added

evdokim:
This additional check prevents invocation of `indexOf()` on possible null `ref` variable and improves therefore error messaging for malformed yaml files. 

Before:
```
Cannot read property 'indexOf' of null
```

After:
```
Uncaught TypeError: Swagger 2.0 does not support null types
```


fehguy:
Can you please add a test?

evdokim:
@fehguy  I checked it by changing this line in [swagger-ui.js](https://raw.githubusercontent.com/swagger-api/swagger-ui/master/dist/swagger-ui.js) and it works fine there.

Issue #672, opened 2016-01-01
Path templating, multiple segments, and RFC 6570


harobed:
Implement this feature described in https://github.com/OAI/OpenAPI-Specification/issues/291 in swagger-client

webron:
As @fehguy mentioned, what you point to is a proposal and not part of the spec, so we can't implement it yet. Should this be accepted to a future version of the spec, it'll definitely become part of the tooling.

PR #673, opened 2016-01-01
Implement issue 672 « Path templating, multiple segments based on RFC6570

harobed:
Implement issue #672 « Path templating, multiple segments based on RFC 6570 »


fehguy:
Hi, this is a good PR but the spec doesn't support it yet.  So we can't merge this for the time being.

PR #707, opened 2016-02-09
Fixes #701: Allow for async actions before API requests

### LINKED ISSUE #701: Allow for async actions before API requests
Use case: it's good to check for for expired tokens before every API request, and refreshing tokens is an asynchronous operation. Right now I have to have this before every api call:

    checkExpired().then(function() {
        return client.default.createSomething(params);
    });

It would be ideal if I could intercept requests asynchronously.

bendavis78:


bendavis78:
Not sure why travis build is failing, I ran locally and everything checked out.

PR #708, opened 2016-02-19
Fix slowness for all of resolvment

markusz:
Plenty of people are experiencing slowness when using SwaggerUI with complex nested specs [here](https://www.google.de/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#safe=off&q=swagger-ui+slow).

I had the same problem for a project and tracked the issue down to a specific call of `swagger-js`: `resolveAllOf`. This call has several problems:

- It is a recursive call (which is not a problem per se)
- It is always called, regardless of actual usage of `allOf` tags
- **This is the biggest issue:** It is called twice in each recursion (see [here](https://github.com/swagger-api/swagger-js/blob/master/lib/resolver.js#L673-L675) and [here](https://github.com/swagger-api/swagger-js/blob/master/lib/resolver.js#L738-L740) ), causing the call stack to completely explode for nested definitions 

The following screenshots outline the severity of the problem 

## Status Quo

For my (rather simple) api-docs.json, `resolveAllOf` is called **104,000,000** times, resulting in a render time of 4 minutes and a CPU consumption of 90 %.

default

##  Deduplication and disableAllOfResolve = false 

The screenshot below shows the behavior after only removing the duplicated recursion in `resolveAllOf` in https://github.com/swagger-api/swagger-js/commit/be29806f56b7c183c4b09d578b6a880d26c1e2ba

`resolveAllOf` is now called "only" **6584** times,  which is still too often, but does not cause problems like hanging of the whole browser. Load time is okayish with 3 seconds

**This is the proposed / implemented default behavior since it does not affect functionality in any way.**

deduplicatedresolveallcall

##  Deduplication and disableAllOfResolve = true

`resolveAllOf` is now not called at all. This is the best approach if `allOf` is not used in the swagger docs. Reduces load time to < 1s

disableallofresolving

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -199,7 +199,7 @@ SwaggerClient.prototype.build = function (mock) {
         if (responseObj.swagger && parseInt(responseObj.swagger) === 2) {
           self.swaggerVersion = responseObj.swagger;
 
-          new Resolver().resolve(responseObj, self.url, self.buildFromSpec, self);
+          new Resolver(options).resolve(responseObj, self.url, self.buildFromSpec, self);
 
           self.isValid = true;
         } else {
@@ -209,7 +209,7 @@ SwaggerClient.prototype.build = function (mock) {
           converter.setDocumentationLocation(self.url);
           converter.convert(responseObj, self.clientAuthorizations, self.options, function(spec) {
             self.swaggerObject = spec;
-            new Resolver().resolve(spec, self.url, self.buildFromSpec, self);
+            new Resolver(options).resolve(spec, self.url, self.buildFromSpec, self);
             self.isValid = true;
           });
         }
@@ -220,7 +220,7 @@ SwaggerClient.prototype.build = function (mock) {
   if (this.spec) {
     self.swaggerObject = this.spec;
     setTimeout(function () {
-      new Resolver().resolve(self.spec, self.buildFromSpec, self);
+      new Resolver(options).resolve(self.spec, self.buildFromSpec, self);
     }, 10);
   } else {
     this.clientAuthorizations.apply(obj);
--- a/lib/resolver.js
+++ b/lib/resolver.js
@@ -670,9 +670,6 @@ Resolver.prototype.resolveAllOf = function(spec, obj, depth) {
     if(item === null) {
       throw new TypeError('Swagger 2.0 does not support null types (' + obj + ').  See https://github.com/swagger-api/swagger-spec/issues/229.');
     }
-    if(typeof item === 'object') {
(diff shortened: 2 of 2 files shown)


fehguy:
Free load testing!  Seems like your server did well with the 100m calls.

Indeed the `allOf` support can get hairy, and multiple calls do occur when there's a problem resolving something that it's looking for.  I'll take a look at a couple different techniques to support the external references better and review this PR.  Disabling allOf isn't an option, but handling it better and certainly avoiding making extra calls should be the goal.

markusz:
The proposed solution would not change behavior unless it is explicitly  disabled by setting `options. disableAllOfResolve: true`.

A number of APIs I worked on didn't really require the use of `allOf`, so an option to disable this by means of a configuration attribute like the introduced `disableAllOfResolve`-flag might be beneficial for large portion of the users.

I understand if this does not fit your project policy, but think this would be a short termed pragmatic solution without disadvantages until a better and cleaner solution is found and implemented

fehguy:
Thanks--I get it.  Let me see if I can fix it properly first.  The `allOf` support should have zero impact on APIs that don't have it--if that's not the case, it's busted.

fehguy:
Reference #708

fehguy:
@markusz Do you mind testing #709 against your issue?

markusz:
Will do, but probably not before tomorrow

fehguy:
@markusz any chance to try this out?  It should solve your issue.

markusz:
Not yet, but will try it out asap

andi-itagent:
+1

fehguy:
@andi-itagent can you please try out #709 and report back if it addresses slowness that you may be seeing?

markusz:
Hey @fehguy, sorry for the delay but I finally managed to test #709 against my issue, but unfortunatelly it did not resolve the problem (see screenshot).



The problem keeps existing, since your change does not address the root of the performance issues (unnecessary recursive calls) in #708 but a different potential problem.

It seems that my proposed fixed does solve these problems according to andi-itagent

fehguy:
OK thanks. Can you share your definition so I can use that to test & verify things?

markusz:
I will have to anonymize some stuff since it is an internal API for a customer, but I'll see to it as soon as I find time

fehguy:
Thank you, you can also send it via email and I will test it privately.

markusz:
@fehguy sent you the docs file via email

fehguy:
this was fixed in https://github.com/swagger-api/swagger-js/commit/57b156337375b8af9e5c2b107f77d5ef4b96080f

PR #715, opened 2016-02-26
Resolving second level references

mxgr7:
I have created a (still broken) test that tests behaviour that I think should be supported.

The general problem that I am facing is that only some `$ref` fields get resolved. Is there a reason for this? To me it would be more consistent to resolve any `$ref` field that is encountered, no matter where and however deeply nested.


mxgr7:
The test case I provided is not valid. I'll make sure it's valid first.

PR #718, opened 2016-03-01
removed duplicate recursion code

fehguy:
Alternate fix for #708


lib/resolver.js
```diff
@@ -735,8 +749,5 @@ Resolver.prototype.resolveAllOf = function(spec, obj, depth) {
         obj[key] = output;
       }
     }
-    if(_.isObject(item)) {
```
fehguy: This was a duplicate call causing all the issues.

PR #727, opened 2016-03-09
Avoid deleting the default response on swaggerObject

sgtcoolguy:
When processing the list of responses for an operation, if there is a default response, the default response would get removed from the client's swaggerObject (which holds the full JSON spec). This can be avoided by operating on a deep clone of the operation's responses. This response typically gets hung on the Operation object as "successResponse", which is useful for the client, but shouldn't do so at the expense of modifying the underlying spec JSON.

This will likely break some tests that assume that the default response will be removed and moved to Operation.successResponse.

### DIFF
--- a/lib/types/operation.js
+++ b/lib/types/operation.js
@@ -153,7 +153,7 @@ var Operation = module.exports = function (parent, scheme, operationId, httpMeth
     param.responseClassSignature = param.signature;
   }
 
-  var defaultResponseCode, response, responses = this.responses;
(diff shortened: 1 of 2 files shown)


fehguy:
I think this will cause problems in swagger-ui, you'll end up with the "success response" being in the list of alternate responses for the same operation, which is not ideal.

fehguy:
This will be handled in a major update of swagger-js, and accompanying update in the ui project.  Thus, closing this out.

Issue #725, opened 2016-03-09
Forced Async


briananderson1222:
Is there any reason why we are forced to use the SwaggerClient in an asynchronous way? If I have a spec that I am passing in (not from a URL), the result of instantiating a new SwaggerClient should return the client itself. It looks like if you don't pass 'usePromise' this is the case.. but for some reason the code still requires you to provide a 'success' callback in the absence of 'usePromise' in order for it to build the spec.

**I am going to add a PR to highlight the lines I am talking about.**

fehguy:
Per the PR, I think best we close this out in an effort to keep behavior consistent between modes.

PR #728, opened 2016-03-09
Allow synchronous SwaggerClient creation from spec

briananderson1222:
## Changes
- Removed timeout that forces the buildFromSpec to be run asynchronously (looks like this was added so that the success callback could have access to itself)
- Changed the conditions for building the spec to not require the success callback to be defined
- If we are building securities let's not call apply on badly formed SwaggerAuthorization (Please inform me what the desired behavior would be here)
- Success callback now enforces the timeout so that it can see itself
- If we are resolving the spec by url let's only build ourself not only if there was no spec provided but only if a URL was provided
- Added a test case that shows that the api was built synchronously from the spec
- ran ```gulp``` ~~(all tests are passing but for some reason gulp test is erroring out with no information other than "failed to parse JSON/YAML response) -- ran ```gulp build```~~

**Updates**
- Update to check if apply is function. removed comment and am now assuming the original status
- Updated test cases that were pointing to a wrong url which was causing an unintended failure
- Regenerated the swagger client using ```gulp build```

## Tested
- Tested by running ```gulp test``` ~~-- I could use some help here figuring out where the test script is erroring out~~ **Fixed: Now showing 289 passing. See 6f1afa3**
- ```mocha -R spec test/**/*.js``` returns 296 passing tests with no error

PR is in reference to this issue: https://github.com/swagger-api/swagger-js/issues/725

**I'm pretty new to open source contribution but would very much open to feedback -- My team will be using my fork until we can get this resolved because we are trying to use this in the context of an angular application as an injectable service/factory**


lib/auth.js
```diff
@@ -59,7 +59,7 @@ SwaggerAuthorizations.prototype.apply = function (obj, securities) {
 
   _.each(this.authz, function (auth, authName) {
     if(applyAll || _.includes(flattenedSecurities, authName)) {
-      var newStatus = auth.apply(obj);
+      var newStatus = (typeof auth.apply !== 'undefined') ? auth.apply(obj) : true; // How should we handle a security that does not implement a SwaggerAuthorization?
```
briananderson1222: I assume maybe ```typeof auth.apply === 'function'```` might be more appropriate. Should a default case here be true or false if the passed in security is not a 'SwaggerAuthorization' object

fehguy: Yes, that's fair

briananderson1222:
@fehguy Updates have been made which have fixed the exception being thrown (wrong url path in an unrelated test case). I believe this test was previously giving a false positive as it wasn't actually building the API and testing what it claimed to be testing.

briananderson1222:
Any comments or thoughts on adding the ability to generate the SwaggerClient synchronously?

briananderson1222:
@fehguy - Is there any reason why we would not want this functionality?

fehguy:
Well, we had a discussion on this same topic here:

https://github.com/swagger-api/swagger-js/issues/751

So I don't know why we would make this change.

briananderson1222:
@fehguy - There was no comments made here and that discussion was from 6 days ago. Sorry I didn't get a chance to read it.. I will take my complaints there then.

hswolff:
Happy to help get this landed any way that we can!

lib/client.js
```diff
@@ -153,10 +153,8 @@ SwaggerClient.prototype.initialize = function (url, options) {
     this.scheme = options.scheme;
   }
 
-  if (this.usePromise || typeof options.success === 'function') {
-    this.ready = true;
-    return this.build();
-  }
+  this.ready = true;
```
briananderson1222: Instead of removing this check, would it be more comfortable to check if options.url or options.spec are set? It doesn't effect the existing test cases either way

hswolff:
Also FYI I did the most minimal of changes for my needs and AFAIK everything is working https://github.com/swagger-api/swagger-js/compare/master...TodayTix:dev

briananderson1222:
I attempted that fix as well @hswolff but doesn't it break the unit tests? I believe it made the client variable unavailable in the success callback. It did work as far as synchronous spec creation though from what I remember

hswolff:
Yeah, I didn't check the unit tests at all so I believe you that it would break it.

prestonvanloon:
👍

fehguy:
Closing per conversation.  We want to keep the same API and behavior

briananderson1222:
Quite unfortunate.. The same behavior is maintained in this PR just provides additional behavior that was extremely helpful in my Angular application so we could easily have an API factory that could leverage this without forcing all of my service calls to do promise chaining or be wrapped in an async callback just to get this reference. Hopefully this is considered in the future.

PR #733, opened 2016-03-14
add support for react-nativesupport for react-native

ios122:


fehguy:
Hi, this is missing the source changes.

ios122:
@fehguy ?? click "Files changed".The change is : change "require" to "require".Current Version don't can be used in ReactNative.Because,the node Ver,lack of "buff" Module in RN;the web Ver,"require" is conflict to RN.RN diy the behavior of "require". Just try use sawgger-js in RN,then u will understand ME.ENJOY!

fehguy:
no, you cannot change any files in the browser folder.  Edit the files in the `lib` folder, run gulp.  browser files are overwritten.

ios122:
@fehguy I know.I am not good at gulp.I want use swaggerjs  in RN,but it not worked.My solution is change file in browser dir.My solution May be low,but it works.Maybe u can solute this in a better way.To Support RN will be great.The reason cannot used is RN is not a standard Node.

PR #746, opened 2016-03-27
Resolve refs in responses which contain arrays

johnfriz:
At present, the following response block will not currently resolve the Pet ref:

```
responses:
        200:
          description: "successful operation"
          schema:
            type: "array"
            items:
              $ref: "#/definitions/Pet"
```

This PR adds support for automatically resolving these type of references in responses where the $ref is part of an array response.


fehguy:
This PR has resolution as in-line.  I believe that they can remain as references in the `definitions` section instead of as `inline` models.

PR #753, opened 2016-04-11
Adding supportedMethods to the client options.

daedeloth:
This allows users to specify if they want custom HTTP Methods included (like LINK, UNLINK, etc.)


webron:
We can't accept this as it is not supported by the spec. This may change with a future version of the spec (as discussed on IRC).

PR #754, opened 2016-04-14
Added support for proxy with superagent

dotansimha:
Uses superagent-proxy package . 

The proxy is can be set via "options" when creating the client. 
The http module will use the proccess http_proxy variable first, and later will use the proxy will set. 
Also, it is possible to update the proxy settings via "swaggerClient.setProxy()"


fehguy:
I'd like this, but it shouldn't be enabled by default.  Seems best done with a custom http client configuration so users can opt in.

PR #765, opened 2016-04-28
Checking for the presence of an existing query string in the request URL before appending new params

jimwhimpey:
If you have a path defined with an query string then additional params should be prefixed with "&" rather than "?".

Here's an example: http://flickr.github.io/flickr-api-swagger/#!/Public/getFavoritesByPersonID


fehguy:
ok, but having a path with a query param actually is forbidden by the spec.

jimwhimpey:
Yeah. Unfortunately there's no way around it for us.

fehguy:
OK this is harmless so I'll merge, even though it's not really within the spec intention.

jimwhimpey:
😄

PR #775, opened 2016-05-10
Allow formData with application/json consumption

qrpike:
This allows you to have something like this:

```
      consumes:
        - application/json
      parameters:
        - name: content
          in: formData
          description: Text content to be analyzed
          required: true
          type: string
```

Which will generate a JSON body of:
```
{
    "content":"given string for this variable"
}
```


fehguy:
this isn't supported by the specification, so it doesn't make sense to add support for it now.  It _is_ coming in the next version of the spec, though.

PR #780, opened 2016-05-23
urlify function now removes URL fragment before building URL

travishaagen:
In the `urlify` function, if `this.path` contains a URL fragment, it blindly adds a URL query _after_ the URL fragment, which is incorrect.

This fix simply removes the URL fragment, before building the URL, because I don't think a user of swagger-js would have a need for URL fragments.

I use URL fragments in paths used with Swagger-UI, as a work-around to overload endpoints, so it would be great to get this fix in for Swagger UI 2.1.5 to incorporate.

### DIFF
--- a/lib/types/operation.js
+++ b/lib/types/operation.js
@@ -447,7 +447,7 @@ Operation.prototype.getHeaderParams = function (args) {
 
 Operation.prototype.urlify = function (args) {
   var formParams = {};
-  var requestUrl = this.path;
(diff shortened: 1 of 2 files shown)


fehguy:
Hi, thank you for this--can you please add a couple tests then it can be merged?

travishaagen:
@fehguy : added two unit tests, thanks!

fehguy:
Thanks!

PR #812, opened 2016-07-18
Return image or PDF from POST request

dgreenbean:
In order to return an image or PDF from a POST request, it needs to be possible to set the response type to blob.  The data returned also should be the raw binary data instead of a string.


fehguy:
OK I want to get support for downloads in here, but we can't expect the `window` object is available.  The swagger-js project works for browser and browser-less environments, so there needs to be conditional checks for the `window`.

fehguy:
If you want to send an updated PR that is safe for non-browser environments, please feel free.

dgreenbean:
This PR was mainly to highlight the issue and a potential fix.  I don't know Javascript well enough to tackle a real fix myself.

PR #817, opened 2016-07-21
fix wrong logical operator to detect absolute uri

tcaesvk:


fehguy:
This was fixed just before your PR.  thank you

PR #821, opened 2016-07-28
readme demo update id from 7 to 1

wfayeBerman:
the first demo code item was not working with the `petId` being set to `7` but setting this to `1` allows it to work as expected

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ var Swagger = require('swagger-client');
 var client = new Swagger({
   url: 'http://petstore.swagger.io/v2/swagger.json',
   success: function() {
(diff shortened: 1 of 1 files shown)


fehguy:
OK well to be honest, this is because someone deleted that particular pet in the online API.  I don't think updating the documentation to a new `petId` will help.

wfayeBerman:
ok, it made sense that the id was 7 in the second example where there is actual error handling. in my case i was confused why it wasn't working, and i thought this would be helpful for beginners because it could be less likely that item 1 would be deleted.

wfayeBerman:
got it, much appreciated!

PR #827, opened 2016-08-04
Implementing Custom HTTP Headers Support

feedsbrain:
This is my workaround to pass-in custom HTTP Header to the Swagger Client. When the code seems covers the flow, I'm not sure which part that cause the "execute" function executed twice. And on the second call, the HTTP Headers is reset to default. Please review ...

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -63,6 +63,7 @@ var reservedClientTags = [
   'spec',
   'supportedSubmitMethods',
   'swaggerRequestHeaders',
+  'swaggerCustomRequestHeaders',
   'tagFromLabel',
   'title',
   'url',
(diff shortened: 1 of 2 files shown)


fehguy:
What does this do that requestInterceptors do not accomplish?

feedsbrain:
It's just suggestion and it's more convenient to pass-in the custom header from the client. Currently it was also confusing, headers options only applied to "Content-Type" and nothing else.

fehguy:
the `swaggerRequestHeaders` were named such because they're only used to fetch the definition, not for general requests.

I'm going to close this as it's functionality already in the `requestInterceptor` feature.

PR #828, opened 2016-08-04
add test to resolve $ref within allOf construct

aguerotSOVINTY:
Based on issue raised on swagger-ui (#2304) [https://github.com/swagger-api/swagger-ui/issues/2304 ](url), as suggested I added a test case 2 with $ref within an allOf construct that does not resolve correctly within swagger-ui.

### DIFF
--- a/test/resolver.js
+++ b/test/resolver.js
@@ -938,6 +938,35 @@ describe('swagger resolver', function () {
     sample = new SwaggerClient(opts);
   });
 
+  it('should resolve external references within allOf', function (done) {
+    var sample;
+    var opts = opts || {};
+    opts.url = opts.url || 'http://localhost:8000/v2/externalRefInAllOf.json';
+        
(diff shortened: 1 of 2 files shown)


aguerotSOVINTY:
Failed test is intended. This PR is about adding a new test case for $ref constructs with an allOf array. See test/spec/v2/externalRefInAllOf.json file provided.

fehguy:
Looks like you have a bug in the tests.  There is no `allOf` in this definition, and therefore no `error` property in the `OrderQueryGetResponse` definition.

Perhaps you had a commit missing?  I've corrected the test and will merge.

aguerotSOVINTY:
Yes indeed there is a commit missing…
Do I have to create a new PR for this or can I just commit and push to my repo?

Tx

fehguy:
Since the PR is closed, please create a new one.

PR #829, opened 2016-08-05
Fix {$ref: "/domain/based/path/"} resolving

neumond:
My spec heavily uses refs like `/api/1/something.json` without domain part, but not relative like `./some.json` or `../../some.json`. Before this fix swagger UI did fail with exception and become unusable.

### DIFF
--- a/lib/resolver.js
+++ b/lib/resolver.js
@@ -483,7 +483,7 @@ Resolver.prototype.finish = function (spec, root, resolutionTable, resolvedRefs,
 
         for (key in resolvedTo.obj) {
           var abs = resolvedTo.obj[key];
-          
+
           if (localResolve !== true) {
             // don't retain root for local definitions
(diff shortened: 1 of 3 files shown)


neumond:
Regarding test failure, it's reproducible even on master:
```
git clone https://github.com/swagger-api/swagger-js.git
cd swagger-js
npm install
./node_modules/.bin/gulp browsertest
```
>Firefox 47.0.0 (Linux 0.0.0) yaml http superagent with promise should call the catch-function when executing an invalid api-call FAILED

fehguy:
OK so the tests are passing fine, however to merge this I'll need a test to confirm that this behavior is working and doesn't regress with other changes.  Can you please add so we can merge?

neumond:
Whoah, it was not easy. RetainRoot contained another buggy place which did fire during writing a test. Tested additionally against my spec, works fine.

PR #833, opened 2016-08-10
Update schema definitin post PR 828

aguerotSOVINTY:
Update schema definitin post PR 828


fehguy:
what is this PR supposed to do?

aguerotSOVINTY:
Found something that looks like a bug when defining an entity that uses an allOf construct with multiple $ref entries.
It is linked to test 'should resolve external references within allOf' in /test/resolver.js

fehguy:
This isn't a bug in the test, closing

PR #846, opened 2016-09-02
Timeout support

jvivs:
Adds support for configuring timeouts on spec and operation requests.

`timeout` option:
- configured in `ms` for requests at the client and operation level
- pass as option when creating a client to set a timeout for all requests
- pass as an option when executing an operation to set a timeout for a specific request

`fetchSpecTimeout` option:
-  option will override `timeout` when specified
- when a value is passed to`timeout`,  pass `null` or `false` to for `fetchSpecTimeout` use agent default when fetching a spec

Will fail with an informative message when a timeout occurs (tested with `superagent` in unit tests and manually in the browser with jQuery's `$.ajax`)

Possible nice-to-haves in the future:
- `x-timeout` vendor extension to specify timeouts at the spec, route, or operation level

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -138,7 +138,13 @@ SwaggerClient.prototype.initialize = function (url, options) {
   this.modelPropertyMacro = options.modelPropertyMacro || null;
   this.parameterMacro = options.parameterMacro || null;
   this.usePromise = options.usePromise || null;
-  this.timeout = options.timeout;
+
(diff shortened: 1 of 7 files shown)


oeiriksson:
:+1:

fehguy:
looks good to me

jvivs:
👏 thank you for the quick turnaround!

PR #847, opened 2016-09-04
Topic/preserve slashes option

holophrastic:
We have an issue where one of the parameters to an endpoint can be a path; for example, the spec is defined as

```
  "paths": {
    "/{foo}/{bar}": {
    ...
```

but bar itself can be of the format 'qux/baz.txt'. This means the slash gets urlencoded, which breaks things.

This patch allows the client to take a preserveSlashes option, so that qux/baz.txt stays intact, instead of coming out as qux%2Fbaz.txt.

Thanks for your consideration.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -137,6 +137,7 @@ SwaggerClient.prototype.initialize = function (url, options) {
   this.defaultErrorCallback = options.defaultErrorCallback || null;
   this.modelPropertyMacro = options.modelPropertyMacro || null;
   this.parameterMacro = options.parameterMacro || null;
+  this.preserveSlashes = options.preserveSlashes || false;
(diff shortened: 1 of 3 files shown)


fehguy:
Hi, I hate to say this but we can't accept this PR as it.  The spec requires that path parameters are escaped or the paths may be indeterminate.

_However_, this is being tossed around for support in the 3.0 spec, and I can suggest how you can make this in, without encouraging breaking the spec.  Consider _not_ escaping based on an extension in the parameter, like such:

```
in: path
name: yourParam
x-escape: false
```

Then you're adding functionality, not encouraging breaking the spec, etc.

holophrastic:
Booooo.

Makes sense, though. Thanks for taking the time to look at it.

PR #850, opened 2016-09-09
Incorrect handling of array values in multipart/form-data environment

xoss:
We have experienced this erroneous behavior since v2.1.4 when sth. significant changes in the `operation.js:getBody` method have occured. Since then, every array-like value in form-data was incorrectly serialized.
For instance, having a spec like this
```json
...
   {
     "name": "lang",
     "in": "formData",
     "type": "array",
     "collectionFormat": "csv",
     "items": {
       "type": "string",
         "enum": [
           "de",
           "en"
          ]
       }
    },
   {
     "name": "country",
     "in": "formData",
     "type": "array",
     "collectionFormat": "multi",
     "items": {
       "type": "string",
         "enum": [
           "DE",
           "US"
          ]
       }
    },
    {
      "name": "file",
       "in": "formData",
       "type": "file"
     }
...
```
would lead to a curl request much like this:
```bash
curl -XPOST 
  --header 'Content-Type: multipart/form-data'
  --header 'Accept: application/json' 
  --header 'Accept-Version: *' 
  --header 'X-Store-Upload: false' 
  --header 'X-API-Key: <YOUR_API_KEY>' 
  -F lang=lang=deu&lang=eng 
  -F country=country=DE&country=US 
  -F file=@"file.jpg" 
  'http://localhost/'
```
1. The `collectionFormat` was ignored, respectively the `collectionFormat` of the last param in the list was taken which could've been `undefined`, just like in this case.
2. The `encodeQueryCollection` was added to the `bodyParam: FormData` object, which itself would create an `key=value` serialization on submit, leading to faulty values such as `lang=lang=deu&lang=eng` and `country=country=DE&country=US`.
3. The documentation says, the default in case no `collectionFormat` is `csv`. This is, also, not correct, as it falls back to `multi` in this case, as can be seen from serialization of `lang`

It was quite tough to reproduce this behavior in the test environment, because `FormData` in `operation.js` is not defined in test environment which let `getBody` always go a different path.


xoss:
Frankly, I wasn't able to get the test scenario running without breaking the production scenario. In that case, I'd choose for the latter to be working.

lib/types/operation.js
```diff
@@ -599,7 +600,7 @@ Operation.prototype.getBody = function (headers, args, opts) {
             bodyParam.append(key, value.value);
           } else {
             if (Array.isArray(value)) {
-              bodyParam.append(key, this.encodeQueryCollection(param.collectionFormat, key, value));
+              bodyParam.append(key, this.encodeQueryCollection(param.collectionFormat, key, value).split('=').slice(1).join('='));
```
fehguy: I'm not sure why we would split like this.  Seems like this would be undoing something that we did already?

xoss: Well, the problem, it would duplicate the `key` in the first place.
See the serialization of the `bodyParam` object.
```js
expected: lang=deu,eng
actual:  lang=lang=deu&lang=eng 
```
and
```js
expected: country=DE&country=US
actual: country=country=DE&country=US 
```

fehguy:
In testing this, looks like the curl output and actual execution is different.  I'll work on massaging both to support the different collection formats.

fehguy:
ok @xoss I've gone through this.  The bad news is, this change causes problems.  The execution and curl commands are out of sync, and the other `collectionFormat`s are not correct.  For example, it's perfectly legal to serialize an array of strings in CSV format, where the multi-part post is just a single CSV string.

The good news is I've added support for the `multi` format (the default serialization format for arrays in Swagger 2.0) to do the Right Thing ™.

See #858

PR #840, opened 2016-08-23
make header parameters case insensitive

NVP12345:
Allows header parameters to be passed insensitive to case, conforming to [RFC 2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2)

PR #851, opened 2016-09-13
Headers should be case insensitive

cesine:
Hi @fehguy we've been using swagger-client for about a year, its a great project! We noticed that our headers weren't being passed if they didn't match the case of the swagger spec.

I added some docs to #840 Nick's PR, let us know if there is anything else we can do to help get this merged in!

http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html
```
HTTP header fields, which include general-header (section 4.5), request-header (section 5.3), 
response-header (section 6.2), and entity-header (section 7.1) fields, follow the same generic format 
as that given in Section 3.1 of RFC 822 [9]. Each header field consists of a name followed by a 
colon (":") and the field value. Field names are case-insensitive.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,6 +224,46 @@ You can add it to the swagger-client like such:
 client.clientAuthorizations.add('my-auth', new CustomRequestSigner());
 ```
 
+### Setting headers
+
+Headers are a type of `parameter`, and can be passed with the other parameters. For example, if you supported translated pet details via the `Accept-Language` header:
+
+```js
+"parameters": [
+  {
+    "name": "petId",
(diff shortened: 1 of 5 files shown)


fehguy:
Hi I merged #840 so if you want to move any other tests or logic from this PR into a new one, I'd be happy to add.

cesine:
will do, its mostly just docs. 

thanks @fehguy we really appreciated it!

cesine:
@fehguy I swashed and rebased and pushed to this branch if you want to re-use this PR you can re-open it

fehguy:
@cesine looks like it won't let me reopen it :(

image

PR #852, opened 2016-09-13
The param is not given value when loop all formParams

yuxiaomin:
The variable param should be given value when looping the formParams.

### DIFF
--- a/lib/types/operation.js
+++ b/lib/types/operation.js
@@ -589,6 +589,7 @@ Operation.prototype.getBody = function (headers, args, opts) {
       bodyParam.type = 'formData';
 
       for (key in formParams) {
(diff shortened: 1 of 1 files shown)


fehguy:
It doesn't need to be, because it's not used inside that loop?

yuxiaomin:
@fehguy the param is used. In Line 608
`bodyParam.append(key, this.encodeQueryCollection(param.collectionFormat, key, value));
`
to get the collectionFormat of param. So it is necessary otherwise param is not correct.

fehguy:
hmm so it is, i missed that!

xoss:
Already part of #850

fehguy:
Merged with #858

PR #857, opened 2016-09-14
added scheme preference to `window`

### LINKED ISSUE #841: location.scheme should have priority over those declared in the spec.
My swagger spec declares http and https, in that order.

When I access my app over over TLS, swagger-js attempts to use http and so the browser stops it from doing so. Can we make it so that it gives priority to location.scheme instead?

fehguy:
Fixes #841


fehguy:
OK this does address #841 but it does not override the specified protocols to favor the `window.location.protocol` value.  So it _is_ still possible to have mixed content per https://github.com/swagger-api/swagger-ui/issues/1670 if the spec defines the `schemes` but does not include the `https` protocol, for example.  

* If the `schemes` is left empty, the `window.location.protocol` is used.
* If the `schemes` has two values, `[http, https]`, the matching value from the protocol will be prefered.

PR #864, opened 2016-09-23
fix the scheme extraction for the API calls

jukart:
Changes the way how the protocol for the API calls is calculated.
swagger specification defines that the protocol usade for the API calls
should be the protocol used to get the specification.

The swagger client implementation allows to provide the protocol in the
scheme option. If this option is set the protocol is always forced to
this option.

If the url to request the spec file contains no protocol
window.location.protocol is used or 'http' if no window object is
available.


fehguy:
I believe this was fixed by https://github.com/swagger-api/swagger-js/pull/866

PR #866, opened 2016-09-26
Fixes incorrect protocol bug

hsingh23:
The protocol is incorrectly assigned here - window.location.scheme is not a thing. Please create a new release as v2.1.19 is broken with a bug where a request sent from a https swagger-ui is calling an http endpoint which the browser disallows because of security reasons.


hsingh23:
This bug was introduced here: https://github.com/swagger-api/swagger-js/commit/535d505cc40314be5b0d010796cc9431d9d8ee3a

Issue #877, opened 2016-10-11
obj.body.keys is not a Function


Pchelolo:
When POSTing form-data I'm getting `Uncaught TypeError: obj.body.keys is not a function` in 
Safari and certain versions of Chrome from this line: https://github.com/swagger-api/swagger-js/blob/master/lib/http.js#L248

The problem is that `FormData.prototype.keys()` doesn't exist in Safari, so it needs to be either polyfilled or removed.

Pchelolo:
The fix doesn't actually fix anything. FormData.keys() is supposed to return an iterator, while _.keys returns an array, so the error just moves a little bit further.

fehguy:
yep this slipped through.  The tests were using www-url-encoded so this was missed.  Thanks!

PR #879, opened 2016-10-11
Use superagent.send to avoid mangling FormData

Pchelolo:
The merged fix for https://github.com/swagger-api/swagger-js/issues/877 doesn't work since actually in older chrome, safari and IE there's no way to get access to FormData contents ( http://stackoverflow.com/a/33821488/1924141 ) however `SuperAgent` supports getting a ready FormData, so could we just give it to it?

This fixes basic POST support in Safari and older chrome, but I'm not sure if it breaks some other use-cases since the `FormData` branches seem to be not covered with unit tests

### DIFF
--- a/lib/http.js
+++ b/lib/http.js
@@ -245,33 +245,7 @@ SuperagentHttpClient.prototype.execute = function (obj) {
       if (contentType.indexOf('multipart/form-data') === 0) {
         delete headers['Content-Type'];
         if({}.toString.apply(obj.body) === '[object FormData]') {
-          var itr = _.keys(obj);
-          var p = [];
-          while(true) {
-            var v = itr.next();
-            if(v.done) {
-              break;
-            }
-            var key = v.value;
-            // only once
-            if(p.indexOf(key) === -1) {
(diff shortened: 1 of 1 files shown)


fehguy:
I don't see an issue for this in Safari Version 10.0 (11602.1.50.0.10) or in IE 11.0.9600.17416.  I'm testing the spec at http://petstore.swagger.io/v2/swagger.json

I'd like to have this fixed so I can do a release, so please send more information on how to reproduce the issue.

Pchelolo:
@fehguy Here's an example: https://en.wikipedia.org/api/rest_v1/#!/Transforms/post_transform_wikitext_to_html_title_revision

Add whatever text to the wiki text parameter and hit `Try it out` in Safari 10 - you'll see the error

fehguy:
thank you, I'll test with that spec.

fehguy:
@Pchelolo I've reproduced this, working on a fix.  I'm looking into why we aren't using superagent's form send, there was a good reason at the time...

PR #889, opened 2016-10-27
Adds ability to override request agent

danteoh:
Even though the current API supports passing in a different http client, you may want to just use a different instance of superagent.  For our use case, we have a number of plugins that get wired up into superagent so we want to use that instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ var client = new Swagger({
 });
 ```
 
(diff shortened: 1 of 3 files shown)


djMax:
I'm confused about this PR - it will replace ALL future swagger client requestAgents with the option that was passed to the options of one of them. That seems gross.

Issue #893, opened 2016-11-03
Handling File objects


alex-seville:
I'm trying to POST a `File` (https://developer.mozilla.org/en-US/docs/Web/API/File) from an HTML file input control.

However, this block of code (https://github.com/alex-seville/swagger-js/blob/master/lib/http.js#L243-L275) interprets the `File` object like a JS object and causes it's data to be lost.  

Am I mistaken in thinking that this line (https://github.com/alex-seville/swagger-js/blob/master/lib/http.js#L244) should read:

```js
 if(_.isObject(obj.body) && !obj.body instanceof File) {
```

?

alex-seville:
@fehguy ?

alex-seville:
Closing as per PR comment.

PR #897, opened 2016-11-11
Don't treat File interface like an object in the browser

alex-seville:
Files don't need to be attached as FormData, or sent as multipart/form-data.
This conditional assumes that's the case, but for an `application/octet-stream`, for instance, this code with stringify the body resulting in an invalid payload.

I'm proposing to ignore this logic if a File interface is passed as the `obj.body` since it will erroneously be identified as an `object` by `isObject`, but it needs to be handled differently from JS objects.

Related: https://github.com/swagger-api/swagger-js/issues/893

### DIFF
--- a/lib/http.js
+++ b/lib/http.js
@@ -221,7 +221,7 @@ SuperagentHttpClient.prototype.execute = function (obj) {
   }
 
   var r = request[method](obj.url);
-  
+
   if (timeout) {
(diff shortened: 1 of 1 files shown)


alex-seville:
Nevermind, I can pass the file as FormData which seems to work with this conditional.

PR #898, opened 2016-11-13
Fixed unnecessary scopes mutation

frol:
The issue due to the mutated scopes popped up in Swagger-UI: https://github.com/swagger-api/swagger-ui/issues/2483

I am not sure if this is a **correct** solution, but it works for me.

/cc @fehguy @janslow

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -310,7 +310,7 @@ SwaggerClient.prototype.buildFromSpec = function (response) {
       for(var ext in securityDefinition) {
         helpers.extractExtensions(ext, securityDefinition);
         if (ext === 'scopes') {
-          var scopes = securityDefinition[ext];
+          var scopes = _.cloneDeep(securityDefinition[ext]);
           if(typeof scopes === 'object') {
             scopes.vendorExtensions = {};
             for (var s in scopes) {


frol:
This attempt seems to be wrong as the tests explicitly expect that `vendorExtensions` will be mounted. Thus, I think, the iteration over the list of scopes should be fixed in Swagger UI project.

janslow:
Overriding `vendorExtensions` on the scopes object could also cause issues if an API has a scope called `vendorExtension`. A safer property name for vendor extensions would be to prefix that name with `x-` (e.g., `x-vendorExtensions`) or use a symbol as the property name (which probably isn't ok due to compatibility with old JS engines).

Another option, which would solve this bug would be to use `Object.defineProperty` to make `vendorExtensions` non-enumerable.

Finally, is there a reason to aggregate all of the vendor extension properties into a single object, as opposed to just adding each of them to the main object?

frol:
Just for the record, this PR was merged and 5 hours later, it was reverted in https://github.com/swagger-api/swagger-js/commit/d916f52db709c8a6bfd2557960fd4c7fdcac317a

fehguy:
The fix will be addressed in a Swagger-UI patch, and believe it or not, for good reason.  The patch supplied here didn't make sense.  See https://github.com/swagger-api/swagger-ui/commit/30f0848ac20d510593aab7238070aa5dae08ebb6

Pull master and enjoy

frol:
> The patch supplied here didn't make sense.

I agree and I pointed this out in my second comment here. I didn't expect it would be merged (and I also stated that), so I was just surprised that it was merged to be reverted a few hours later. I wrote the last comment just so people don't get confused.

Thank you for the proper fix!

Issue #773, opened 2016-05-09
Maximum call stack exceeded between 2.1.10 and 2.1.11


djMax:
Same swagger docs in both cases. It has something to do with definitions, as I can fix it by removing them all, but even a single one, seemingly no matter how simple, will fail.

```
/myproject/node_modules/swagger-client/lib/resolver.js:523
      var o = this.getRefs(item);
                   ^

RangeError: Maximum call stack size exceeded
    at Resolver.getRefs (/myproject/node_modules/swagger-client/lib/resolver.js:523:20)
    at Resolver.getRefs (/myproject/node_modules/swagger-client/lib/resolver.js:523:20)
    at Resolver.getRefs (/myproject/node_modules/swagger-client/lib/resolver.js:523:20)
    at Resolver.getRefs (/myproject/node_modules/swagger-client/lib/resolver.js:523:20)
    at Resolver.getRefs (/myproject/node_modules/swagger-client/lib/resolver.js:523:20)
    at Resolver.getRefs (/myproject/node_modules/swagger-client/lib/resolver.js:523:20)
    at Resolver.getRefs (/myproject/node_modules/swagger-client/lib/resolver.js:523:20)
```

fehguy:
Hi, can you please share a spec?  We have a lot of tests around this but would like to see what's triggering this issue.  FYI we're on 2.1.14 now.

djMax:
Yeah, that's what's confusing the hell out of us - this was working a week ago, and 2.1.10 is much older. Working on isolating such that we could post some reasonable spec.

fehguy:
OK happy to help if you can share a definition, even privately.  I'm currently working on a release of swagger-js right now, so it could be fast.

djMax:
```
{
  "swagger": "2.0",
  "info": {
    "version": "0.0.1"
  },
  "paths": {
    "/token": {
      "post": {
        "description": "Create a token for sensitive payment information"
      }
    }
  },
  "definitions": {
    "Foobar": {
      "type": "object",
      "required": [
        "token"
      ],
      "properties": {
        "token": {
          "type": "string"
        }
      }
    }
  }
}
```

djMax:
Here's the other trick - you have to load two swagger docs, of which this should be one.

fehguy:
Then I don't really understand.  Can you explain that more?  Perhaps with a test?

djMax:
[...]

Issue #29, opened 2013-04-15
RangeError: Maximum call stack size exceeded


jplock:
swagger-ui is able to fetch my individual API endpoint JSON documents, but when it starts parsing the models, it never displays anything and Chrome shows:

```
Uncaught RangeError: Maximum call stack size exceeded
SwaggerModelProperty.getSampleValue
SwaggerModel.createJSONSample
SwaggerModelProperty.getSampleValue
SwaggerModel.createJSONSample
SwaggerModelProperty.getSampleValue
SwaggerModel.createJSONSample
SwaggerModelProperty.getSampleValue
SwaggerModel.createJSONSample
SwaggerModelProperty.getSampleValue
SwaggerModel.createJSONSample
SwaggerModelProperty.getSampleValue
...
```

in the console.  I'm using https://github.com/ryankennedy/swagger-jaxrs-doclet to generate my documentation off of a Dropwizard Java service.  The JSON files it outputs are correct, so I'm not sure why the UI is looping.  This is using the latest checkout of swagger-ui from master.

PR #874, opened 2016-10-06
Path parameter loops would include functions in the prototype to parameters array

alechirsch:
Although it looks like a lot has changed, I have only added 6 lines to the code. The rest of the changes are because of indentations.

I needed to add a conditional for three of the for-in loops that tests for hasOwnProperty().
My use case for this is that I am using Swagger-UI in an Ember application, and Ember automatically converts all arrays to Ember arrays. Ember arrays have a few dozen functions attached to the prototype, which were being caught by the for-in loop.

The result of this displayed all of the Ember array function names in the parameters list for every path.

PR #899, opened 2016-11-16
Use Array.forEach instead of for...in

seb0zz:
Had the problem that in the application where I use this library all `Array` objects get enriched with some extra methods. So when using the `for...in` I always had the problem hat also these extra methods where added to the `parameters` array leading to a recursion problem in the end (like in #29 or #773). Would be great if you could merge this. I could also bring this to the browser version. Then we would need to check if the `Array.forEach` method is available before (e.g. IE8 does not have it).


alechirsch:
I also had this issue #874, waiting for response

PR #900, opened 2016-11-21
Fix errors in embedded sourcemaps

richvdh:
The sourcemaps in the browserified libraries were incorrect due to the addition
of the header. We can fix this using gulp-sourcemaps to export correct
sourcemaps.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -9,6 +9,7 @@ var header = require('gulp-header');
 var istanbul = require('gulp-istanbul');
 var jshint = require('gulp-jshint');
 var mocha  = require('gulp-mocha');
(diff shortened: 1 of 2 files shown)


fehguy:
@richvdh looks like the build doesn't like es6 syntax.  Can you address and update the PR?

richvdh:
Thanks for the feedback: looks like a dependency of gulp-sourcemaps (strip-bom) is using the arrow notation.

For now I've pinned gulp-sourcemaps to 1.x, which claims compatibility with node 0.10. Obviously an alternative would be to bump the version of node requested by travis.yml:  0.10 is pretty old these days [1].

[1] https://github.com/nodejs/LTS#lts-schedule

PR #916, opened 2017-01-01
Pass request to response interceptor, cleanup operation object passed to them both

djMax:
My previous PR passed "operation.operation" but that's missing some important stuff like the client itself. So now it passes "operation" to the request/response interceptors. Additionally, the response interceptor receives and argument for the original request, which allows things to pass between those two handlers without requiring a new closure every time.


lib/http.js
```diff
@@ -73,14 +73,14 @@ SwaggerHttp.prototype.execute = function (obj, opts) {
 
   var responseInterceptor = function(data) {
     if(opts && opts.responseInterceptor) {
-      data = opts.responseInterceptor.apply(data);
+      data = opts.responseInterceptor.apply(data, [obj]);
```
fehguy: Why set `obj` in an array?

djMax: Sorry I didn't reply to this - apply's second arg is an array which is "spread" to the function, so essentially the function gets one argument which is obj, not an array.

PR #938, opened 2017-01-12
shouldnt remove 0 values from query params

cesine:
Hi guys, 

We pulled in the latest version 2.1.31 and found a breaking change from 2.1.2x

It seems to be converting `bar=0` into `bar=`, I added a test to show with a sample parameter and resulting url 

```
$ npm test

> swagger-client@2.1.31 test /Users/gina/git/swagger-js
> gulp test

[12:52:09] Using gulpfile ~/git/swagger-js/gulpfile.js
[12:52:09] Starting 'test'...


  SwaggerClient
    1) doesnt remove 0 from query params


  0 passing (85ms)
  1 failing

  1) SwaggerClient doesnt remove 0 from query params:

      AssertionError: 'http://localhost:8000/double/foo?bar=' === 'http://localhost:8000/double/foo?bar=0'
      + expected - actual

      -http://localhost:8000/double/foo?bar=
      +http://localhost:8000/double/foo?bar=0
```

If the value is not falsy, (for example 1) the test passes we found that https://github.com/swagger-api/swagger-js/pull/935/files#diff-e9afd75580721a5b884cc82272c23984R1245 is cleaning falsey values

```
$ npm test

> swagger-client@2.1.31 test /Users/gina/git/swagger-js
> gulp test

[12:52:53] Using gulpfile ~/git/swagger-js/gulpfile.js
[12:52:53] Starting 'test'...


  SwaggerClient
    ✓ doesnt remove 0 from query params


  1 passing (80ms)
```

### DIFF
--- a/lib/types/operation.js
+++ b/lib/types/operation.js
@@ -1242,7 +1242,7 @@ Operation.prototype.encodeQueryParam = function (arg, maskPasswords) {
   if(maskPasswords) {
     return "******";
   }
-  if(arg) {
+  if(arg !== undefined && arg !== null) {
     return encodeURIComponent(arg);
   }
   else {
--- a/test/client.js
+++ b/test/client.js
@@ -1619,6 +1619,41 @@ describe('SwaggerClient', function () {
     });
   });
 
+
+  it('doesnt remove 0 from query params', function(done) {
+    var spec = {
+      basePath: '/double/',
+      paths: {
+        '/foo': {
+          get: {
+            tags: [
+              'test'
+            ],
+            operationId: 'slash',
+            parameters: [{
+              in: 'query',
+              name: 'bar',
+              type: 'integer',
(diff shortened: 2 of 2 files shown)


fehguy:
Hmm.  The change was to support `allowEmptyValues` as a query parameter option.  But it's possible that there's a check like this:

```
if(something) {
  // it exists
}
```

Which would fail if the value is zero or false.  Will look into this.

cesine:
@fehguy yeah i found that line, updated the description and pushed the fix. lets see if travis passes

jvivs:
👍 looks good to me.

fehguy:
I'll push out a 2.1.32 release with this.  Swagger-ui is waiting for it as well.

dyang-7:
3.x ?

Issue #943, opened 2017-02-06
Uncaught TypeError: Cannot read property 'prop' of undefined


ankitpopli1891:
Uncaught TypeError: Cannot read property 'prop' of undefined

The error occurs while building the client using `SwaggerClient.buildFromSpec` API.

shockey:
Thanks for the report! Unfortunately, Swagger-JS 2.x is no longer under development. We'd be happy to consider any reports about Swagger-JS 3, though.

Closing as 2.x wontfix.

PR #944, opened 2017-02-06
closes #942, #943

### LINKED ISSUE #942: TypeError: undefined is not an object (evaluating 'window.location.host')
It seems like location is `undefined` [here](https://github.com/swagger-api/swagger-js/blob/2cdccc05c787cf551ffe2ac7f0f5223e26c9598d/lib/types/operation.js#L56)

ankitpopli1891:


webron:
Apologies for the delayed feedback.

Unfortunately, our code base might have deviated from the time this PR was submitted, and we cannot easily merge it (certainly not against master), and so, we're going to close this PR.

If you find the change important and would like to resubmit it, please let us know. Be aware that we are focused on the 3.X version at the moment (now master), and new changes should be applied to it.

We appreciate the time you've taken into submitting this PR. We're now making extra effort to review PRs within a few days at most, to avoid situations like this.

Issue #948, opened 2017-02-18
File formData not working in cURL commands


louisburton:
This commit by @fehguy is handling arrays for type `multi` in collections of form data.
https://github.com/swagger-api/swagger-js/commit/12dc4de84b5b89a7b7e91f7ed5bc37f76154cb9d

I think inadvertently however, this may have regressed the support for when the parameter is of type 'file'. In changing `obj.body.get(parameter.name)` to `obj.body.getAll(parameter.name)`, the return type becomes an array, and the subsequent `paramValue.name` reference is not appropriate for the array (but is fine for the first element of the array).
https://github.com/swagger-api/swagger-js/blame/master/lib/types/operation.js#L1119

As a result, file params don't correctly end up in the cURL command.

shockey:
Thanks for the report! Unfortunately, Swagger-JS 2.x is no longer under development. We'd be happy to consider any reports about Swagger-JS 3, though.

Closing as 2.x wontfix.

Issue #947, opened 2017-02-18
cURL percentage encoding seems inappropriate for json/xml data


louisburton:
Form encoding is expected when the content-type is for form data. I don't think it's correct when the body is not of this type.
I believe @fehguy switched from the unicode escaping to this approach here:
https://github.com/swagger-api/swagger-js/commit/e81b5f97556f2f4b2c173d3331ade7746ace1dd3#diff-e9afd75580721a5b884cc82272c23984L985

Behaviour can be observed using the PetStore sample from swagger-ui (https://github.com/swagger-api/swagger-ui/).
Try creating a user with the following JSON:
```
{
  "id": 123,
  "username": "louis@mytenant",
  "surname": "O'Burton"
}
```

It will generate a cURL command like so:
```
curl -X POST --header 'Content-Type: application/json' --header 'Accept: application/json' -d '{ \ 
   "id": 123, \ 
   "username": "louis%40mytenant", \ 
   "surname": "O%27Burton" \ 
 }' 'http://petstore.swagger.io/v2/user'
```

When issuing this from Linux bash I receive:
`{"code":400,"type":"unknown","message":"bad input"}`

I also believe that the server will unmarshal the JSON with string values that include the literal percentage encodings, which does not seem the desired or expected behaviour to me. I can see that there are tests asserting this however (https://github.com/swagger-api/swagger-js/blob/master/test/request.js#L352), which confuses me.
I can work around it by hacking some post processing into our version of swagger-ui, which is not very desirable. Something like the following, but ideally just for the data part of the curl command:
```
      if (contentType === 'application/json' || contentType === 'application/xml') {
        curlCommand = curlCommand.replace(/%27/g, '\\u0027').replace(/ \\ \n /g, '\n').replace(/%40/g, '@');
      }
```

Apologies if I'm missing something with cURL or missing some prerequisite of using this command.

shockey:
Thanks for the report! Unfortunately, Swagger-JS 2.x is no longer under development. We'd be happy to consider any reports about Swagger-JS 3, though.

Closing as 2.x wontfix.

PR #949, opened 2017-02-20
Fix for #947 & #948 - cURL changes

louisburton:
Fix for #947 & #948 - curl changes - 

- only form encode form content type data (it doesn't make sense to URL encode values in JSON data)
- still ensure apostrophe ' is encoded via unicode if not URL encoding
- remove the new line patterns that prevent cURL commands from executing (for me at least)
- ensure @ is literal outside of file usage (--data-raw usage - 'post body with special chars' use case) - an alternative to this might have been unicode
- fix file submission support

Some test expectations were changed:
- I wouldn't expect the values in my JSON to be received with URL encoding on the server
- I wouldn't expect JS escaping to happen via cURL, this should be the responsibility of the server and the client rendering any feedback.

I have not regenerated the swagger-client.js and swagger-client-min.js as this seemed to introduce more than the desired changes building locally.

I have checked this works for cURL commands with file uploads, and cURL commands with special characters in JSON body with respective server interpretation. Happy to amend this PR - but thought I'd suggest something.

### DIFF
--- a/test/request.js
+++ b/test/request.js
@@ -339,21 +339,28 @@ describe('swagger request functions', function () {
     var petApi = sample.pet;
     var curl = petApi.addPet.asCurl({body: {id:10101}});
 
-    expect(curl).toBe('curl -X POST --header \'Content-Type: application/json\' --header \'Accept: application/json\' -d \'{"id":10101}\' \'http://localhost:8000/v2/api/pet\'');
+    expect(curl).toBe('curl -X POST --header \'Content-Type: application/json\' --header \'Accept: application/json\' --data-raw \'{"id":10101}\' \'http://localhost:8000/v2/api/pet\'');
   });
 
   it('prints a curl post statement from a string', function () {
     var petApi = sample.pet;
     var curl = petApi.addPet.asCurl({body: '{"id":10101}'});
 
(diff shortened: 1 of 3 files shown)


test/request.js
```diff
@@ -339,21 +339,28 @@ describe('swagger request functions', function () {
...
 
   it('prints a curl post statement from a string containing a single quote', function () {
     var petApi = sample.pet;
     var curl = petApi.addPet.asCurl({body: '{"id":"foo\'bar"}'});
 
-    expect(curl).toBe('curl -X POST --header \'Content-Type: application/json\' --header \'Accept: application/json\' -d \'{"id":"foo%27bar"}\' \'http://localhost:8000/v2/api/pet\'');
+    expect(curl).toBe('curl -X POST --header \'Content-Type: application/json\' --header \'Accept: application/json\' --data-raw \'{"id":"foo\\u0027bar"}\' \'http://localhost:8000/v2/api/pet\'');
+  });
+
+  it('prints a curl post statement from an object containing a file', function () {
```
louisburton: This test doesn't actually cover the fixed functionality. It does cover file upload when FormData is not present, but I need a browser or some kind of mock if wanting to check the FormData flow without this.

fehguy:
Hi, thanks for the submission.  I fear that this is trading one problem for another--the `data-raw` may will better handle situations with high-level characters but the marjority of cases will have worse curl examples.  I need to look more but something doesn't look right.  I'll try to give some concrete feedback on the PR shortly.

PR #963, opened 2017-03-16
Binary requests return buffer instead of string 

jdupouy:
Allows to set request content type to `application/pdf` or `application/octet-stream` and receiving a buffer instead of a string when using with node and superagent.

I had to update the superagent dependency since responseType is only supported since version 3.2.1. See https://github.com/visionmedia/superagent/commit/a0e00504906317430a43706d1036ce232c7f127a.


shockey:
Closing, as this PR is against the 2.x codebase.

Thanks, @jdupouy!

PR #966, opened 2017-03-19
Remove access-control-allow-origin from all requests

ponelat:
The header broke the CORS contract, which was evident in the petstore example.
This PR likely breaks the oAuth security flow, which relied on this header. @bodnia, could you let me know how to test the oAuth flow, the reason we added this header in the first place? We may need to find an alternative

### DIFF
--- a/src/execute.js
+++ b/src/execute.js
@@ -54,7 +54,9 @@ export function buildRequest({
   let req = {
     url: baseUrl(spec, scheme),
     headers: {
-      'access-control-allow-origin': '*'
+      // This breaks CORSs... removing this line... probably breaks oAuth. Need to address that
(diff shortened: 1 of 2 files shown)


bodnia:
@ponelat this was made due to breaking of GET requests in `try out`, nothing related to oauth.
Here is the issue https://github.com/SmartBear/swagger-ux/issues/386

ponelat:
@bodnia, I'll give that API a bash and see if we can find a workaround.

bodnia:
@ponelat that was reproducible for other apis too

shockey:
This fixes Try It Out 🎉 

LGTM, will cut a release too.

Issue #965, opened 2017-03-18
Regression: Version 3 does not merge path and operation properties


kevinoid:
In a specification where parameters are defined on both the path and the operation, the parameters from the path appear to be ignored.  For example:

```js
'use strict';

const assert = require('assert');
const http = require('http');
const SwaggerClient = require('swagger-client');

const TEST_HOST = 'localhost';
const TEST_ID = 123;
const TEST_PORT = 8888;

http
  .createServer((req, res) => {
    assert.strictEqual(req.url, `/cars/${TEST_ID}`);
    res.writeHead(200, {'Content-Type': 'application/json'});
    res.end('{}');
  })
  .listen(TEST_PORT, TEST_HOST, () => {
    new SwaggerClient({
      usePromise: true,
      spec: {
        swagger: '2.0',
        info: {
          title: 'Path Parameter API',
          version: '1.0.0'
        },
        host: `${TEST_HOST}:${TEST_PORT}`,
        schemes: [
          'http'
        ],
        paths: {
          '/cars/{carId}': {
            parameters: [
              {
                name: 'carId',
                in: 'path',
                type: 'number',
                required: true
              }
            ],
            get: {
              operationId: 'getCar',
              parameters: [
                {
                  name: 'clean',
                  in: 'query',
                  type: 'boolean'
                }
              ],
              responses: {
                200: {
                  description: 'A car',
                  schema: {
                    type: 'object'
                  }
                }
              }
            }
          }
        }
      }
    })
      .then(client => client.apis.default.getCar({carId: TEST_ID}))
      .then(
        () => {
          console.log('It works.');
          process.exit();
        },
        err => {
          console.error(err);
          process.exit(1);
        }
      );
  });
```

With `swagger-js` version 3.0.1 this fails with `AssertionError: '/cars/%7BcarId%7D' === '/cars/123'` while `swagger-js` version 2.1.32 prints `It works.`.  I bisected the regression to 7678d40, but that's obviously not very helpful since almost everything was replaced in that commit.

Thanks,
Kevin

webron:
@kevinoid Thanks for the report. This is a complete rewrite so previous code may not be relevant. I was sure we had a test for this case, but we'll definitely look into it.

kevinoid:
Great, thanks @webron!

webron:
[...]

PR #972, opened 2017-03-22
Merge Path and Operation parameters

saharj:
Fixes https://github.com/swagger-api/swagger-js/issues/965

### DIFF
--- a/src/execute.js
+++ b/src/execute.js
@@ -89,29 +89,29 @@ export function buildRequest({
 
   // Add values to request
   arrayOrEmpty(operation.parameters) // operation parameters
-  .concat(arrayOrEmpty(spec.paths[pathName].parameters)) // path parameters
-  .forEach((parameter) => {
-    const builder = parameterBuilders[parameter.in]
-    let value
+    .concat(arrayOrEmpty(spec.paths[pathName].parameters)) // path parameters
+    .forEach((parameter) => {
+      const builder = parameterBuilders[parameter.in]
+      let value
 
-    if(parameter.in === 'body' && parameter.schema && parameter.schema.properties) {
-      value = parameters
-    }
+      if(parameter.in === 'body' && parameter.schema && parameter.schema.properties) {
+        value = parameters
+      }
 
-    value = parameter && parameter.name && parameters[parameter.name]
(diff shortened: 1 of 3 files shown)


saharj:
@ponelat

src/execute.js
```diff
@@ -86,8 +86,11 @@ export function buildRequest({
   if (requestContentType) {
     req.headers['content-type'] = requestContentType
   }
+
   // Add values to request
-  arrayOrEmpty(operation.parameters).forEach((parameter) => {
+  arrayOrEmpty(operation.parameters) // operation parameters
+  .concat(arrayOrEmpty(spec.paths[pathName].parameters)) // path parameters
```
buunguyen: Please add indent to this and the next line

test/execute.js
```diff
@@ -1196,6 +1196,80 @@ describe('execute', () => {
...
+                  in: 'path',
+                  type: 'number',
+                  required: true
+                }
+              ],
+            }
+          }
+        }
+
+        const req = buildRequest({ spec, operationId: 'getPetsById', parameters: { id: 123, test: 567 } })
```
buunguyen: Please remove empty space inside `{}`.

src/helpers.js
```diff
@@ -60,6 +60,9 @@ export function eachOperation(spec, cb, find) {
   // Iterate over the spec, collecting operations
   for (const pathName in paths) {
     for (const method in paths[pathName]) {
+      if (method.toUpperCase() === 'PARAMETERS') {
```
buunguyen: What is this check for? Would be great to add a comment clarifying the intention.

saharj: Oddly, when path-parameters come before the operation like below, the method becomes `PARAMETERS` (here it should be `GET`), and that causes problems.
```
paths: {
  '/pet/{id}': {
    parameters: [
      {...}
    ],
    get: {
      operationId: 'getPetsById',
      parameters: [
        {...}
      ],
    }
  }
}
```

buunguyen: I see. The loop iterates the keys of each path, so "parameters" will appear. Doesn't matter if it appears before or after the "get".

PR #977, opened 2017-03-25
Fixing bug swagger-ui#2785: Encoding only characters with a reserved purpose in path segments

carlos-azb:
AFAIK, the only characters with a reserved purpose in path segments are the percent sign (used to encode the other reserved characters), the slash (used to separate path segments), the question mark (used to separate the path from the query) and the hashtag (used to separate the path or the query from the fragment). Please see https://tools.ietf.org/html/rfc3986#section-3

### DIFF
--- a/src/execute.js
+++ b/src/execute.js
@@ -153,7 +153,9 @@ export function headerBuilder({req, parameter, value}) {
 
 // Replace path paramters, with values ( ie: the URL )
 export function pathBuilder({req, value, parameter}) {
-  req.url = req.url.replace(`{${parameter.name}}`, encodeURIComponent(value))
+  const encodedValue = value.toString().replace('%', '%25').replace('/', '%2F').replace('?', '%3F')
+    .replace('#', '%23')
+  req.url = req.url.replace(`{${parameter.name}}`, encodedValue)
 }
 
 // Add a query to the `query` object, which will later be stringified into the URL's search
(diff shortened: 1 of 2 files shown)


buunguyen:
I traced the grammar in that spec and encodeURIComponent is indeed too aggressive. However, I couldn't tell for sure if only `%#?/` should be escaped. For example, it's not clear to me whether `^`, `<`, `>` are allowed or not in the path segment. Unless we can come up with an unambiguous blacklist, I would think the aggressive approach is better, especially when it relies on the built-in implementation of the browsers.

webron:
Okay, so the the characters that need to be escaped in *path segments* are`%?/#[]` and white spaces. Technically, any non-printable ASCII character also needs to be escaped, but would recommend sticking with the basics for now, and should the time come, use a library for URI encoding. 

@carlos-azb Please update the PR to reflect those changes.

buunguyen:
I've just discussed with @webron, the spec is not exhaustive in the list of chars that must be escaped in path segments. It's probably better to escape based on a whitelist, which can be concluded from the spec, instead of a blacklist. Alternatively, we might consider encode the fully assembled URL, if there's a method or lib to do that efficiently.

@carlos-azb any thought?  Would you be willing to go ahead making the change based on these ideas?

webron:
@carlos-azb any feedback?

webron:
@carlos-azb pinging you again

shockey:
Closing this; our encoding implementation has changed since this PR was made.

Thanks for your effort here, @carlos-azb!

PR #984, opened 2017-03-29
Fix problem cannot resolve relative references

buunguyen:
See https://github.com/swagger-api/swagger-ui/issues/2767#issuecomment-289913619

There are 2 problems:

* SwaggerUI passes in the wrong parameter name. To avoid having to change SwaggerUI, I change SwaggerJS to allow both names (because the correct name is used both internally and possibly other clients). We should clean up interface anyway in the next refactor.

* There's a certain behavior in the context-tree component that causes the baseDoc to be clear for all but the first reference. I talked to Josh and it was confirmed as a bug. Already add failed test and fix it.

I've tested the latest code with UI & the reported spec, it works without problem.  (Also, unrelated, but I also fixed some failed tests after merging from upstream master, we should enforce code change better and avoid code not passing tests to land into master.)

cc @webron @shockey @ponelat

### DIFF
--- a/src/resolver.js
+++ b/src/resolver.js
@@ -20,7 +20,11 @@ export function clearCache() {
   plugins.refs.clearCache()
 }
 
-export default function resolve({http, fetch, spec, baseDoc, mode, allowMetaPatches = true}) {
+export default function resolve({http, fetch, spec, url, baseDoc, mode, allowMetaPatches = true}) {
+  // @TODO Swagger-UI uses baseDoc instead of url, this is to allow both
+  // need to fix and pick one.
+  baseDoc = baseDoc || url
+
   // Provide a default fetch implementation
   // TODO fetch should be removed, and http used instead
   http = fetch || http || Http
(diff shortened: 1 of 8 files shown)


src/resolver.js
```diff
@@ -20,19 +20,23 @@ export function clearCache() {
   plugins.refs.clearCache()
 }
 
-export default function resolve({http, fetch, spec, url, mode, allowMetaPatches = true}) {
+export default function resolve({http, fetch, spec, url, baseDoc, mode, allowMetaPatches = true}) {
+  // @TODO Swagger-UI uses baseDoc instead of url, this is to allow both
+  // need to fix and pick one.
+  baseDoc = baseDoc || url
```
ponelat: I would do the reverse.
`url = url || baseDoc` with a note saying `baseDoc` is deprecated.

ponelat:
lgtm. 
@buunguyen do you have an issue to update swagger-ui, removing `baseDoc`?

buunguyen:
I added `baseDoc` so that ui doesn't have to change. The TODO will be dealt with when I have a chance to review everything.

Issue #979, opened 2017-03-25
Internal error for non-text responses


kevinoid:
Responses with a non-text `Content-Type` cause `swagger-client` to fail with an uninformative error.  For example:

```js
'use strict';

const assert = require('assert');
const http = require('http');
const SwaggerClient = require('swagger-client');

const TEST_HOST = 'localhost';
const TEST_ID = 123;
const TEST_PORT = 8888;

http
  .createServer((req, res) => {
    assert.strictEqual(req.url, '/binary-data');
    res.writeHead(200, {'Content-Type': 'application/octet-stream'});
    res.end(new Buffer([1, 2, 3, 4, 5]));
  })
  .listen(TEST_PORT, TEST_HOST, () => {
    new SwaggerClient({
      usePromise: true,
      spec: {
        swagger: '2.0',
        info: {
          title: 'Path Parameter API',
          version: '1.0.0'
        },
        host: `${TEST_HOST}:${TEST_PORT}`,
        schemes: [
          'http'
        ],
        paths: {
          '/binary-data': {
            get: {
              operationId: 'getBinaryData',
              responses: {
                200: {
                  description: 'Some binary data',
                  schema: {
                    type: 'file'
                  }
                }
              }
            }
          }
        }
      }
    })
      .then(client => client.apis.default.getBinaryData())
      .then(
        () => {
          console.log('It works.');
          process.exit();
        },
        err => {
          console.error(err);
          process.exit(1);
        }
      );
  });
```
Produces the following error:

    TypeError: oriRes[(intermediate value)(intermediate value)(intermediate value)] is not a function
        at Object.serializeRes (dist/index.js:443:43)
        at dist/index.js:414:17
        at process._tickCallback (internal/process/next_tick.js:103:7)

The cause appears to be that [`node-fetch`](https://github.com/bitinn/node-fetch) (used by [`isomorphic-fetch`](https://github.com/matthew-andrews/isomorphic-fetch)) only supports the `.blob` response property in version 2 (which is currently in alpha) causing the failure [when `useText` is `false` and `.blob` is called in serializeRes](https://github.com/swagger-api/swagger-js/blob/v3.0.3/src/http.js#L68).

[...]

PR #985, opened 2017-03-30
Use .buffer() for binary responses when .blob() is not available

kevinoid:
This PR implements the fix requested in https://github.com/swagger-api/swagger-js/issues/979#issuecomment-290249977 by using `.buffer()` to get the binary response body when `.blob()` is not available, which occurs with `node-fetch` prior to version 2 (currently in alpha).

It also fixes the existing test for `serializeRes` which would always pass due to not returning the Promise value it created.  If you would prefer this to be split into a separate PR, let me know.

Thanks,
Kevin

### DIFF
--- a/src/http.js
+++ b/src/http.js
@@ -54,7 +54,7 @@ function shouldDownloadAsText(contentType) {
 }
 
 // Serialize the response, returns a promise with headers and the body part of the hash
-export function serializeRes(oriRes, url, {loadSpec = false}) {
+export function serializeRes(oriRes, url, {loadSpec = false} = {}) {
   const res = {
     ok: oriRes.ok,
     url: oriRes.url || url,
--- a/src/http.js
+++ b/src/http.js
@@ -65,7 +65,10 @@ export function serializeRes(oriRes, url, {loadSpec = false}) {
 
   const useText = loadSpec || shouldDownloadAsText(res.headers['content-type'])
 
-  return oriRes[useText ? 'text' : 'blob']().then((body) => {
+  // Note: Response.blob not implemented in node-fetch 1.  Use buffer instead.
+  const getBody = useText ? oriRes.text : (oriRes.blob || oriRes.buffer)
+
+  return getBody.call(oriRes).then((body) => {
     res.text = body
     res.data = body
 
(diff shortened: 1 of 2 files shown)


buunguyen:
Hi, the new tests are failing. Can you check?

BTW, some old tests fail, not because your code, I've merged the fixes to master already. Be sure to pull from upstream and run test locally.

Thanks!

kevinoid:
The new `serializeRes` tests passed in https://travis-ci.org/swagger-api/swagger-js/jobs/216616487  I rebased against the master branch fixes and the tests are passing on my machine.  Should pass after the next Travis run.

kevinoid:
https://travis-ci.org/swagger-api/swagger-js/builds/216618260 is complete.  The failure on Node 4.7 is due to use of `const` in strict mode in `src/server.js` which is not modified by this PR.

buunguyen:
Interesting, the error is in a lib, fetch-mock...  Wondering why. I'll check locally.

```js
/home/travis/build/swagger-api/swagger-js/node_modules/fetch-mock/src/server.js:2
const fetch = require('node-fetch');
```

kevinoid:
Ah, right you are.  I thought `src/server.js` was in this package.  Good catch.

buunguyen:
Ha! Local works. Turn out Travis use 0.10 even though the env is set to 4.7. Probably a Travis bug. But your code is good. If you don't mind, please make this minor change:

Replace
```js
export function serializeRes(oriRes, url, {loadSpec = false}) {
```

With
```js
export function serializeRes(oriRes, url, {loadSpec = false} = {}) {
```

And then modify the tests so that it doesn't supply the empty object.

I'll deal with Travis and merge this PR later. Thanks!

kevinoid:
Sure thing.  Done.  Let me know if you see anything else.  Thanks!

buunguyen:
Thanks! I've fixed Travis, so the mystic error with fetch-mock has gone away. But your tests are failing in Node 4 (works in Node 6). Can you check?

kevinoid:
That took some digging, but I think I've fixed it.  Waiting on the test results to confirm.

buunguyen:
Good stuff, thank you!

kevinoid:
My pleasure.  Thank you for reviewing and merging it!

PR #987, opened 2017-03-30
Allow mocking http function

ponelat:
xmock doesn't support `fetch` that makes it difficult to do end to end http tests.
In the meantime we can mock out the `http` function. It was possible to do this in the static functions... this PR adds it to the resolve _method_

NOTE: It also skips the network tests.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -62,6 +62,7 @@ Swagger.prototype = {
 
   resolve() {
     return Swagger.resolve({
+      http: this.http,
       spec: this.spec,
       url: this.url,
       allowMetaPatches: this.allowMetaPatches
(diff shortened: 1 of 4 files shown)


buunguyen:
@ponelat @fehguy the skipped tests (interceptors) don't use network, they use xmock.

ponelat:
@buunguyen you're right. My apologies

webron:
@ponelat can you revert the skipped tests then?

ponelat:
@webron they are in another PR...https://github.com/swagger-api/swagger-js/pull/988 ( along with the real reason they failed... globals! )

I can make it a minimal change if required. The above PR is a general solution. But it would be tomorrow morning. Or we can revert the PR in master.... since I don't think Tony needs it this very instant.

buunguyen:
Can I revert this then? I want to merge a user contributed PR that seems to be conflict with this change. Then @ponelat can issue a new PR that combines this and 988, we can discuss how to make the tests run before merging that new PR.

ponelat:
@buunguyen go ahead and revert

PR #988, opened 2017-03-30
Remove globals from plugins.refs, to avoid test hassles

ponelat:
Does two things...
One removes the need to override globals on plugins.refs. Which caused a bit of mischief in tests.
This is reasonably large change... and is untested in swagger-ui. No public APIs were changed ( only specmap).

ping @buunguyen @shockey @fehguy 
Not sure when I get a chance to verify this works well in swagger-ui. 

NOTE: skipped one test, no idea how it worked before... probably relied on the global modification, but not sure...

### DIFF
--- a/test/specmap/refs.js
+++ b/test/specmap/refs.js
@@ -101,7 +101,9 @@ describe('refs', function () {
       })
     })
 
-    it('should parse YAML docs into JSON', function () {
+    // How is this supposed to work within the confines of plugins.refs?
+    // It makes sense if we override... but not sure this worked before
+    it.skip('should parse YAML docs into JSON', function () {
       const url = 'http://example.com/common.yaml'
 
       xapp.get(url, (req, res, next) => {
@@ -154,7 +156,7 @@ describe('refs', function () {
       refs.docCache['some-path'] = {
         one: '1'
       }
-      return refs.extractFromDoc('some-path', '/one')
+      return refs.extractFromDoc({url: 'some-path', pointer: '/one'})
         .then((val) => {
           expect(val).toEqual('1')
         })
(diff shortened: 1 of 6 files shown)


ponelat:
@fehguy your swagger: 3.0 issue was a cache on the `http://petstore.swagger.io/v2/swagger.json` url. You can just clear the cache to run your test. But this is a general solution.

ponelat:
Also removed fetch-mock...  we're only testing in nodejs ( which doesn't have fetch ).

test/specmap/refs.js
```diff
@@ -101,7 +101,9 @@ describe('refs', function () {
       })
     })
 
-    it('should parse YAML docs into JSON', function () {
+    // How is this supposed to work within the confines of plugins.refs?
+    // It makes sense if we override... but not sure this worked before
```
buunguyen: It worked before.

ponelat: @buunguyen ok.. could you point to the place where the code lives, perhaps?

buunguyen:
Close as discussed, this will be combined with other changes in a new PR. BTW, we need fetch-mock, which is used in https://github.com/swagger-api/swagger-js/pull/985.

test/specmap/refs.js
buunguyen: @ponelat which code, Josh?

ponelat:
Would you like this PR remade, with the changes that were reverted in master?
As to fetch-mock... I found it wasn't _needed_ in the node tests... since `xmock` mocks out `http`, which is what was used in nodejs. To make the tests work in a browser requires much broader changes ( where `fetch` will be used natively ).

buunguyen:
Yes please.

node-fetch is being used by some tests. So if we move to something else, we need to fix those tests. Do you have any proposal for a replacement? IMO, it's not a high priority though. Currently we only run tests on Travis in Node env not browser env.

PR #990, opened 2017-03-31
Add path parameters to operation parameters

saharj:
Currently path params are not being added to operation params, therefore path params are not shown in the `Try It Out` section.

This PR is fixing it.

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -144,7 +144,6 @@ export function normalizeSwagger(parsedSpec) {
           toBeInherit.parameters = pathParameters
           inheritsList.push(toBeInherit)
         }
-        
 
         if (inheritsList.length !== 0) {
           for (const inherits of inheritsList) {
@@ -153,7 +152,8 @@ export function normalizeSwagger(parsedSpec) {
             for (const inhName in inherits) {
               if (!operation[inhName]) {
                 operation[inhName] = inherits[inhName]
-              } else if(inhName === "parameters") {
+              }
+              else if(inhName === "parameters") {
                 var exists = false
                 for(const item in inherits[inhName]) {
 
@@ -169,9 +169,9 @@ export function normalizeSwagger(parsedSpec) {
                 }
               }
             }
-
           }
         }
+
       }
     }
   }
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -144,7 +144,7 @@ export function normalizeSwagger(parsedSpec) {
           toBeInherit.parameters = pathParameters
           inheritsList.push(toBeInherit)
         }
-
+        
 
         if (inheritsList.length !== 0) {
           for (const inherits of inheritsList) {
@@ -153,6 +153,10 @@ export function normalizeSwagger(parsedSpec) {
             for (const inhName in inherits) {
               if (!operation[inhName]) {
                 operation[inhName] = inherits[inhName]
+              } else if(inhName === "parameters") {
+                for(const item in inherits[inhName]) {
+                  operation[inhName].push(inherits[inhName][item])
+                }
               }
             }
           }
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -154,11 +154,22 @@ export function normalizeSwagger(parsedSpec) {
               if (!operation[inhName]) {
                 operation[inhName] = inherits[inhName]
               } else if(inhName === "parameters") {
+                var exists = false
                 for(const item in inherits[inhName]) {
-                  operation[inhName].push(inherits[inhName][item])
+
+                  for(const param in operation[inhName]) {
+                    if(operation[inhName][param].name === inherits[inhName][item].name) {
+                      exists = true
+                    }
+                  }
+
+                  if(!exists) {
+                    operation[inhName].push(inherits[inhName][item])
+                  }
                 }
               }
             }
+
           }
         }
       }
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -153,17 +153,17 @@ export function normalizeSwagger(parsedSpec) {
               if (!operation[inhName]) {
                 operation[inhName] = inherits[inhName]
               }
-              else if(inhName === "parameters") {
+              else if (inhName === "parameters") {
                 var exists = false
-                for(const item in inherits[inhName]) {
+                for (const item in inherits[inhName]) {
 
-                  for(const param in operation[inhName]) {
-                    if(operation[inhName][param].name === inherits[inhName][item].name) {
+                  for (const param in operation[inhName]) {
+                    if (operation[inhName][param].name === inherits[inhName][item].name) {
                       exists = true
                     }
                   }
 
-                  if(!exists) {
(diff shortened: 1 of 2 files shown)


src/helpers.js
```diff
@@ -144,7 +144,7 @@ export function normalizeSwagger(parsedSpec) {
           toBeInherit.parameters = pathParameters
           inheritsList.push(toBeInherit)
         }
-
+
```
ponelat: @saharj get `editorconfig` plugin for you IDE... it'll help remove these little line things.

test/helpers.js
```diff
@@ -254,14 +254,18 @@ describe('helpers', function () {
     })
 
     describe('parameters', function () {
-      it('should not overwrite parameters from path when exists in operation', function () {
+      it('should add parameters from path to operation without overwriting', function () {
```
ponelat: @webron is this the desired behaviour?
Should global `parameters` get merged with operation-level `parameters`? 
Just want to clarify this point... since we've done both in the past.

webron: Yup. They are merged. The only thing is if they exist in both (based on both the `name` and the `in` as the unique identifier), then the operation level parameter will override the path level one.

ponelat: @saharj I don't think that second part is covered in this PR
so...
```yaml
parameters:
  - name: josh
    in: query
    type: string
paths:
  /one:
    parameters:
    - name: sahar
      in: query
      type: string
    - name: josh
      in: query
      type: number
```

`/one.parameters` should have two parameters...
```yaml
- name: sahar
  in: query
  type: string
- name: josh
  in: query
  type: number
```

ponelat: ie: what @webron said

src/helpers.js
```diff
@@ -153,6 +153,10 @@ export function normalizeSwagger(parsedSpec) {
             for (const inhName in inherits) {
               if (!operation[inhName]) {
                 operation[inhName] = inherits[inhName]
+              } else if(inhName === "parameters") {
```
buunguyen: Please put `else` in a new line, to be consistent with current code.

src/helpers.js
```diff
@@ -153,6 +153,10 @@ export function normalizeSwagger(parsedSpec) {
...
+                for(const item in inherits[inhName]) {
```
buunguyen: We need a whitespace space after `for`.

src/helpers.js
```diff
@@ -153,7 +152,8 @@ export function normalizeSwagger(parsedSpec) {
             for (const inhName in inherits) {
               if (!operation[inhName]) {
                 operation[inhName] = inherits[inhName]
-              } else if(inhName === "parameters") {
+              }
+              else if(inhName === "parameters") {
```
buunguyen: Sorry I missed this. But we do need space after `for`, `if` etc.

src/helpers.js
```diff
@@ -154,9 +153,25 @@ export function normalizeSwagger(parsedSpec) {
...
+
+                  if(!exists) {
+                    operation[inhName].push(inherits[inhName][item])
+                  }
+                }
+              }
             }
           }
         }
+
```
saharj: This line is added intentionally to make the code easier to follow.

saharj:
@buunguyen Can you please review the new changes?

src/helpers.js
```diff
@@ -154,9 +153,25 @@ export function normalizeSwagger(parsedSpec) {
               if (!operation[inhName]) {
                 operation[inhName] = inherits[inhName]
               }
+              else if (inhName === "parameters") {
+                var exists = false
```
buunguyen: `var` -> `let`

buunguyen: @ponelat it's already addressed here https://github.com/swagger-api/swagger-js/pull/990/files#diff-42a353e98c906af524c7d3d48feaf72aR359

src/helpers.js
```diff
@@ -154,9 +153,25 @@ export function normalizeSwagger(parsedSpec) {
               if (!operation[inhName]) {
                 operation[inhName] = inherits[inhName]
               }
+              else if (inhName === "parameters") {
```
buunguyen: Please rename to `inheritName`.
Also, we use single quotes.

src/helpers.js
```diff
@@ -154,9 +153,25 @@ export function normalizeSwagger(parsedSpec) {
...
+                var exists = false
+                for (const item in inherits[inhName]) {
```
buunguyen: Please change to a more descriptive name than `item`.

src/helpers.js
```diff
@@ -154,9 +153,25 @@ export function normalizeSwagger(parsedSpec) {
...
+
+                  for (const param in operation[inhName]) {
+                    if (operation[inhName][param].name === inherits[inhName][item].name) {
+                      exists = true
+                    }
+                  }
+
+                  if (!exists) {
```
buunguyen: This looks wrong because `exists` is never reset. So after the first `exists = true`, all subsequent params will be be pushed to operation.

src/helpers.js
```diff
@@ -154,9 +153,25 @@ export function normalizeSwagger(parsedSpec) {
               if (!operation[inhName]) {
                 operation[inhName] = inherits[inhName]
               }
+              else if (inhName === "parameters") {
+                var exists = false
+                for (const item in inherits[inhName]) {
+
+                  for (const param in operation[inhName]) {
```
buunguyen: Too verbose.
```
exists = operation[inheritName].some((param) => param.name === inherit[inheritName][item].name)
```

src/helpers.js
buunguyen: `inherits[inhName]` is an array. Should use `for... of`. Following code would also be shorter as no need to deref into array index.

buunguyen:
@saharj please also resolve merge conflicts.

src/helpers.js
buunguyen: Confirmed: above code logic is incorrect. This test will fail:

```
        const spec = {spec: {
          paths: {
            '/two': {
              parameters: [
                {name: 'a', in: 'path'},
                {name: 'b', in: 'path'}
              ],
              get: {
                parameters: [
                  {name: 'a', in: 'query'},
                  {name: 'c', in: 'query'}
                ]
              }
            }
          }
        }}

        const resultSpec = normalizeSwagger(spec)

        expect(resultSpec).toEqual({spec: {
          paths: {
            '/two': {
              parameters: [
                {name: 'a', in: 'path'},
                {name: 'b', in: 'path'}
              ],
              get: {
                parameters: [
                  {name: 'a', in: 'query'},
                  {name: 'b', in: 'path'}, // <---  'b' isn't pushed because `exists` isn't reset
                  {name: 'c', in: 'query'}
                ]
              }
            }
          }
        }})
```

buunguyen:
Close in favor of https://github.com/swagger-api/swagger-js/pull/999, which fixes all the coding issues & bugs.

PR #992, opened 2017-04-04
Remove globals from plugins.refs, allowing mocking this.http and skip specmap-yaml-parsing

ponelat:
Adds https://github.com/swagger-api/swagger-js/pull/988

@buunguyen regarding https://github.com/swagger-api/swagger-js/blob/master/test/specmap/refs.js#L104 there is no yaml-parser in `./specmap/**`. The only way I can imagine this test passing, is if the `plugins.refs.fetchJSON` was added, in some other test. That used `./src/http.js` ( which does have yaml parsing ). 

I've skipped it again, since that is a flaw for me. We should be testing that in another way, on swagger-js level.

PS: I've removed fetch-mock and the tests that used it. They now use `xmock`

### DIFF
--- a/src/http.js
+++ b/src/http.js
@@ -54,7 +54,7 @@ function shouldDownloadAsText(contentType) {
 }
 
 // Serialize the response, returns a promise with headers and the body part of the hash
-export function serializeRes(oriRes, url, {loadSpec = false} = {}) {
+export function serializeRes(oriRes, url, {loadSpec = false}) {
   const res = {
     ok: oriRes.ok,
     url: oriRes.url || url,
--- a/src/resolver.js
+++ b/src/resolver.js
@@ -4,14 +4,17 @@ import {normalizeSwagger} from './helpers'
 
 export function makeFetchJSON(http) {
   return (docPath) => {
-    return http({
+    return Promise.resolve(http({
       url: docPath,
       loadSpec: true,
       headers: {
         Accept: 'application/json'
       }
+    }))
+    .then((res) => {
+      // To allow overriding with spies
+      return res.body || res
     })
-    .then(res => res.body)
   }
 }
 
--- a/src/resolver.js
+++ b/src/resolver.js
@@ -31,6 +31,7 @@ export default function resolve({http, fetch, spec, url, baseDoc, mode, allowMet
   // Provide a default fetch implementation
   // TODO fetch should be removed, and http used instead
   http = fetch || http || Http
+  const docCache = {}
 
   if (!spec) {
     // We create a spec, that has a single $ref to the url
@@ -39,11 +40,11 @@ export default function resolve({http, fetch, spec, url, baseDoc, mode, allowMet
   }
   else {
     // Store the spec into the url provided, to cache it
-    plugins.refs.docCache[baseDoc] = spec
+    docCache[baseDoc] = spec
   }
 
   // Build a json-fetcher ( ie: give it a URL and get json out )
-  plugins.refs.fetchJSON = makeFetchJSON(http)
+  const fetchJSON = makeFetchJSON(http)
 
   const plugs = [plugins.refs]
 
@@ -53,6 +54,8 @@ export default function resolve({http, fetch, spec, url, baseDoc, mode, allowMet
 
   // mapSpec is where the hard work happens, see https://github.com/swagger-api/specmap for more details
   return mapSpec({
+    fetchJSON,
+    docCache,
     spec,
     context: {baseDoc},
     plugins: plugs,
--- a/test/http.js
+++ b/test/http.js
@@ -1,6 +1,7 @@
+/* global Headers */
 import expect from 'expect'
 import xmock from 'xmock'
-import fetchMock from 'fetch-mock'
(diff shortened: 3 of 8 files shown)


test/http.js
```diff
@@ -159,64 +162,26 @@ describe('http', () => {
...
 
-      return fetch('http://swagger.io').then((_res) => {
+      fetch('http://swagger.io').then((_res) => {
         return serializeRes(_res, 'https://swagger.io')
       }).then((resSerialize) => {
         expect(resSerialize.headers).toEqual({authorization: ['Basic hoop-la', 'Advanced hoop-la']})
-      }).then(fetchMock.restore)
-    })
-
-    it('should set .text and .data to body Blob or Buffer for binary response', function () {
```
buunguyen: @ponelat why do you remove this and the following test?

test/specmap/refs.js
```diff
@@ -101,7 +101,8 @@ describe('refs', function () {
       })
     })
 
-    it('should parse YAML docs into JSON', function () {
+    // There is no yaml parser in specmap... only in swagger-js.
+    it.skip('should parse YAML docs into JSON', function () {
```
buunguyen: @ponelat That's okay. I can move this to http.

src/http.js
```diff
@@ -54,7 +54,7 @@ function shouldDownloadAsText(contentType) {
 }
 
 // Serialize the response, returns a promise with headers and the body part of the hash
-export function serializeRes(oriRes, url, {loadSpec = false} = {}) {
+export function serializeRes(oriRes, url, {loadSpec = false}) {
```
buunguyen: We need the default val in case user doesn't pass in an object for the 3rd arg, which would cause an destructuring error.

src/resolver.js
```diff
@@ -4,14 +4,17 @@ import {normalizeSwagger} from './helpers'
 
 export function makeFetchJSON(http) {
   return (docPath) => {
-    return http({
+    return Promise.resolve(http({
```
buunguyen: Why `Promise.resolve()` is necessary here? Doesn't `http()` always return a promise or is there any case it returns value directly?

src/resolver.js
```diff
@@ -4,14 +4,17 @@ import {normalizeSwagger} from './helpers'
...
       url: docPath,
       loadSpec: true,
       headers: {
         Accept: 'application/json'
       }
+    }))
+    .then((res) => {
+      // To allow overriding with spies
```
buunguyen: Would be great if this is clearer, overriding what with spies? What spies? Why returning `res.body || res` instead of `res.body` would achieve that?

src/resolver.js
```diff
@@ -36,11 +40,11 @@ export default function resolve({http, fetch, spec, url, baseDoc, mode, allowMet
   }
   else {
     // Store the spec into the url provided, to cache it
-    plugins.refs.docCache[baseDoc] = spec
+    docCache[baseDoc] = spec
```
buunguyen: It seems like something like this could be done inside mapSpec. That ways we don't have to leak docCache to JS.

test/http.js
ponelat: sorry, that's an artifact of the way I tried to recover the files. I'll fix that in the morning.

src/http.js
ponelat: yeah... also an artifact. Apologies.

src/resolver.js
ponelat: Such that we can mock, without having to provide `.json`
eg
```js
http({ one: 1})
// vs
http({ json() { return Promise.resolve({ one: 1 }) })
```
Given that we're using `xmock` throughout... we don't strictly need mocking the `http` function anymore. I'll revise this...

src/resolver.js
ponelat: The spy in this case is the `createSpy()` exposed in `expect`
Its to allow putting in a mock'd version of `http` ( as referred to in previous comments ).

This is here, so that we don't need to provide a `.body` property on what the mock'd  `http` method returns. Given that this will be rarely used. I don't mind removing it. It was a workaround to start with... now that I know we only test in nodejs... we can continue to use `xmock`

src/resolver.js
ponelat: cacheing is hard.
I don't mind putting it inside specmap... as to the cache. I would prefer we avoided a global. At the same time, I don't know of the performance of recreating the cache each time.
Your call on that.

buunguyen: Yes, we should avoid global. The `docCache` here is created inside `resolve()` method, I suggest we move that inside `mapSpec()`. Basically just encapsulate the cache creation inside SpecMap, `resolve()` doesn't have to know about the cache at all. Therefore, there's no performance difference whatsoever.

ponelat:
Will try to get to this over the weekend

ponelat:
Will try to get this revised over the weekend!

ponelat:
I clearly lied. Will try to wrap this up soon as I can.

buunguyen:
Pretty outdated. Close for now, please reissue a new PR.

ponelat:
Sorry for not getting this out there...

PR #994, opened 2017-04-06
Fixed sending header parameters when value is not set.

maveron58:
When user API specification contains custom header input parameter which is optional (required field equals false) and user doesn't provide value for it then client sends header with value of undefined. Example:
```
POST http://localhost:8080/example/url HTTP/1.1
Content-Length: 2
Content-Type: application/json
X-Customer-Header: undefined

{}
```

When header is optional and allowEmptyValue is set to false then header should not be sent. Example:
```
POST http://localhost:8080/example/url HTTP/1.1
Content-Length: 2
Content-Type: application/json

{}
```

When header is optional and allowEmptyValue is set to true then header should be sent with empty value. Example:
```
POST http://localhost:8080/example/url HTTP/1.1
Content-Length: 2
Content-Type: application/json

{}
```

According to https://tools.ietf.org/html/rfc7230 empty header value is valid.

### DIFF
--- a/src/execute.js
+++ b/src/execute.js
@@ -150,7 +150,9 @@ export function formDataBuilder({req, value, parameter}) {
 // Add a header to the request
 export function headerBuilder({req, parameter, value}) {
   req.headers = req.headers || {}
-  req.headers[parameter.name] = value
+  if (typeof value !== 'undefined') {
+    req.headers[parameter.name] = value
+  }
 }
 
 // Replace path paramters, with values ( ie: the URL )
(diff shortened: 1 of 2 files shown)


buunguyen:
@maveron58 can you add tests that would fail without this fix?

buunguyen:
Agreed. I'm not sure if it's changed in the spec somewhere but the last time I looked at it, `allowEmptyValue` doesn't say anything about a header must be set if the value doesn't not present. So I want to know if there's any valid test case that would fail without this fix.

maveron58:
I've added two tests. Without this fix both test cases are failing (generated request has api_key parameter set to 'udefined').
I 've also changed headerBuilder to yours request (removed empty string sending).

PR #997, opened 2017-04-06
Catch baseUrls that have trailing slashes in them

shockey:
Fixes https://github.com/swagger-api/swagger-ui/issues/2815

### DIFF
--- a/src/execute.js
+++ b/src/execute.js
@@ -184,7 +184,10 @@ export function baseUrl({spec, scheme, contextUrl = ''}) {
   applyScheme = applyScheme ? `${applyScheme}:` : ''
 
   if (host || basePath || contextUrl) {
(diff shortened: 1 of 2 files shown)


src/execute.js
```diff
@@ -184,7 +184,10 @@ export function baseUrl({spec, scheme, contextUrl = ''}) {
   applyScheme = applyScheme ? `${applyScheme}:` : ''
 
   if (host || basePath || contextUrl) {
-    return `${applyScheme}//${host || parsedUrl.host || ''}${basePath || ''}`
+    let res = `${applyScheme}//${host || parsedUrl.host || ''}${basePath || ''}`
+
+    // If last character is "/", trim it off
+    return res[res.length - 1] === "/" ? res.slice(0, -1) : res
```
buunguyen: Should use single quote for consistency.

PR #999, opened 2017-04-07
Fix bug param merging and add better tests

buunguyen:


buunguyen:
@ponelat @webron fyi.

PR #1011, opened 2017-04-10
make node-inspector a dev dependency

fgnass:
This PR removes the `node-inspector` optional dependency and converts it into a dev-dependency so that it no longer gets installed when _swagger-js_ is added to a project (npm installs optional deps by default unless --no-optional is used).


fgnass:
Since the installation of `node-inspector` might fail in unsupported environments, removing it altogether is probably even a better option.

ponelat:
I think let's remove it altogether.
There may be a script in package.json the uses it. But its purely for convenience, and there's no need to burden build pipelines as a result.

@buunguyen @shockey are you happy with removing it?
If so, the only addition I'd ask for in this PR is to remove any `package.json` scripts that use `node-inspector`.

buunguyen:
@ponelat agreed. @fgnass would you please remove it as well as the associated npm command that use it?

fgnass:
Done.

buunguyen:
Thank you!

Issue #1006, opened 2017-04-07
A parameter in body that is json has to be explicitly serialized


dwilson6:
I've noticed that if I have a POST endpoint with a parameter in the body that is of type object and I call that operation with the tags interface like client.apis.default.operation({param_name: {}}) (I explicitly set the Content-Type header to application/json), I get an error from my server: `unexpected token o in JSON at position 1...`

To get it to work, I have to serialize the value of the parameter explicitly `client.apis.default.operation({param_name: JSON.stringify({})})`.

Would it be reasonable to automatically convert an object to a string if the content type is application/json and it has an "in: body" parameter and the value of said parameter is an object?  

Also, if an api endpoint only has one parameter, what do you think about allowing you to call the operation with the interface `operation(value)` instead of `operation({param_name: value})`? If an operation only has one parameter, you could check if the argument passed to the operation is an object with a property matching the parameter name (existing behavior) and if not then just use the passed in argument as the value for the parameter (after validating the type).  Are there problems with that approach or does that make it too complicated?

fehguy:
You are using the 3.x version I take it?  This is a regression

dwilson6:
Yes, I am

fehguy:
We'll get it addressed shortly

buunguyen:
Fixed in https://github.com/swagger-api/swagger-js/pull/1034

PR #1012, opened 2017-04-10
Serialize body parameter value #1006

dwilson6:
This fixes the issue in #1006. There was a test case for this but it wasn't testing what it said it was testing.

### DIFF
--- a/src/execute.js
+++ b/src/execute.js
@@ -104,11 +104,13 @@ export function buildRequest({
       const builder = parameterBuilders[parameter.in]
       let value
 
-      if(parameter.in === 'body' && parameter.schema && parameter.schema.properties) {
+      //use the parameter value from the parameters object if it is set
+      if(parameter.name && parameters[parameter.name]) {
+        value = parameters[parameter.name];
+      } else if(parameter.in === 'body' && parameter.schema && parameter.schema.properties) {
+        //if this parameter is in the body and has an object schema, use the whole parameters object
         value = parameters
       }
-
-      value = (parameter && parameter.name && parameters[parameter.name]) || value
 
       if (typeof parameter.default !== 'undefined' && typeof value === 'undefined') {
         value = parameter.default
--- a/src/execute.js
+++ b/src/execute.js
@@ -108,10 +108,14 @@ export function buildRequest({
         value = parameters
       }
 
-      value = parameter && parameter.name && parameters[parameter.name]
+      value = (parameter && parameter.name && parameters[parameter.name]) || value
 
       if (typeof parameter.default !== 'undefined' && typeof value === 'undefined') {
         value = parameter.default
+      }
+
+      if (parameter.in === 'body' && typeof value === 'object') {
+        value = JSON.stringify(value)
       }
 
       if (typeof value === 'undefined' && parameter.required && !parameter.allowEmptyValue) {
--- a/test/execute.js
+++ b/test/execute.js
@@ -536,7 +536,7 @@ describe('execute', () => {
     })
   })
 
-  it('should pass through the parameters object as the value if parameter is in the body', function () {
+  it('should use the whole parameters object as the parameter value if parameter is in the body', function () {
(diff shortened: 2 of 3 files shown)


src/execute.js
```diff
@@ -108,12 +108,16 @@ export function buildRequest({
         value = parameters
       }
 
-      value = parameter && parameter.name && parameters[parameter.name]
+      value = (parameter && parameter.name && parameters[parameter.name]) || value
```
dwilson6: The if statement above was dead code because of this line. This change allows you to use the parameters object as the value of a body parameter (parameters = {mydata: ''}) instead of having to have parameters be: {body_param_name: {mydata: ''}}

buunguyen: If the `if` above is dead code, should it be removed?  And if it's removed, what value does `value` have before this line?

Also, `paramater` can't be null, can it? If yes, no need to check for `parameter`.

Should rewrite this thing to be consistent with the stuff below.

```
if (condition) {
  value = ...
}
```

test/execute.js
```diff
@@ -529,13 +529,60 @@ describe('execute', () => {
...
-        body: {
+        body: JSON.stringify({
           json: 'rulez'
-        }
+        })
       })
     })
   })
 
+  it('should pass through the parameters object as the value if parameter is in the body', function () {
```
buunguyen: I don't understand what "pass through" here means. And what value does this test add compared to existing one such as "should build a request for all given field"?

src/execute.js
dwilson6: The if statement above is not dead code anymore.  I can refactor it a bit to make it more readable.  Parameter could technically be null since it is just iterating through an array though there is code above this line that would throw an error if parameter is null.  Also, that code checking parameter was there before my change so I assumed there is a reason it is there.  I can remove it since it seems pointless

test/execute.js
dwilson6: Yeah, that is a little confusing.  I mean that the whole `parameters` object will be used as the parameter value instead of `parameters[parameter.name]`. I'll clarify the wording.  To your second question, the difference in this test is the parameters object in the buildRequest call below.  

In the other tests it is:
`parameters: {
          ...,
          body: {
        name: 'johny',
        id: '123'
      },
        }`

in this test it is:
`parameters: {
        name: 'johny',
        id: '123'
      }`

This allows you to call operations like `client.apis.default.operation({name: 'johny', id: '123'})` if there is a body parameter.

test/execute.js
```diff
@@ -529,13 +529,60 @@ describe('execute', () => {
...
-        body: {
+        body: JSON.stringify({
           json: 'rulez'
-        }
+        })
       })
     })
   })
 
+  it('should use the whole parameters object as the parameter value if parameter is in the body', function () {
```
buunguyen: @webron this is about how the spec works. I'm not sure if what is proposed is valid. Please check.

webron: I don't really know how to read the test so I don't understand what it does.

buunguyen: See https://github.com/swagger-api/swagger-js/issues/1006.

> Also, if an api endpoint only has one parameter, what do you think about allowing you to call the operation with the interface operation(value) instead of operation({param_name: value})? If an operation only has one parameter, you could check if the argument passed to the operation is an object with a property matching the parameter name (existing behavior) and if not then just use the passed in argument as the value for the parameter (after validating the type). Are there problems with that approach or does that make it too complicated?

Maybe @fehguy can chime in.

webron: Doesn't sound like a spec issue, but rather a simplified interface for call execution. It does make sense to me, but would rather hear from @fehguy as well.

dwilson6: @fehguy @webron anything new on this? The code for this feature was already there but was broken

buunguyen:
Seems obsolete compared to https://github.com/swagger-api/swagger-js/pull/1034.

Issue #1002, opened 2017-04-07
throw an error when spec at URL cannot be loaded


fehguy:
If you can't load the spec, you get an empty client.  Instead, throw an error

buunguyen:
Fixed https://github.com/swagger-api/swagger-js/pull/1015/files

PR #1015, opened 2017-04-11
Fix error handling

### LINKED ISSUE #1014: Catch errors
Seems like #1002

Trying to create a client, when swagger.json can't be loaded:

```
Swagger('/not.exist.api/swagger.json')
  .then((client) => {
    console.log('client', client);
  })
  .catch((error) => {
    console.log('error', error);
  });
```

But instead of catch 404 error, you get an client with spec like this
`
["H",
"T",
"T",
"P",
" ",
"4",
"0",
"4",
" ",
"N",
"o",
"t",
" ",
"F",
"o",
"u",
"n",
"d"]
`

buunguyen:
This fixes #1014 and #1002.

@fehguy please take a look. One thing of note: because fetch is shared for both spec loading & operation execution, the change affects operations, see https://github.com/swagger-api/swagger-js/compare/master...buunguyen:fix-error-handling-2?expand=1#diff-7e3bd8e7059a21bfd255649cbfc84b33R130 (must change from `then` to `catch`). This can be considered a breaking change from semver perspective. UI/editor might need to change.

### DIFF
--- a/src/resolver.js
+++ b/src/resolver.js
@@ -11,7 +11,9 @@ export function makeFetchJSON(http) {
         Accept: 'application/json'
       }
     })
-    .then(res => res.body)
+    .then((res) => {
+      return res.body
+    })
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/resolver.js
```diff
@@ -30,29 +32,29 @@ export default function resolve({http, fetch, spec, url, baseDoc, mode, allowMet
...
-    plugins: plugs,
-    allowMetaPatches // allows adding .meta patches, which include adding `$$ref`s to the spec
-  }).then(normalizeSwagger)
+    const plugs = [plugins.refs]
+
+    if (mode !== 'strict') {
+      plugs.push(plugins.allOf)
+    }
+
+    // mapSpec is where the hard work happens, see https://github.com/swagger-api/specmap for more details
```
webron: Hmm... not anymore ;)

buunguyen:
@webron this is technically breaking change of a corner case. Given we just release, should we ignore semver rule?

webron:
@buunguyen I think so. Based on our chat, we might introduce additional such changes. Until we stabilize it more, it's ok. We should note that well in the release notes though.

PR #1016, opened 2017-04-14
Issue 995

bodnia:


test/execute.js
```diff
@@ -454,14 +454,143 @@ describe('execute', () => {
...
-        headers: {
-          'content-type': 'application/josh',
-        },
+        headers: {},
         credentials: 'same-origin',
         method: 'GET'
       })
     })
 
+    it('should not add content-type with no form-data or body param', function () {
```
ponelat: should not? I see `content-type` lower down

ponelat: I would like to see a test, where _no_ contentType is added. As one of our current issues, is that GET requests get a `content-type: application/json`. And I'd like to make sure that doesn't happen in swagger-js ( I'm pretty sure it doesn't, but still ).

test/execute.js
```diff
@@ -441,7 +441,7 @@ describe('execute', () => {
       })
     })
 
-    it('should handle requestContentType', function () {
+    it('should not add content-type with no form-data or body param', function () {
```
bodnia: @ponelat this test is about not to set content type

bodnia:
Issue https://github.com/swagger-api/swagger-ui/issues/2909

test/execute.js
ponelat: ok... but this test includes a `requestContentType`. If that is present, it should always set the header.

ponelat: With the issue in https://github.com/swagger-api/swagger-ui/issues/2909, there is likely a `requestContentType` being passing in from swagger-ui. And that shouldn't happen

bodnia: Here https://github.com/swagger-api/swagger-js/pull/1016/files#diff-2932c1430f43e1d22b8cf8b98fa5d2a8R127 first we check if there is body or form and only after that - if there is content-type passed from UI

ponelat: I see that. But we shouldn't :stuck_out_tongue: if there is a `requestContentType` we must use it. Its an override provided by the user ( or swagger-ui in our case ).

ponelat: you're right @bodnia it does _only_ make sense if there is a body. The user can override the `content-type` header in other ways, if need be IIRC.

bodnia:
@ponelat merge?

Issue #1018, opened 2017-04-18
Getting "failed to parse JSON/YAML response" in IE10


estermer:
I am having no issues in any other browser except IE 10 and down. IE 11 and up works.

When I catch an error when making a request using swagger-client in IE10, I get this error "failed to parse JSON/YAML response", I am having an extremely hard time narrowing down where i could find a solution to this error. 

I was wondering is anyone had experienced this or if there is a direction someone can point me in to find a way to narrow down the problem?

webron:
Which version do you use?

estermer:
@webron I am using `"swagger-client": "^2.1.17"` but recently upgraded to the newest version to see if it will help and it did not

webron:
The newest version (3.X) doesn't support IE10 (see [prerequisites](https://github.com/swagger-api/swagger-js#prerequisites)).

estermer:
Would there be a reason it doesn't work on 2.1?

webron:
I don't believe so.

estermer:
This is my set up of initializing the swagger client:

```js
import { API_URI } from '../config/app_config'; // '/accountservice/swagger.json'
import Swagger from 'swagger-client'; //   "swagger-client": "^2.1.17"

export const buildAccountServiceClient = () => {
  const swagger = new Swagger({
    url: API_URI,
    usePromise: true,
  });

  // Reconfigure swagger client to override service path if we're using a reverse proxy:
  // /accountservice/swagger.json -> /accountservice
  // Originally tried setting basePath to null, undefined, and '', but that didn't work
  let basePath;
  if (API_URI.startsWith('/')) {
    basePath = API_URI.substring(0, API_URI.lastIndexOf('/'));
    swagger.then((client) => {
      client.setBasePath(basePath);
      if (typeof(window) !== 'undefined') {
        // use current protocol, so either http or https
        client.setSchemes([window.location.protocol.slice(0, -1)]);
      }
    });
  }
  return swagger;
};

// cache swagger client, so we only load swagger.json once:
let accountServiceSwagger = null;
export const accountServiceClient = () => {
  if (accountServiceSwagger === null) {
    accountServiceSwagger = buildAccountServiceClient();
  }
  return accountServiceSwagger;
};
```

Then when I implement the swaggerAPI in my redux actions is where I get the error:

```js
// ./actions/auth.js
// Connect to the API
const swaggerApi = accountServiceClient();


// Log User In
export function logUserIn(credentials) {
[...]

PR #1029, opened 2017-04-24
Fix for IE10 failed to parse JSON/YAML response

estermer:
lib/client.js line 129 changed to this.url = window.location.protocol + // + window.location.host + this.url; to fix error in IE10 and other browsers that has a window object that does not have an 'origin' key in it's location.

explanation in my answer here:

http://stackoverflow.com/questions/43458634/ie10-swagger-error-failed-to-parse-json-yaml

@webron 
Referencing my issue #1018

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -126,7 +126,7 @@ SwaggerClient.prototype.initialize = function (url, options) {
   if(this.url && this.url.indexOf('http:') === -1 && this.url.indexOf('https:') === -1) {
     // no protocol, so we can only use window if it exists
     if(typeof(window) !== 'undefined' && typeof(window.location) !== 'undefined') {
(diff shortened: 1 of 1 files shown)


webron:
@estermer - thanks for the PR. Can you add a test that shows the failing behavior before?

estermer:
@webron yeah I can add that.

buunguyen:
@estermer just want to check if you have a chance to follow up with this.

estermer:
@buunguyen  I have sort of. I am struggling with creating a test for it, but I also found a easier fix possibly. I got caught up with another task at work , but the fix is easy so I can maybe crank it out this morning if it means I don't have to write a test hahaha

shockey:
Closing this since it's a 2.x change.

PR #1034, opened 2017-04-29
stringify request.body in #execute

ponelat:
and NOT in buildRequest, so as to allow other code to modify the body,
before finally getting processed by execute


webron:
Does this resolve https://github.com/swagger-api/swagger-js/issues/1006?

buunguyen:
I think so. #1006 also asks for an enhancement, but as far as being a bug, this should fix that bug.

ponelat:
I forgot to test FormData with this... probably need to add a test + ensure that didn't break. :(

PR #1038, opened 2017-05-02
Fix cyclic ref problem when baseDoc is specified

buunguyen:
Fix root cause of https://github.com/swagger-api/swagger-ui/issues/2892#event-1064243022. Pretty subtle bug. The fix isn't elegant, but that's to avoid changing the internal structure too much, better leave that for more significant refactoring.

cc @webron @shockey

### DIFF
--- a/src/resolver.js
+++ b/src/resolver.js
@@ -38,7 +38,9 @@ export default function resolve({http, fetch, spec, url, baseDoc, mode, allowMet
   return doResolve(spec)
 
   function doResolve(_spec) {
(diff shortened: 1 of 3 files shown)


webron:
Does this mean #2892 can be closed?

buunguyen:
Only after @shockey release new version of JS and UI

shockey:
Looks like this fixes https://github.com/swagger-api/swagger-ui/issues/3012#issuecomment-298762107 as well.

PR #1039, opened 2017-05-03
Include .response property on Errors as documented

kevinoid:
The [HTTP Client section of README.md](https://github.com/swagger-api/swagger-js/blob/v3.0.8/README.md#http-client) documents the Error object as having a `.response` property with "response or null".  It does not appear that this feature has been implemented, as demonstrated by the following code:

```js
'use strict'

const assert = require('assert')
const http = require('http')
const SwaggerClient = require('swagger-client')

const TEST_HOST = 'localhost'
const TEST_ID = 123
const TEST_PORT = 8888

http
  .createServer((req, res) => {
    assert.strictEqual(req.url, '/data')
    res.writeHead(400, {'Content-Type': 'application/json'})
    res.end(JSON.stringify({message: 'Server-defined error message'}))
  })
  .listen(TEST_PORT, TEST_HOST, () => {
    // Usage as documented in HTTP Client section of README.md
    SwaggerClient.http({url: `http://${TEST_HOST}:${TEST_PORT}/data`})
      .catch((err) => {
        console.log('HTTP Client err.response', err.response)
      })

    // Usage with defined spec including 400 response
    new SwaggerClient({
      spec: {
        swagger: '2.0',
        info: {
          title: 'Example Error API',
          version: '1.0.0'
        },
        host: `${TEST_HOST}:${TEST_PORT}`,
        schemes: [
          'http'
        ],
        paths: {
          '/data': {
            get: {
              operationId: 'getData',
              responses: {
                200: {
                  description: 'Some data',
                  schema: {
                    type: 'string'
                  }
                },
                400: {
                  description: 'An error',
                  schema: {
                    type: 'object',
                    properties: {
                      message: {
                        type: 'string'
                      }
                    }
                  }
                }
              }
            }
          }
        }
      }
    })
      .then(client => client.apis.default.getData())
      .catch((err) => {
        console.error('API method err.response', err.response)
      })
  })
```
which prints

    HTTP Client err.response undefined
    API method err.response undefined

This PR adds the `.response` property to the `Error` object returned by `fetch` when the response can be read or `.responseError` if the response could not be read.

Note that it does not pass the response to `responseInterceptor` in the case of errors.  Although I think it would be preferable to do so, it would be an API change which could break code which expects the current behavior and does not handle error responses correctly, so I've left it out for now.

Thanks for considering,
Kevin

### DIFF
--- a/src/http.js
+++ b/src/http.js
@@ -37,17 +37,29 @@ export default function http(url, request = {}) {
   }
 
   return fetch(request.url, request).then((res) => {
-    if (!res.ok) {
-      const error = new Error(res.statusText)
-      error.statusCode = error.status = res.status
-      throw error
-    }
-    return self.serializeRes(res, url, request).then((_res) => {
+    const serialized = self.serializeRes(res, url, request).then((_res) => {
       if (request.responseInterceptor) {
         _res = request.responseInterceptor(_res) || _res
       }
       return _res
     })
+
+    if (!res.ok) {
+      const error = new Error(res.statusText)
+      error.statusCode = error.status = res.status
+      return serialized.then(
+        (_res) => {
+          error.response = _res
+          throw error
+        },
+        (resError) => {
+          error.responseError = resError
+          throw error
+        }
+      )
+    }
+
+    return serialized
   })
 }
 
--- a/test/http.js
+++ b/test/http.js
@@ -38,6 +38,67 @@ describe('http', () => {
     })
   })
 
+  it('should include status code and response with HTTP Error', () => {
+    xapp = xmock()
+    xapp.get('http://swagger.io', (req, res) => res.status(400).send('hi'))
+
+    return http({
+      url: 'http://swagger.io'
+    })
+    .then(
+      (res) => {
+        throw new Error('Expected rejection for HTTP status 400')
+      },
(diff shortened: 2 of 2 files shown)


buunguyen:
Thanks for this. I think we should pass the response to the interceptor. Also suggest refactor a bit:

```js
return self.serializeRes(res, url, request).then((_res) => {
   // call interceptor
   // if !res.ok return error
   // return _res
})
```

kevinoid:
@buunguyen Sounds good for the interceptor.  From the snippet you posted, if `serializeRes` encounters an Error, it will be returned instead of the server Error.  Is that what you intended?

buunguyen:
@kevinoid you're right, I didn't notice the error handler in !ok case. Keep your current structure then.

kevinoid:
PR updated to call `responseInterceptor` for all responses.  Let me know what you think.

buunguyen:
LGTM. Thank you!

kevinoid:
Great.  Thanks @buunguyen!

RomanGotsiy:
@buunguyen do you have any plans when the new package is released?
Thanks

webron:
Today.

PR #1041, opened 2017-05-03
Do not depend the Authorization header on the token type.

### LINKED ISSUE #1040: OAuth2 incorrect Authorization header
I'm using the swagger-js code (within swagger-ui) against an OAuth2 implementation.

I run into some problems which in my opinion should be fixed within the swagger-js code.

From the OAuth2 token endpoint I get a response with the following data:
```
{
  "access_token":"access_token",
  "token_type":"bearer",
  "expires_in":3600,
  "refresh_token":"value of refresh token"
}
```

The swagger-js code makes a request with the following header:
`Authorization: bearer access_token`

The OAuth2 implementation does not accept this header, because of the lowercase "bearer" scheme.

As per RFC 6749 section 4.2.2 the token_type value is case insensitive. So it can be bearer, Bearer, BeArEr, etc.
But per RFC 6750 section 2.1 the Authorization header should have the authorization scheme 'Bearer'.

The authorization scheme should not depend on the token_type returned when fetching an access token.

stefangr:
The token type is case insensitive.
The authorization scheme of the Authorization header is case sensitive.

Fixes #1040

### DIFF
--- a/src/execute.js
+++ b/src/execute.js
@@ -241,7 +241,6 @@ export function applySecurities({request, securities = {}, operation = {}, spec}
       const schema = securityDef[key]
       const {type} = schema
       const accessToken = token && token.access_token
-      const tokenType = token && token.token_type
 
       if (auth) {
         if (type === 'apiKey') {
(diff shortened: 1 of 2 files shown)


buunguyen:
LGTM. @bodnia what do you think?

PR #1043, opened 2017-05-04
replace isomorphic-fetch dependency by fetch-everywhere

jpoutrin:
I came across some nasty issue on react native.
here is a solution that was suggested in some other projects.

tests are passing fine.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "btoa": "1.1.2",
     "deep-extend": "^0.4.1",
     "fast-json-patch": "^1.1.8",
-    "isomorphic-fetch": "2.2.1",
+    "fetch-everywhere": "^1.0.2",
     "isomorphic-form-data": "0.0.1",
     "js-yaml": "^3.8.1",
     "lodash": "4.16.2",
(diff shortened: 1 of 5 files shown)


buunguyen:
Hi, thanks for the PR. I want to make sure I understand the motivation of the lib change. Did you mean the old lib doesn't support React Native? Or what is the issue?

jpoutrin:
@buunguyen yes the current does not support react native 
the solution was inspired by https://github.com/agraboso/redux-api-middleware/pull/120
the current implementation does not initialize some global variables properly.

jpoutrin:
@buunguyen please do not merge the change.
The issue  seems deeper than that.
I messes up the final react-native bundle for some reason.
it seems to be a webpack bundling issue. (one might not exclude the other).
I am no expert in webpack so this should be put on hold or discarded...

symptoms in my app when debugging:
`Failed to print error:  regeneratorRuntime.async is not a function`

symptoms without:
client.apis is empty

if someone has an idea would appreciate.

sorry for that.

jpoutrin:
I opened an issue for this:
https://github.com/swagger-api/swagger-js/issues/1047

PR #1048, opened 2017-05-10
Rewire operationIds

### LINKED ISSUE #1003: generated operation ids are very ugly
It is true that operation ids are not required, but what we are currently generating is quite ugly.  For example:

```yaml
paths:
  /test:
    get:
      # no operation id
      parameters: []
      responses:
        200:
          description: well done
```

Generates this for `operationId`:

```
get-/test
```

shockey:
⚠️ **Work in progress, don't merge - but please do review code.**

Fixes #1003, paves the way for https://github.com/swagger-api/swagger-ui/issues/2884.

Changelog forthcoming.

#### TODO

- [ ] Confirm this doesn't break anything in UI; coordinate changes with UI PR.

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -27,15 +27,13 @@ export function getOperationRaw(spec, id) {
     return null
   }
 
-  const normalized = normalizeSwagger({ spec }, { addOpIds: true })
 
-
-  return findOperation(normalized.spec, ({pathName, method, operation}) => {
+  return findOperation(spec, ({pathName, method, operation}) => {
     if (!operation || typeof operation !== 'object') {
       return false
     }
 
-    const {operationId} = operation
+    const operationId = opId(operation, pathName, method)
 
     if (operationId && operationId === id) {
       return true
@@ -86,7 +84,7 @@ export function eachOperation(spec, cb, find) {
   }
 }
 
-export function normalizeSwagger(parsedSpec, { addOpIds } = {}) {
+export function normalizeSwagger(parsedSpec) {
   const {spec} = parsedSpec
   const {paths} = spec
   const map = {}
@@ -118,12 +116,6 @@ export function normalizeSwagger(parsedSpec, { addOpIds } = {}) {
         }
         else {
           map[oid] = [operation]
-        }
-
-        if(addOpIds) {
-          // oid ( via opId() ) will preserve the set operationId if it is well-formed,
-          // i.e. it isn't purely whitespace
-          operation.operationId = oid
         }
 
         Object.keys(map).forEach((op) => {
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -8,7 +8,7 @@ const escapeString = str => {
 // Strategy for determining operationId
 export function opId(operation, pathName, method = '') {
   const idWithoutWhitespace = (operation.operationId || '').replace(/\s/g, '')
-  if(operation.operationId && idWithoutWhitespace.length) {
+  if(idWithoutWhitespace.length) {
     return escapeString(operation.operationId)
   } else {
     return idFromPathMethod(pathName, method)
@@ -35,11 +35,7 @@ export function getOperationRaw(spec, id) {
 
     const operationId = opId(operation, pathName, method)
 
-    if (operationId && operationId === id) {
-      return true
-    }
-
-    return false
+    return operationId && operationId === id
   })
 }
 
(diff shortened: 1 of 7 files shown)


buunguyen:
Test fail.

src/helpers.js
```diff
@@ -1,23 +1,24 @@
...
+const escapeString = str => {
+  return str.replace(/[^\w]/gi, '_')
+}
 
 // Strategy for determining operationId
 export function opId(operation, pathName, method = '') {
-  return operation.operationId
-    ? operation.operationId : idFromPathMethod(pathName, method)
+  const idWithoutWhitespace = (operation.operationId || '').replace(/\s/g, '')
+  if(operation.operationId && idWithoutWhitespace.length) {
```
buunguyen: I think only check `if (idWithoutWhitespace)` is enough.

src/helpers.js
```diff
@@ -26,19 +27,19 @@ export function getOperationRaw(spec, id) {
...
+
   return findOperation(spec, ({pathName, method, operation}) => {
     if (!operation || typeof operation !== 'object') {
       return false
     }
 
-    const {operationId} = operation
+    const operationId = opId(operation, pathName, method)
 
     if (operationId && operationId === id) {
```
buunguyen: Should just `return operationId && operationId === id` here.

src/interfaces.js
```diff
@@ -83,9 +83,15 @@ export function mapTagOperations({spec, cb = nullFn, defaultTag = 'default'}) {
       // Id already exists?
       if (typeof tagObj[id] !== 'undefined') {
         // Bump counter ( for this operationId )
-        operationIdCounter[id] = (operationIdCounter[id] || 0) + 1
+        let originalCounterValue = (operationIdCounter[id] || 1)
```
buunguyen: It took me a bit to decipher the logic. Maybe clearer if refactor:

```js
const originalCounterValue = (operationIdCounter[id] || 1)
tagObj[`${id}${originalCounterValue}`] = tagObj[id]
tagObj[`${id}${originalCounterValue + 1}`] = cbResult
delete tagObj[id]
operationIdCounter[id] = originalCounterValue + 1
```

Question. Say there are 3 'abc'. The 2nd 'abc', this logic will run, both will get renamed to 'abc1' and 'abc2'. The last 'abc', it will still be 'abc', as opposed to 'abc3'. Is this expected?

test/client.js
```diff
@@ -39,7 +39,7 @@ describe('http', () => {
       }
 
       fs.exists(filename, function (exists) {
-        if (exists) {
+        if (exists && filename[filename.length - 1] !== "/") {
```
buunguyen: Single quote please. Also, what's the intention here? To check if filename is really a file? If so, we should probably use Node API to check.

src/helpers.js
```diff
@@ -1,23 +1,24 @@
...
 }
 
 
 // Create a generated operationId from pathName + method
 export function idFromPathMethod(pathName, method) {
-  return `${toLower(method)}-${pathName}`
-}
-
-// Extract pathName + method from generated operationId
-export function pathMethodFromId(id) {
```
buunguyen: I remember seeing `pathMethodFromId` being used in SwaggerUI. So this change might need to be coordinated with UI as well.

src/interfaces.js
shockey: You can see a test for this logic here: https://github.com/shockey/swagger-js/blob/bug/1003-operation-ids/test/interfaces.js#L345

Two operations with the id `getOne` would yield a `getOne1` and `getOne2`, but no `getOne`. The test confirms this 😄

test/client.js
shockey: I removed this and all tests continued to pass.

I think this was a transient issue related to the failing XMock test.

src/interfaces.js
buunguyen: I meant when there are 3 "getOne". The 3rd one should be "getOne3" right?

shockey: @buunguyen Ah, I understand you now. I just pushed a fix for this case, and expanded the test case to match: https://github.com/swagger-api/swagger-js/pull/1048/commits/8dfc734f3641b821dfc410f19678b7ad01657666

shockey: Note that the test also makes sure that the tags are in order and not being inadvertently overwritten, by use of the `count` variable in the test.

src/helpers.js
shockey: Looks like it's not currently in the codebase 😄

Issue #1022, opened 2017-04-20
Reopen 1010: Default request Content-Type header and parameter in body as object


srihakum:
I do not have permission to reopen 1010 hence new issue. Sorry for duplication. With the latest fix it seems the consumes is being used to fill up Content-Type, However if the spec does not have consumes it still remains the same. In 2.x it was getting defaulted to application/json. The body serialisations issue also still exists.

This issue is in further reference to https://github.com/swagger-api/swagger-js/issues/991

With version 3.0.5 a swagger client instance tries to make an API request the following issues are observed 
1. Content-Type header is not added in the request. 
2. If a parameter used in HTTP body is added to the request the client expects it to be in the string format.

However while using 2.0.32 there is no need to set these explicitly. So is there a way we can get back the old behaviour with latest version, like adding "application/json" as default content type and the adding body parameter as JSON object rather than string.

Code example:
```
var Swagger = require('swagger-client');
new Swagger({
    "url":"https://staging.cloud-elements.com/elements/api-v2/elements/44/docs"
})
    .then(function (client) {
        client.apis.zohocrm.createAccount({"Authorization":"...","body":{"Account Name":"XYZ1"}})
            .then(function(response){
                console.log(response);
            })
            .catch(function (error) {
                console.log(error);
            });
    }).catch(function (error) {
    console.log(error);
});
```

webron:
Thanks, @srihakum. For future reference, if you feel a ticket was wrongly closed, just add a comment to it and we'll reopen it. 

#1010 does mention that there are two separate tickets covering it, which are not included in 3.0.5.
In fact, the one related to the headers was resolved only yesterday, so it's not published yet.

If you'd like to try it, clone the repository, build it locally, and use it in your app. We have a release slated for tomorrow, if you'd rather wait, but if you can spare the time to test it and validate it before we publish it, we'd appreciate it.

srihakum:
@webron I just tried today with 3.0.0.7 but it this issue is not fixed.

kwv:
I'm running into the same issue.  Given the [spec](https://raw.githubusercontent.com/openstf/stf/master/lib/units/api/swagger/api_v1_generated.json) consumes `application/json` I would expect an empty post to honor that. 

When calling using v3 of swagger-js 
`client.apis.user.remoteConnectUserDeviceBySerial( { 'serial': '1234'] })`
it maps the request to:
/user/devices/{serial}/remoteConnect

and fails:

```
response:
   { ok: false,
[...]

PR #1052, opened 2017-05-12
honors spec content-type for POST requests when no body.  see https:/…

kwv:
https://github.com/swagger-api/swagger-js/issues/1022

### DIFF
--- a/src/execute.js
+++ b/src/execute.js
@@ -127,7 +127,7 @@ export function buildRequest({
   // Add securities, which are applicable
   req = applySecurities({request: req, securities, operation, spec})
 
-  if (req.body || req.form) {
+  if (req.body || req.form || req.method === 'POST') {
     if (requestContentType) {
       req.headers['content-type'] = requestContentType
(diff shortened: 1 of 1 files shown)


shockey:
The code looks good to me, but based on the discussion in #1022, it's not clear to me if this a use case we're going to support (sending content-type for empty POSTs).

@webron, would appreciate your input here 😄

marjan-georgiev:
This feature is the only thing preventing us from using this library in our application. It would be great if it got merged in.

shockey:
@webron, bump

webron:
I'm not convinced this is the right solution, and will be closing the PR without merging. Merging this PR will cause, in some cases, completely invalid requests. If you take the `application/json` example, you're telling the API producer that you're sending it an `application/json`- making it try to parse an empty string as a JSON which is just not a valid JSON. I'd be happy to reconsider if someone shows concrete information as to why it should be allowed, but until then...

PR #1059, opened 2017-05-18
[Fix] Wrong isomorphic-fetch import

Hugome:
Change the import method for fetch (Recommended by : https://github.com/matthew-andrews/isomorphic-fetch).

I have a problem with the current exported `fetch` of `isomorphic-fetch` : 
I want to mock fetch for unit testing (Not just swagger part), and actually its impossible because of webpack.
Ex : 
This es6 code :
```js
import fetch from 'isomorphic-fetch'
```
will be transpiled by babel, transformed by webpack and the result something like this : 
```js
var _isomorphicFetch2 = _interopRequireDefault(_isomorphicFetch);
var _isomorphicFetch = __webpack_require__(15);
// And the webpack require =>
/* 15 */
/***/ (function(module, exports) {
module.exports = require("isomorphic-fetch");
/***/ })
```
If we look at es6 swagger code for fetch
```js
  return fetch(request.url, request).then((res) => {
    const serialized = self.serializeRes(res, url, request).then((_res) => {
```
The transform result is
```js
return (0, _isomorphicFetch2.default)(request.url, request).then(function (res) {
    var serialized = self.serializeRes(res, url, request).then(function (_res) {
```
but `_isomorphicFetch2.default` is the export of package `isomorphic-fetch` at loading time (In browser this is the native fetch function), so it will use `_isomorphicFetch2.default` not `global.fetch` and you cannot mock the fetch.

### DIFF
--- a/src/http.js
+++ b/src/http.js
@@ -1,4 +1,4 @@
-import fetch from 'isomorphic-fetch'
+import 'isomorphic-fetch'
 import qs from 'qs'
 import jsYaml from 'js-yaml'
 import assign from 'lodash/assign'


buunguyen:
In browser, can't you just override windows.fetch before this script is loaded?

Hugome:
This is possible but you cannot revert mock, because `_isomorphicFetch2.default` remember the `global.fetch` at loading time, and in tests you can need to do :
- Mock fetch
- Test
- "Unmock" fetch
- Test
- Test
- Test
- Mock fetch
- Test
- "Unmock" fetch

(And if you use es6 import, you cannot replace the function before import)

src/http.js
```diff
@@ -1,4 +1,5 @@
-import fetch from 'isomorphic-fetch'
+import 'es6-promise/auto'
```
buunguyen: Why is this needed?

Hugome: This is recommended by the maintainer of `isomorphic-fetch`, this provide a polyfill for es6 promises.
But i dont think it is really really important, because many browser have implemented the ES6 Promises
- https://kangax.github.io/compat-table/es6/#test-Promise
- https://caniuse.com/#feat=promises
So i remove the import and dependency :)

PR #1061, opened 2017-05-19
Avoid encoding form data value

buunguyen:
Fixes https://github.com/swagger-api/swagger-ui/issues/3101

### DIFF
--- a/test/http.js
+++ b/test/http.js
@@ -174,6 +174,29 @@ describe('http', () => {
         url: 'https://swagger.io?one=1&two=2&three=3'
       })
     })
+
+    it.only('should not encode form-data', function () {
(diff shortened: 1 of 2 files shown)


test/http.js
```diff
@@ -174,6 +174,29 @@ describe('http', () => {
         url: 'https://swagger.io?one=1&two=2&three=3'
       })
     })
+
+    it.only('should not encode form-data', function () {
+      const FormData = require('isomorphic-form-data')
```
buunguyen: The Node version of FormData doesn't expose standard methods like `get`, `keys`... Easiest way to test is monkey patching `append`... Not great, but quick and simple.

PR #1065, opened 2017-05-24
Remove instance of .only in a test

ponelat:
just a typo from a previous PR


ponelat:
Merging, please undo if that breaks protocol.
cc @webron @buunguyen

PR #1068, opened 2017-05-26
Add the normalized operation id, back into the spec. In an edgecase

ponelat:


shockey:
@ponelet, the changes look good - but what edge use case does this cover?

Feel free to take this conversation out-of-band if needed.

ponelat:
Hey @shockey 
The edgecase ( where I bumped into this ), was in swagger-ui's try-it-out feature.
It would die, if the `operationId` of the endpoint had `spaces` or `dashes`.

webron:
😱

shockey:
@ponelat, I'm pretty sure I fixed this in https://github.com/swagger-api/swagger-js/pull/1070. 

Can you take a look at the changes there, and the related issue in UI https://github.com/swagger-api/swagger-ui/issues/3146?

ponelat:
Yup #1070 looks like it'll do it.

webron:
So this one can be closed? :)

ponelat:
@webron @shockey 
even though the main issue is fixed. This PR addresses an inconsistency. Sometimes we update the `operationId` ( ie: when there are duplicates ). This PR will ensure we _always_ update `operationId`.

I'm not the biggest fan of updating it at all, but I understand its required for compatibility ( amongst reasons ). If we do update, we might as well _always_ do it.

Since I no longer require this. I'll leave it up to you guys to decide.

shockey:
In order to satisfy what we're doing in https://github.com/swagger-api/swagger-ui/pull/3152, I added some logic to add the unaltered operationId to the operaiton object as `__originalOperationId`.

@ponelat, I agree with the sentiment.

PR #1079, opened 2017-06-09
Add `operation not found` exception

extempl:


src/execute.js
```diff
@@ -219,7 +236,7 @@ export function baseUrl({spec, scheme, contextUrl = ''}) {
 // Add security values, to operations - that declare their need on them
 export function applySecurities({request, securities = {}, operation = {}, spec}) {
   const result = assign({}, request)
-  const {authorized = {}, specSecurity = {}} = securities
+  const {authorized = {}, specSecurity = []} = securities
```
shockey: Why was this changed?

src/execute.js
```diff
@@ -48,6 +54,10 @@ export function execute({
 
   const request = self.buildRequest({spec, operationId, parameters, securities, ...extras})
 
+  if (request.error) {
```
shockey: Why are we throwing in `execute`, via the `request.error` switch, instead of throwing the error directly in `buildRequest`?

src/execute.js
extempl: Because of variable type. When `{}` there will be exception  on `forEach` it: 'forEach is not a function'
    const {authorized = {}, specSecurity = []} = securities    
    const security = operation.security || specSecurity
  
    security.forEach

src/execute.js
extempl: Good point. That was done due to approach I used before. Updated

src/execute.js
shockey: aha. Good catch, thanks!

extempl:
@shockey Test is added.

PR #1082, opened 2017-06-12
added browser notes and example into readme

filinus:


webron:
Thanks for the PR, @filinus!

Issue #1025, opened 2017-04-21
Cookies not sent when requesting $ref in swagger.json 


glipman:
Ticket #970 described a problem where cookies where not send because credentials: "same-origin" was missing.
In my model I have several $ref references to external files. The HTTP-request for these requests still don't send the ASPSESSIONID (or any) cookies.

My guess: makeFetchJSON in resolver.js needs to get the credentials: 'same-origin' as well.

(I do not have any experience with docker-based PHP development so I cannot create a PR)

shockey:
This was solved in https://github.com/swagger-api/swagger-js/pull/1083.

Thanks for the report, @glipman!

PR #1083, opened 2017-06-13
Add same-origin credentials policy to makeFetchJSON

dylancwood:
Adds [credentials: same-origin](https://github.com/github/fetch#sending-cookies) to resolved fetch options. This will instruct the browser to send cookies and authorization headers when fetching external definitions.

Existing issue: https://github.com/swagger-api/swagger-js/issues/1025
Symptom issue in swagger-ui: https://github.com/swagger-api/swagger-ui/issues/3220

Note on testing: Since these tests do not run in a browser, it would be difficult to assert that cookies are being set properly in a test. We could however create a spy that ensures that `fetch` is getting the `credentials: same-origin` option passed to it. 

@shockey

### DIFF
--- a/src/resolver.js
+++ b/src/resolver.js
@@ -9,7 +9,8 @@ export function makeFetchJSON(http) {
       loadSpec: true,
       headers: {
         Accept: 'application/json'
-      }
+      },
+      credentials: 'same-origin' // same-origin to send cookies and auth headers
     })
     .then((res) => {
       return res.body


dylancwood:
@shockey: What do you think of this? I'm sure you've got lots of other stuff going on, so I understand that you can't review everything right away. Just replying here to bring this back to the top of your inbox.

Thanks in advance.

shockey:
I get that this solves a problem with authentication, which is good.

My concern is that this lessens the amount of control that we have over the request... the browser will attach cookies to all requests to the origin, which could get weird when using Try-It-Out, if your API is on the origin.

To sidestep all this, we _could_ set `credentials: "same-origin"` only when resolving references.

dylancwood:
@shockey: I agree: we should only do it when resolving references. To be honest, that is what I assumed this did: it is in `resolver.js` after-all. Let me know if that is not the case.

shockey:
@dylancwood,



You're right, this is only for the resolver. My apologies!

In that case, this looks OK to me. Going to add a couple reviewers ( @webron @ponelat ).

dylancwood:
Thanks. Any genius ideas on testing this to avoid regressions in the future? This is a little tricky, since it is just an option passed to a third-part lib (`isomorphic-fetch`). That option then has a side-effect in the browser, which we cannot test directly in node (it's not worth it anyway).

shockey:
@dylancwood We'd need to start doing browser testing, which is something I've wanted for our whole suite for a while.

I'm pretty sure that our current tests would catch a leaking credentials attachment if they were run in a browser context, so new tests may not be required. 

Getting a node test suite running in the browser may not be trivial, though.

PR #1084, opened 2017-06-13
Add repository URL to package.json

dylancwood:
When viewing this package on npm.org, there is no link to the source code in this repo. This is especially confusing because the package name and repo name are different (swagger-js vs swagger-client).

We could also consider renaming the repository to `swagger-client`: git will automatically handle redirecting old urls that still reference `swagger-js`...

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "3.0.14",
   "description": "SwaggerJS - a collection of interfaces for OAI specs",
   "main": "dist/index.js",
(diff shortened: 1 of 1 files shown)


shockey:
@webron:

> We could also consider renaming the repository to swagger-client: git will automatically handle redirecting old urls that still reference swagger-js...

shockey:
Thanks @dylancwood! No doubt this will make the source repository easier to find.

PR #1093, opened 2017-06-23
Fix boolean false parameter

extempl:


webron:
Thanks for the PR. What behavior do you see without the fix?

extempl:
@webron actual behavior is missed parameter. `true` works properly here, so I added workaround for `false` only.

extempl:
@webron so without fix added test will fail with `http://swagger.io/v1/pets/findByStatus` != `http://swagger.io/v1/pets/findByStatus?status=false`

shockey:
Thanks for the PR, @extempl!

This will go live in the release we run later today.

PR #1098, opened 2017-06-30
Add userFetch interfaces for users to replace the fetch method & add really basic test

Robbilie:


Robbilie:
@shockey would be awesome if you could review this one since a friend and i are waiting for this feature :P

shockey:
@Robbilie, sorry for the delay, had to sync with a team member on this!

It's not really clear to me what this enables that can't be done via the `http` (or even, `fetch`) options in execute.

Further, attaching the fetch to the request when the request is built is interesting, but I don't see what benefit it provides over overriding http at execute time.

It's possible that I'm missing your use case here - can you share what you're doing that isn't possible without this PR?

Robbilie:
fetch cant be passed in toplevel as far as ive seen and replacing http requires one to implement:
- requestInterceptor
- responseInterceptor
- form formatting
- and so on…

which is not what i want, i just want to replace the "agent"

buunguyen:
Can you override `window.fetch` or `global.fetch` instead?

Robbilie:
why would anyone do that? this would force every other code running on the same website / in the same node app to use that other fetch method aswell which is not desired.

buunguyen:
Again, going back to what your use cases are, which you still haven't clarified.

Robbilie:
damnit i missed the response, prolly left the tab open.

but the response to your question lies within
https://github.com/swagger-api/swagger-js/pull/1098#issuecomment-313815863

currently you can only replace the whole http handler which means if i want to just replace the fetch method i have to implement all the logic again like said in the comment.

if i want to use another agent, not isomorphic-fetch, maybe a h2 client or some other lib that handles a huge load of requests better, uses a proxy or what not, this feature would be of great help.

buunguyen:
I'm okay with the change. But consider:
* Rename `userFetch` to something like `httpAgent`?
* Need to document this API

@shockey what you think?

Robbilie:
i would advise against httpAgent because httpAgents do have their own predefined set of internal api methods, if you want to go this way, sure, would just overcomplicate things here imo

documenting this api is something to be done for sure (writing down the existence of this additional param and describing the args and such)

Robbilie:
any news @shockey ?

shockey:
Approved by me!

I'll wait for formal approval via review from @buunguyen before merging - I want to make sure he's happy with the changes you made, and doesn't want any others.

PR #1100, opened 2017-06-30
Allow false query parameters

aaronfullerton:
Currently, when passing a boolean query parameter with a value of `false` to `http`, the following error surfaces:

```
 TypeError: value.map is not a function
      at formatValue (src/http.js:171:16)
      at src/http.js:187:36
      at Array.reduce (native)
      at encodeFormOrQuery (src/http.js:182:42)
      at Context.<anonymous> (test/http.js:218:14)
```

This PR resolves this error by adding a boolean type check in the `formatValue` function.


shockey:
Thanks for the PR, @aaronfullerton! Appreciate you taking the time to fix that issue.

Issue #1091, opened 2017-06-20
BasePath ignored when schema and host aren't set


JohnRowleySEL:
Currently if a spec doesn't have a schema and host set then the basepath is not used when executing an operation.

The source responsible for this can be found here. https://github.com/swagger-api/swagger-js/blob/master/src/execute.js#L212-L227

I have a fix on my own fork where I pretty much do the same thing that line 224 does but with the *computedPath*. I will create a pull request hopefully later this week.

ereOn:
Any news on this ?

shockey:
Fixed by #1166!

PR #1106, opened 2017-07-14
Default to computedPath if schema or host is not set

TheLocust3:
Fix for swagger-ui #1091
Default to using a relative basePath if host isn't set


shockey:
@TheLocust3, not sure what use case this is solving - can you explain, and also add some tests?

TheLocust3:
I actually found a better way of doing this (without needing a pr) but just to answer your question this would allow the user to specify a relative path to their Swagger spec file (during initialization), while not having a host name specified in their spec file.

PR #1122, opened 2017-08-19
return object if passed to header serializer

Robbilie:
if an object is passed to the serializeHeaders method, return it instead of an empty object, if nothing is passed to the method the default, an empty object is returned

### DIFF
--- a/src/http.js
+++ b/src/http.js
@@ -124,7 +124,7 @@ export function serializeHeaders(headers = {}) {
     return obj
   }
 
-  return obj
+  return headers
 }
 
 function isFile(obj) {


shockey:
Seems like a fair change... what's the use case here, though? I'm concerned that the only function of this change is to make it easier to use the interface incorrectly.

Robbilie:
if the response passed to the serializeRes method is a pure object (no Headers class used), currently the headers are completely ignored since the object already has the correct shape, just no forEach method

at least thats my use case :D

shockey:
Closing; I would prefer to accommodate non-Array header objects more directly (i.e. use a `for...of`, or some Lodash method that allows array and object iteration). Returning the input directly leaves the door open to browser implementations causing inconsistent behavior.

Sorry for the delay here, and thanks for your efforts @Robbilie!

Robbilie:
i still think that returning wrong data is worse then possibly browser dependant data but your call…

Issue #1066, opened 2017-05-24
run npm build fail


SimCordy:
Hi,
I've downloaded the swagger-js zip, and I have modified some file, so I need to rebuild it. The command _npm install_ works correctly, but when I run the command _npm run build_ it fail and return me some errors, this is the content of the npm-debug.log:

0 info it worked if it ends with ok
1 verbose cli [ 'C:\\Program Files\\nodejs\\node.exe',
1 verbose cli   'C:\\Program Files\\nodejs\\node_modules\\npm\\bin\\npm-cli.js',
1 verbose cli   'run',
1 verbose cli   'build' ]
2 info using npm@3.10.10
3 info using node@v6.10.3
4 verbose run-script [ 'prebuild', 'build', 'postbuild' ]
5 info lifecycle swagger-client@3.0.11~prebuild: swagger-client@3.0.11
6 silly lifecycle swagger-client@3.0.11~prebuild: no script for prebuild, continuing
7 info lifecycle swagger-client@3.0.11~build: swagger-client@3.0.11
8 verbose lifecycle swagger-client@3.0.11~build: unsafe-perm in lifecycle true
9 verbose lifecycle swagger-client@3.0.11~build: PATH: C:\Program Files\nodejs\node_modules\npm\bin\node-gyp-bin;C:\Users\Sibo\Desktop\swagger-ui\swagger-js-master\swagger-js-master\node_modules\.bin;C:\Python27;C:\PCI Geomatics\Geomatica 2017\exe;C:\ProgramData\Oracle\Java\javapath;C:\Windows\system32;C:\Windows;C:\Windows\System32\Wbem;C:\Windows\System32\WindowsPowerShell\v1.0\;C:\Program Files\Intel\WiFi\bin\;C:\Program Files\Common Files\Intel\WirelessCommon\;C:\Program Files\MATLAB\R2016b\bin;C:\Program Files (x86)\Brackets\command;C:\Program Files\nodejs\;C:\Users\Sibo\AppData\Local\Microsoft\WindowsApps;C:\Program Files\Intel\WiFi\bin\;C:\Program Files\Common Files\Intel\WirelessCommon\;C:\Users\Sibo\AppData\Roaming\npm
10 verbose lifecycle swagger-client@3.0.11~build: CWD: C:\Users\Sibo\Desktop\swagger-ui\swagger-js-master\swagger-js-master
11 silly lifecycle swagger-client@3.0.11~build: Args: [ '/d /s /c',
[...]

PR #1126, opened 2017-09-07
Use cross-env to make build scripts Windows-compatible

hkosova:
Currently, `npm run build` fails on Windows - it chokes on `NODE_ENV=production`, because Windows requires the `set` keyword to set environment variables: `set NODE_ENV=production`. See also: https://github.com/swagger-api/swagger-js/issues/1066#issuecomment-304204774.

This PR adds [cross-env](https://www.npmjs.com/package/cross-env) to make environment variable assignment work on both Windows and *nix.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,11 +16,11 @@
     "deps_check_dir": ".deps_check"
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


shockey:
Thanks @hkosova! 😄

PR #1130, opened 2017-09-10
Fixes #1129 - allOf fix for $$ref values

### LINKED ISSUE #1129: `allOf` pulls in incorrect `$ref` value
Say we have two definitions: `Pet` and `Cat`. `Cat` extends `Pet` using the `allOf` function.

After `swagger-js` parses the spec, the resulting definition for `Cat` contains a `$ref` property with the value of `PATH/Pet`, instead of the expected value of `PATH/Cat`.

This results in `swagger-ui` displaying the wrong name and tooltips for models that were built using `allOf`, as seen in the screenshots below:

_This screenshot shows the name of the model, and its `$ref` value. Note that the Cat model has an incorrect `$ref` value._

screen shot 2017-09-09 at 12 57 39 pm

_This screenshot shows the incorrect value of the tooltip when hovering the Cat model name._

screen shot 2017-09-09 at 12 57 08 pm

_This screenshot shows the incorrect name of the model and incorrect value of the tooltip for the `Cat` model, which is inside a `oneOf` option of the `Animal` model._

screen shot 2017-09-09 at 12 57 23 pm

The issue is happening with OAS 2.0 and OAS 3.0 specs.

```
openapi: 3.0.0
info:
  version: 0.0.0
  title: test
paths: {}
components:
  schemas:
    Pet:
      type: object
      properties:
        name:
          type: string
      required:
        - name
    Cat:
      allOf:
        - $ref: '#/components/schemas/Pet'
        - type: object
          properties:
            meowVolume:
              type: string
              enum:
              - loud
              - quiet
    Animal:
      type: object
      properties:
        pet:
          $ref: '#/components/schemas/Pet'
        cat:
          $ref: '#/components/schemas/Cat'
```

```
swagger: "2.0"
info:
  version: 0.0.0
  title: test
paths: {}
definitions:
  Pet:
    type: object
    properties:
      name:
        type: string
    required:
      - name
  Cat:
    allOf:
      - $ref: '#/definitions/Pet'
      - type: object
        properties:
          meowVolume:
            type: string
            enum:
            - loud
            - quiet
  Animal:
    type: object
    properties:
      pet:
        $ref: '#/definitions/Pet'
      cat:
        $ref: '#/definitions/Cat'
```

owenconti:
Fixes #1129 

Should also fix the following swagger-ui issues:

https://github.com/swagger-api/swagger-ui/issues/2946
https://github.com/swagger-api/swagger-ui/issues/3490
https://github.com/swagger-api/swagger-ui/issues/3342
https://github.com/swagger-api/swagger-ui/issues/3127
https://github.com/swagger-api/swagger-ui/issues/3093

and possibly others...

Added a patch at the end of the `allOf` plugin to make sure the original values for the definition are merged back into the result of the `allOf`. This includes the `$$ref` value. For definitions that did not originally include `$$ref` values, we make sure the result of `allOf` does not have a `$$ref` value.

I wrote a test for the new functionality, but during test runs (any test, not just mine), the `$$ref` value that is set during the `refs` plugin, is not available on the `patch` object inside the `allOf` plugin. 

Once we get that figured out, then we should be able to remove the `skip` from my new test.

Here's a screenshot showing the fix in action in the UI project:

Before:

screen shot 2017-09-09 at 1 40 55 pm

After:

screen shot 2017-09-10 at 9 49 11 am

### DIFF
--- a/src/specmap/lib/all-of.js
+++ b/src/specmap/lib/all-of.js
@@ -17,7 +17,7 @@ export default {
     const parent = fullPath.slice(0, -1)
     let alreadyAddError = false
 
-    return [specmap.replace(parent, {})].concat(val.map((toMerge, index) => {
+    const allOfPatches = [specmap.replace(parent, {})].concat(val.map((toMerge, index) => {
       if (!specmap.isObject(toMerge)) {
         if (alreadyAddError) {
           return null
@@ -31,5 +31,17 @@ export default {
 
       return specmap.mergeDeep(parent, toMerge)
     }))
+
+    // Find the $$ref value from the patch's copy of the spec
+    // and add it as the last patch for the `allOf` resolution
+    let specObj = patch.value
+    parent.forEach((part) => {
+      specObj = specObj[part]
+    })
+    allOfPatches.push(specmap.mergeDeep(parent, {
+      $$ref: specObj.$$ref
+    }))
+
+    return allOfPatches
   }
 }
--- a/src/specmap/lib/all-of.js
+++ b/src/specmap/lib/all-of.js
@@ -17,6 +17,15 @@ export default {
     const parent = fullPath.slice(0, -1)
     let alreadyAddError = false
 
+    // Find the original definition from the `patch.value` object
+    // Remove the `allOf` property so it doesn't get added to the result of the `allOf` plugin
+    let originalDefinitionObj = patch.value
+    parent.forEach((part) => {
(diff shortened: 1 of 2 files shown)


webron:
@shockey this solves oh-so-many issues, would love to see it merged

dballance:
Great fix - just saw this while following up on one of the linked issues!

shockey:
Okay, _now_ it's ready 😄 

Thanks @owenconti!

PR #1131, opened 2017-09-11
Fix empty oAuth2 Bearer token added to req header

c4milo:
When security definitions are present in the spec and an API call is made without providing an `access_token` value, the resulting auth2 authorization header is invalid: `authorization: Bearer `. 

This PR makes sure an oauth2 authorization header is added only if `access_token` has a value.


shockey:
Thanks, @c4milo!

c4milo:
You got it! Thanks for reviewing and merging! 👍

PR #1135, opened 2017-09-17
Fixes #1134 - Merging nested arrays

### LINKED ISSUE #1134: swagger-ui issue #3674 - Inheritance using 'allOf' seems broken
"Inheritance using 'allOf' seems broken" from swagger-ui repo: https://github.com/swagger-api/swagger-ui/issues/3674

owenconti:
Fixes #1134 

Fixes https://github.com/swagger-api/swagger-ui/issues/3674

Original implementation of merging arrays did not take into account nested arrays. I replaced the `deep-extend` package with lodash's `mergeWith`. 

Updated the existing test for merging arrays to test with `enum` options. Added new test for the new `mergeDeep` function.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,6 @@
   "dependencies": {
     "babel-runtime": "^6.23.0",
     "btoa": "1.1.2",
-    "deep-extend": "^0.4.1",
     "fast-json-patch": "~1.1.8",
     "isomorphic-fetch": "2.2.1",
     "isomorphic-form-data": "0.0.1",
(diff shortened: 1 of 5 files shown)


shockey:
From https://github.com/swagger-api/swagger-ui/issues/3674 (now closed):

> enums can never be merged though - on the contrary, only the values that are available in both of them should remain.

@owenconti, it sounds like this should be closed, since this concats `enum`s. Can you confirm?

owenconti:
@shockey Yes I think so.

Issue #1136, opened 2017-09-19
POST values in tags interface aren't sent


fehguy:
When using the tags interface like [this](https://github.com/swagger-api/swagger-js#tags-interface), content in POST values is _not_ sent to the server.  Sample PR with test on the way...

fehguy:
@shockey ping on this?

shockey:
@fehguy 👋 

I changed the test you wrote in #1137 to the following:
```js
client.apis.Testing.tryMe({
  theBody: {
    name: 'tony'
  }
})
```

and the test passed. This is how I imagine the interface should work, but I may be missing something.

Should we be matching parameter values to body parameter schema properties (i.e. `name => body.name`)? Is this something we were doing in 2.x?

fehguy:
oh! I think that what you put is the right behavior.  I'll retest and report back...

shockey:
@fehguy bump 😄

CameronAckermanSEL:
Hey guys. I'm seeing something similar.

Given this spec:

```
"/projects/{projectKey}/containers/{containerKey}/stuffs":{  
   "parameters":[  
      {  
         "in":"path",
         "name":"projectKey",
         "required":true,
         "type":"string"
      },
      {  
         "in":"path",
         "name":"containerKey",
         "required":true,
         "type":"string"
      }
   ],
   "post":{  
      "description":"Inserts a new `Stuff` for the `Container`.",
      "operationId":"insertStuff",
      "parameters":[  
         {  
            "in":"body",
            "name":"stuff",
            "required":true,
            "schema":{  
               "$ref":"#/definitions/StuffParameters"
            }
         }
      ],
      "responses":{  
         "201":{  
            "description":"successfully added stuff.",
            "schema":{  
               "$ref":"#/definitions/Stuff"
            }
         },
         "404":{  
            "$ref":"#/responses/NotFound"
         },
         "500":{  
            "$ref":"#/responses/Error"
         }
      },
      "tags":[  
         "stuff"
      ]
   }
}
```

And this method of calling the client:

```

```

I have two cases where swagger-client will fail to validate the request parameters and issue the request with no body.

1. If I pass in an explicitly defined object for the stuff argument


2. If I pass in an anonymous object that has extra parameters



I understand that the validation is useful. But my expectations are:

1. If there's a validation error, a descriptive exception is thrown and the request is NOT sent.
2. OR The client ignores the extra stuff.
[...]

PR #1137, opened 2017-09-19
added tests for #1136

fehguy:


fehguy:
The build shouldn't pass as the post test validation [here](https://github.com/swagger-api/swagger-js/pull/1137/files#diff-7e3bd8e7059a21bfd255649cbfc84b33R181) will fail.

PR #1140, opened 2017-09-28
OAS3: Style/Explode parameter serialization

shockey:
This PR:
- adds support for `style` and `explode` 
- constructs OAS3 requests based on default style/explode modes for each parameter type
- adds tests for every style/explode/value type combination found at https://swagger.io/docs/specification/serialization/
- adds `allowReserved` functionality, along with tests

Limitations:
- `deepObject` does not handle nested objects. The specification and swagger.io documentation does not provide an example for serializing deep objects. Flat objects will be serialized into the `deepObject` style just fine.


shockey:
@webron, mostly looking for you to either sign off on this PR with the limitation mentioned above, or give guidance on how to go about implementing nested object support 😄 thanks!

webron:
Why do you hate me, @shockey.

webron:
Okay. If you followed the documentation in https://swagger.io/docs/specification/serialization, and made sure the default values mentioned there are taken into account when no value is explicitly provided, it should be fine.

As for `deepObject` and nested objects - that was explicitly left out of the spec, and it's ok to just Not Support It™.

PR #1148, opened 2017-10-09
Tighten `fetch-mock` version constraint

shockey:
`fetch-mock@5.13.0` is blowing up our tests (25 failing, see https://travis-ci.org/swagger-api/swagger-js/jobs/285076321).

This PR tightens the version constraint on it, in order to avoid the bad version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "eslint-config-airbnb-base": "^11.1.1",
     "eslint-plugin-import": "^2.2.0",
     "expect": "^1.20.2",
-    "fetch-mock": "^5.9.3",
+    "fetch-mock": "~5.12.0",
     "glob": "^7.1.1",
     "json-loader": "^0.5.4",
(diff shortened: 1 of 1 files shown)


shockey:
CC @wheresrhys - I haven't gotten to the bottom of this problem, so I haven't opened a `fetch-mock` issue (might be on our end), but thought you'd like to be aware of this.

wheresrhys:
Thanks - I'll checkout and run locally too and let you know what I find

wheresrhys:
Yep it was me. I'll add a regression test to make sure this doesn't happen again https://github.com/wheresrhys/fetch-mock/pull/223/files#diff-2917a039dc612e08d6fae9869d0e5c42L181

wheresrhys:
fixed in 5.13.1

shockey:
Awesome! Thanks @wheresrhys 😄

PR #1149, opened 2017-10-09
.then(fetchMock.restore) -> .finally(fetchMock.restore)

shockey:
CC: #1148 

Previously, a failing test that uses `fetchMock` would fail to run `fetchMock.restore`, causing a cascading failure of tests below it.

This PR ensures that restore is always being called after a test is complete, regardless of outcome.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "swagger-client",
-  "version": "3.2.2",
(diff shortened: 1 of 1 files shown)


shockey:
Build fails on Node.js 4 only... I'll need to come back to this another time.

PR #1156, opened 2017-10-15
Fixes #1155

### LINKED ISSUE #1155: `bearer` token type is not uppercased
Original issue: https://github.com/swagger-api/swagger-ui/issues/3594

owenconti:
Fixes #1155 

Pass 'Bearer' if the token_type property equals 'bearer' or is falsy.

Fixes original issue found in UI: https://github.com/swagger-api/swagger-ui/issues/3594

### DIFF
--- a/src/execute/swagger2/build-request.js
+++ b/src/execute/swagger2/build-request.js
@@ -84,7 +84,7 @@ export function applySecurities({request, securities = {}, operation = {}, spec}
           }
         }
         else if (type === 'oauth2' && accessToken) {
-          tokenType = (!tokenType || tokenType === 'bearer') ? 'Bearer' : tokenType
+          tokenType = (!tokenType || tokenType.toLowerCase() === 'bearer') ? 'Bearer' : tokenType
           result.headers.authorization = `${tokenType} ${accessToken}`
         }
       }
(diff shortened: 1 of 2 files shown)


src/execute/swagger2/build-request.js
```diff
@@ -84,7 +84,8 @@ export function applySecurities({request, securities = {}, operation = {}, spec}
           }
         }
         else if (type === 'oauth2' && accessToken) {
-          result.headers.authorization = `${tokenType || 'Bearer'} ${accessToken}`
+          tokenType = (!tokenType || tokenType === 'bearer') ? 'Bearer' : tokenType
```
shockey: I'd prefer for the `token_type` detection for bearer to be case-insensitive, i.e. our `Bearer` override will kick in for token_type `bearer`, `BEARER`, `BeArEr`.

I know this is a minor point, and there may not be an OAuth server implementation that sends those values, but if we're not paying attention to token_type `bearer` vs `Bearer`, I'd like to go all the way with it, instead of adding a new magic token_type that triggers the correct header value.

owenconti: Updated

shockey:
Thanks @owenconti!

PR #1157, opened 2017-10-18
Don't try to access operation.operationId if operation isn't an object

shockey:
No report for this bug, but I ran into it today.


webron:
😱

Issue #1165, opened 2017-10-24
Optimize bundle size when included in another project


arty-name:
Hi there,

I am trying to use swagger-client in my React application which is built using webpack. I have figured that swagger-client increases the bundle size by a whopping 500 kB instead of expected 40. Here’s why.

The simplest thing to fix is the fixed version of lodash that you have in your package.json. If it were a range, my own version of lodash would be reused between my code and swagger-client, removing ~100 kB of code. Until that I have to work around the issue by configuring Webpack like this: 

```javascript
resolve: { alias: { lodash: path.resolve(__dirname, 'node_modules/lodash') }}
```

There are similar issues with core-js and regenerator-runtime, so similar workaround removes additional 100 kB: 
```javascript
resolve: { alias: { 
    'core-js': path.resolve(__dirname, 'node_modules/core-js'),
    'regenerator-runtime': path.resolve(__dirname, 'node_modules/regenerator-runtime'),
}}
```

Also I deliver the Swagger Spec as a JSON object embedded in the bundle, however swagger-client unconditionally brings in 40 kB of `js-yaml` which in turn carries whopping 250 kB of `esprima` with it. Neither of them is needed, so I can alias esprima to the `empty-module`, however I cannot do that to js-yaml because of these lines in your `http.js`: 

```javascript
// Optimistically try to convert all bodies
const obj = jsYaml.safeLoad(body)
```

I am pretty confident that yaml parser is not needed for json ;)

Would you like to work together on reducing the footprint of swagger-client in react bundles?

shockey:
Hi @arty-name!

> The simplest thing to fix is the fixed version of lodash that you have in your package.json

This is an oversight - a PR that loosens the version constraint to major version 4 would be appreciated 😄 

I am surprised that this pulls 100kB into your bundle though - we cherry-pick functions from lodash (i.e. `import reduce from "lodash/reduce"`), so the entire module shouldn't be in your bundle, but I digress...

> There are similar issues with core-js and regenerator-runtime, so similar workaround removes additional 100 kB

Not as sure what's going on here - is your project using a different major version of `babel-runtime` and friends?

> I deliver the Swagger Spec as a JSON object embedded in the bundle, however swagger-client unconditionally brings in 40 kB of js-yaml which in turn carries whopping 250 kB of esprima with it. Neither of them is needed

In accordance with the specifications we support, we support the loading of Swagger/OpenAPI documents represented in YAML or JSON:

[...]

PR #1175, opened 2017-11-02
only use YAML parser for YAML responses

arty-name:
### Description
Check the content-type of response and use appropriate parser: `js-yaml`, json, or no parser.

### Motivation and Context
As described in #1165 even if one doesn’t use YAML at all, providing the spec via JSON object and using JSON to talk to backend, the swagger-client still uses `js-yaml` to parse responses. The downside to that approach is that one cannot forcefully remove `js-yaml` from Webpack bundle. 

If someone however have been using invalid JSON in the responses, this change will break things. Apparently the JSON parser of `js-yaml` is quite relaxed and handles even invalid JSON like this `{"123": NULL}`. It is your call whether to treat this as a breaking change or not.

### How Has This Been Tested?
`npm test` has passed

### Types of changes

### Checklist:

### DIFF
--- a/src/http.js
+++ b/src/http.js
@@ -67,6 +67,13 @@ export default function http(url, request = {}) {
 // exported for testing
 export const shouldDownloadAsText = (contentType = '') => /json|xml|yaml|text/.test(contentType)
 
+function parseBody(body) {
+  if (/^\s*\{/.test(body)) {
+    return JSON.parse(body)
+  }
+  return jsYaml.safeLoad(body)
+}
+
 // Serialize the response, returns a promise with headers and the body part of the hash
 export function serializeRes(oriRes, url, {loadSpec = false} = {}) {
   const res = {
@@ -88,8 +95,7 @@ export function serializeRes(oriRes, url, {loadSpec = false} = {}) {
 
     if (useText) {
       try {
-        // Optimistically try to convert all bodies
-        const obj = jsYaml.safeLoad(body)
+        const obj = parseBody(body)
         res.body = obj
         res.obj = obj
       }


shockey:
@arty-name, my suggestion going forward here would be to do a naive content check, something like:

```
if(req.body.trim()[0] === "{") {
  // json 
} else { 
  // yaml
 }
```

---

Also, concerning your use case of wanting to jettison `js-yaml`: have you considered aliasing `js-yaml` to `JSON.parse`? Like this:

```js
// alias.js: `js-yaml` interface mock

export default {
  safeLoad(json) {
    return JSON.parse(json)
  }
}
```

```js
// webpack.config.js
resolve: { 
  alias: { 
    "js-yaml": path.resolve(__dirname, 'alias.js') 
  }
}
```

arty-name:
Thank you for adding tests, @shockey! I have updated my change as you suggest - almost. For performance reasons I have replaced trimming with regexp. How does it look now?

The tip with aliasing safeLoad to JSON.parse will technically work, but it looks to me too much like a tight coupling.

shockey:
Thanks @arty-name!

arty-name:
@shockey, there’s actually something wrong about the OpenAPI argument. 

Yes, the OpenAPI Document is always an object. However the specification allows the **responses** to be arrays, it even includes [examples](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#responseObject). 

The code in `serializeRes` is used for both Swagger specs and responses, so it’s perfectly fine for it to receive a JSON-encoded array, string, number, null, or object. Naive content check seems not too reliable for that case.

shockey:
@arty-name, you're correct.

Fortunately, this shouldn't create any issues for folks that aren't excluding `js-yaml` from their projects 😄 

I'd be open to a second swing on this, where we swap out the naive check for `contentType === 'application/json'`. 

This would require you (and others that want to ditch `js-yaml`) to always be sure to declare a content type, but would handle the issue you've pointed out.

arty-name:
I’m happy to hear that you don’t mind to check another pull request for this issue! I have created #1183 with this approach.

PR #1183, opened 2017-11-16
detect JSON by provided content-type instead of content check

arty-name:
### Description
Check the content-type of response and use JSON parser only for `application/json`

### Motivation and Context
As described in #1165 even if one doesn’t use YAML at all, providing the spec via JSON object and using JSON to talk to backend, the swagger-client still uses js-yaml to parse responses. The downside to that approach is that one cannot forcefully remove js-yaml from Webpack bundle.

If someone however have been using invalid JSON in the responses, this change will break things. Apparently the JSON parser of js-yaml is quite relaxed and handles even invalid JSON like this {"123": NULL}. It is your call whether to treat this as a breaking change or not.

### How Has This Been Tested?
`npm test` has passed

### Types of changes

### Checklist:

### DIFF
--- a/src/http.js
+++ b/src/http.js
@@ -67,8 +67,8 @@ export default function http(url, request = {}) {
 // exported for testing
 export const shouldDownloadAsText = (contentType = '') => /json|xml|yaml|text/.test(contentType)
 
-function parseBody(body) {
-  if (/^\s*\{/.test(body)) {
+function parseBody(body, contentType) {
+  if (contentType === 'application/json') {
     return JSON.parse(body)
   }
   return jsYaml.safeLoad(body)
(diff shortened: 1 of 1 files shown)


shockey:
> If someone however have been using invalid JSON in the responses, this change will break things. Apparently the JSON parser of js-yaml is quite relaxed and handles even invalid JSON like this {"123": NULL}. It is your call whether to treat this as a breaking change or not.

I'm inclined to say no: if you're working with invalid data you shouldn't expect your use case to be officially supported.

shockey:
> Apparently the JSON parser of js-yaml is quite relaxed and handles even invalid JSON like this {"123": NULL}. It is your call whether to treat this as a breaking change or not.

I believe that the js-yaml parser technically only parses YAML - but with YAML being a superset of JSON, we can also use it to parse JSON. Because of that, though, you can also mix YAML and JSON notation and get away with calling it JSON.

arty-name:
Sweet! Thank you @shockey!

PR #1190, opened 2017-11-29
Add lookaside to `traverse` - Limits DFS revisit

### LINKED ISSUE #1150: Limiting DFS in `createKeyBasedPlugin`
I'm struggling with some absolutely massive OAS documents in swagger-js (~3.5MB on disk). They contain many JSON pointers ($ref plugin produces ~28000 patches). They're also _potentially_ highly graphical - they're OData based and the OData `$expand` query param can be used to request related entities. Typically this is depth limited by the API, but as far as the OData metadata / OAS 2.0 metadata they're simply references that could repeat. This leads to lots of revisiting in `createKeyBasedPlugin`'s `traverse` method.

To the best of my knowledge, as patches are applied, `traverse` will revisit the same referenced path over and over, even if they are the same underlying type - here `#/responses/400` and `#/responses/401` both reference `#/definitions/OData.Error`. The underlying reference is the same, but we traverse the entire tree for both patches, because we don't know that we've already visited the  `#/definitions/OData.Error` node in search for additional modifications. (Note: I think is is what is occurring, but I'm not sure here) I believe the second traversal is redundant, as are any future traversals of `#/definitions/OData.Error`.

image

The paths contain the results of logging the path argument to traverse
```js
        function* traverse(obj, path, patch, depth) {
          console.log(path.join('.'))
```

Since I have a limited knowledge of the design / intent behind the implementation - I wanted to ask - **Is there any reason NOT keep a reference to this node and simply skip traversal once we've already traversed a particular JSON pointer's referenced node?** This would be on a 'per-plugin' basis - one lookaside for each plugin. That means once a pointer is visited once in the `$ref` plugin, it wouldn't be visited again as patches are evaluated.

I've implemented a lookaside that utilizes the `allowMetaPatches` option - and the results in speed are astounding. Previously, spec resolution would take a long time - in the multi-hour range. I don't have an exact number because I never got it to complete. After adding the lookaside it resolves the specs within 10-20 seconds. I'm unsure of the completeness - as the resulting JSON is massive and I cannot get a swagger-ui utilizing this patched version of swagger-js to load - but wanted to ask about the intent here before I write some tests....

Result of traversal with look aside. 
image

#1117

dballance:
### Description
See #1150. Adds a 'per-patch' lookaside to the `traverse` method within `createKeyBasedPlugin`.

### Motivation and Context
Previously, a `$ref` would be traversed multiple times during spec resolution. This change allows `$ref`s to be resolved once and prevents future unnecessary revisits.

Fixes #1150 - but still some questions about handling `allOf` resolution that may be handled in another PR.

### How Has This Been Tested?
All tests appear to pass - unsure if additional tests are required or need to be added for this change.

Existing code. Small spec (~5k lines formatted JSON). Average Resolution of ~1100ms.
image

With change. Average Resolution of ~250ms.
image

Existing code. Large spec (~141k lines formatted JSON). **NOTE: `allOf` plugin disabled for the below tests - only running `$ref` plugin.**
**????** I couldn't actually get this to complete in any reasonable time. 😄 

With change. Average Resolution of ~5200ms.
image

### Types of changes

### Checklist:

### DIFF
--- a/src/specmap/index.js
+++ b/src/specmap/index.js
@@ -117,8 +117,8 @@ class SpecMap {
                   if (specmap.allowMetaPatches && objRef) {
                     traversedRefs[objRef] = true
                   }
-                yield* traverse(val, updatedPath, patch)
-              }
+                  yield* traverse(val, updatedPath, patch)
+                }
               }
 
               if (!isRootProperties && key === pluginObj.key) {
--- a/src/specmap/index.js
+++ b/src/specmap/index.js
@@ -83,6 +83,8 @@ class SpecMap {
     // We might consider making this (traversing & application) configurable later.
     function createKeyBasedPlugin(pluginObj) {
       return function* (patches, specmap) {
+        const traversedRefs = {}
+
         for (const patch of patches.filter(lib.isAdditiveMutation)) {
           yield* traverse(patch.value, patch.path, patch)
         }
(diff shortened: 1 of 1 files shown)


shockey:
Thanks for this, @dballance! Very excited to ship this, I expect significant gains across the board for large definitions :smile:

PR #1195, opened 2017-12-04
Add failing test case for reserved keywords

ponelat:
This adds a failing test-case to highlight the issue around reserved keywords.
Definition taken from: https://github.com/swagger-api/swagger-editor/issues/1343

The keywords "$ref" and "allOf" will result in an error, if used in places reserved for freely-named fields, such as model, parameter, sub-schema names, etc.


shockey:
I addressed this in https://github.com/swagger-api/swagger-js/pull/1214, didn't realize a test PR was open.

I'll merge it into here and see what we get.

PR #1197, opened 2017-12-04
Fix #1180 - spammy warnings about ambiguous parameters

### LINKED ISSUE #1180: Incorrect parameter ambiguity warnings
In our swagger specification we have the following:

```
"/projects/{projectKey}/containers/{containerKey}/stuffs":{  
   "get":{"description":"Retrieves all stuff in specified container",
      "operationId":"getStuff",
      "responses":{  
         "200":{  
            "description":"The `stuff` for the containerwith the given `containerKey`",
            "schema":{  
               "items":{  
                  "$ref":"#/definitions/Stuff"
               },
               "type":"array"
            }
         },
         "404":{  
            "$ref":"#/responses/NotFound"
         },
         "500":{  
            "$ref":"#/responses/Error"
         }
      },
      "tags":[  
         "stuff"
      ]
   },
   "parameters":[  
      {  
         "in":"path",
         "name":"projectKey",
         "required":true,
         "type":"string"
      },
      {  
         "in":"path",
         "name":"containerKey",
         "required":true,
         "type":"string"
      }
   ],
   "post":{  
      "description":"Inserts a new `Stuff` for the `Container`.",
      "operationId":"insertStuff",
      "parameters":[  
         {  
            "in":"body",
            "name":"stuff",
            "required":true,
            "schema":{  
               "$ref":"#/definitions/StuffParameters"
            }
         }
      ],
      "responses":{  
         "201":{  
            "description":"successfully added stuff.",
            "schema":{  
               "$ref":"#/definitions/Stuff"
            }
         },
         "404":{  
            "$ref":"#/responses/NotFound"
         },
         "500":{  
            "$ref":"#/responses/Error"
         }
      },
      "tags":[  
         "stuff"
      ]
   }
}
```

This yields a large amount of spammy warnings about the parameter being ambiguous
```
console.js?32ad:32 Parameter 'projectKey' is ambiguous because the defined spec has more than one parameter with the name: 'projectKey' and the passed-in parameter values did not define an 'in' value.
console.js?32ad:32 Parameter 'containerKey' is ambiguous because the defined spec has more than one parameter with the name: 'containerKey' and the passed-in parameter values did not define an 'in' value.
console.js?32ad:32 Parameter 'projectKey' is ambiguous because the defined spec has more than one parameter with the name: 'projectKey' and the passed-in parameter values did not define an 'in' value.
console.js?32ad:32 Parameter 'containerKey' is ambiguous because the defined spec has more than one parameter with the name: 'containerKey' and the passed-in parameter values did not define an 'in' value.
```

Demonstrably per the JSON specification above, the path does not have duplicate parameters,

CameronAckermanSEL:
There was subtle defect involving a combination of `normalizeSwagger` and `buildRequest`, where `normalizeSwagger` would generate a spec where operations would inherit the path parameters, and then `buildRequest` would interpret these inherited parameters as redundant and ambiguous because it was somehow not aware of the inheritance in `normalizeSwagger`. This led to us experiencing about 4 warning messages per request.

### Description

When `buildRequest` creates a complete list of parameters from the operation and from the path, we examine the combined array and eliminate true duplicates (that is parameters that have the same name, and 'in').

### Motivation and Context
This fix eliminates a huge amount of spam in our application. We would get 2x warnings for each parameter in the path of an operation.

This fixes #1180

### How Has This Been Tested?
There are two new unit tests. One unit test exercises a situation where the ambiguous parameter warnings are issued, and then one unit test exercises a situation where the ambiguous parameter warnings would have been issued prior to the fix, but no longer emits warnings.

image

### Types of changes

### Checklist:


src/execute/index.js
```diff
@@ -151,9 +169,10 @@ export function buildRequest(options) {
     req.headers.accept = responseContentType
   }
 
-  const combinedParameters = []
+  const combinedParameters = deduplicateParameters([]
     .concat(arrayOrEmpty(operation.parameters)) // operation parameters
     .concat(arrayOrEmpty(path.parameters)) // path parameters
+  )
```
CameronAckermanSEL: As mentioned in #1180, this code would grab the parameters from the operation and the parameters from the path, and merge them into a single array. Because this code is called after `normalizeSwagger`, the operation.parameters array would already have the parameters from the path. This would create duplicate parameters. The duplicate parameters aren't necessarily a problem, but the logic surrounding the ambiguous parameter warning doesn't handle this situation properly.

test/execute/main.js
```diff
@@ -1626,6 +1627,95 @@ describe('execute', () => {
...
+                      type: 'number'
+                    }
+                  ],
+                }
+              }
+            }
+          }
+        }
+
+        const resultSpec = normalizeSwagger(spec)
```
CameronAckermanSEL: the existing unit tests were not catching this condition because none of the tests normalizeSwagger before calling buildRequest.

CameronAckermanSEL: Now we have such a test, and we have a regression test verifying that the warning is not thrown in this situation where the parameters are inherited by the operation.

test/execute/main.js
```diff
@@ -1626,6 +1627,95 @@ describe('execute', () => {
...
+        expect(req).toEqual({
+          url: 'http://swagger.io/v1/pet/123?test=567',
+          headers: {},
+          credentials: 'same-origin',
+          method: 'GET'
+        })
+        expect(warnSpy.calls.length).toEqual(0)
+      })
+
+      it('should warn for ambiguous parameters in normalized swagger specifications', function () {
```
CameronAckermanSEL: this test verifies that when there are in fact ambiguous parameters (that is two parameters with the same name that have different 'in's) the warning is thrown.

src/execute/index.js
CameronAckermanSEL: There's a small performance cost to this, but I felt it was much less risky than trying to make normalizeSwagger not pass the path parameters down to the operation, and it was better than not concatenating the arrays here.

CameronAckermanSEL:
@shockey yo! I'm not sure what's up with the Travis build. It says abuse detected and it's not running the build. Could it be because of the spy on console.warn? 

Want me to send them a support ticket or do you have a way of dealing with it? 

Thanks! :D

shockey:
@CameronAckermanSEL, hmm this is strange. Can you link me to where you're seeing `abuse detected`?

CameronAckermanSEL:
@shockey sure thing! At the top of the list here: https://travis-ci.org/swagger-api/swagger-js/requests

I initially thought the build was hanging because I hadn't merged with master so that's why you'll see the build request for this PR twice.

shockey:
Well that's certainly not right... I'll email them and circle back here when I hear from them.

CameronAckermanSEL:
Thanks very much! Let me know if I can help in any way

shockey:
@CameronAckermanSEL, here's their response: [Travis CI support.pdf](https://shipusercontent.com/07ce475eacbf6b4eb597555f3628b606/Travis%20CI%20support.pdf)

Sounds like you need to provision your fork with Travis CI.

CameronAckermanSEL:
@shockey  thanks for the help!! It looks like everything is in order

shockey:
thanks @CameronAckermanSEL!

PR #1201, opened 2017-12-09
$ref resolution issues

shockey:
### Description

This PR replaces the existing traversal cache with a result cache. The traversal cache was prone to issues that were causing display bugs in UI/Editor.

### Motivation and Context

Fixes https://github.com/swagger-api/swagger-editor/issues/1585. Fixes https://github.com/swagger-api/swagger-ui/issues/4000.

Follow-up to #1190. It's not as performant, but it is more accurate.

### How Has This Been Tested?

- I added a unit test that was previously failing.
- I linked the changes to Swagger-UI and observed that the display issues went away.

### Types of changes

### Checklist:

### DIFF
--- a/src/specmap/index.js
+++ b/src/specmap/index.js
@@ -83,7 +83,7 @@ class SpecMap {
     // We might consider making this (traversing & application) configurable later.
     function createKeyBasedPlugin(pluginObj) {
       return function* (patches, specmap) {
-        const traversedRefs = {}
+        const resultCache = {}
 
         for (const patch of patches.filter(lib.isAdditiveMutation)) {
           yield* traverse(patch.value, patch.path, patch)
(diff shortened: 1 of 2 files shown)


shockey:
cc @dballance: pretty sure this is a step back for performance (i'm caching the plugin result, not the traversal), so feel free to build on top of this.

dballance:
I still need to make changes to `allof` to test my large spec, but the smaller one I tested in my PR showed an increase to ~580ms.  Still close to half the original resolution time, but with greater accuracy - works for me. 👍 

Will grab some results on the massive spec sometime soon. Have `allOf` work scheduled in January to try and resolve our issues....

shockey:
Well, this is embarrassing™

Turns out this PR wasn't passing all tests - I left a `describe.only` in the tests that kept.... all the tests, from running.

I've reverted this PR, so we're back to no optimizations for now - I thought it better to ship unoptimized than to hold up the release.

I'll circle back here on Monday to get this in full working order 😄

PR #1209, opened 2017-12-19
Re-implement lookaside for DFS during resolution

dballance:
### Description
cc @shockey - Lets try this again! 😆 

In #1190 - the lookaside actually prevented traversal of all properties of a reference - because we marked it as traversed and checked for this traversal before fully iterating the properties in the patch. This means that DFS was never fully complete for a `$ref` that we marked as traversed. This moves the check for previous traversal out of the loop and allows a complete traversal but prevents future traversals. Should be accurate as far as I can tell. 

I believe this was the root of "required" props not being parsed in swagger-api/swagger-ui#4000

### Motivation and Context
Followup to #1201 and #1190. Again reduces resolution times to the #1190 range, but with greater accuracy. 

### How Has This Been Tested?
All tests pass. Tested with samples from swagger-api/swagger-editor#1585 & swagger-api/swagger-ui#4000 - doesn't appear to break those specs.

@shockey Unsure if there needs to be additional tests - so please grab and test as well! Might need to merge tests from #1201?

### Types of changes

### Checklist:

### DIFF
--- a/src/specmap/index.js
+++ b/src/specmap/index.js
@@ -83,6 +83,8 @@ class SpecMap {
     // We might consider making this (traversing & application) configurable later.
     function createKeyBasedPlugin(pluginObj) {
       return function* (patches, specmap) {
+        const refCache = {}
+
         for (const patch of patches.filter(lib.isAdditiveMutation)) {
           yield* traverse(patch.value, patch.path, patch)
         }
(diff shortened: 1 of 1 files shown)


shockey:
Reviewing now - looks like my reversions did drop the tests from #1201, going to resolve that.

shockey:
Thanks @dballance! 😄

PR #1213, opened 2017-12-26
Fixed broken link to swagger-client.min.js

FirefoxMetzger:
The link was pointing to master, where there is no `.min.js`. This PR fixes it, pointing the link to the `.min.js` in this branch (2.x).


shockey:
Thanks for this, @FirefoxMetzger!

PR #1218, opened 2018-01-07
Fixing support for matrix parameters when not present in the path definition

ilgrosso:
### Description

While already part of the OpanApi Spec 3.0, Matrix parameters are currently ignored by swagger-js and, by consequence, by swagger-ui.

### Types of changes

### Checklist:

### DIFF
--- a/src/execute/oas3/parameter-builders.js
+++ b/src/execute/oas3/parameter-builders.js
@@ -10,14 +10,19 @@ export default {
 function path({req, value, parameter}) {
   const {name, style, explode} = parameter
   const styledValue = stylize({
-    key: parameter.name,
+    key: name,
     value,
     style: style || 'simple',
-    explode: explode || false,
-    escape: false,
+    explode: typeof explode === 'undefined' ? false : explode,
+    escape: style === 'matrix'
   })
 
-  req.url = req.url.replace(`{${name}}`, styledValue)
+  if (style === 'matrix' && req.url.indexOf(`{${name}}`) === -1) {
+    req.url = req.url.concat(styledValue)
+  }
+  else {
+    req.url = req.url.replace(`{${name}}`, styledValue)
+  }
 }
 
 function query({req, value, parameter}) {
--- a/test/oas3/execute/style-explode/path.js
+++ b/test/oas3/execute/style-explode/path.js
@@ -824,5 +824,43 @@ describe('OAS 3.0 - buildRequest w/ `style` & `explode` - path parameters', func
         headers: {},
       })
     })
+
+    it('should build a path parameter in a matrix/no-explode format when not present in the path definition', function () {
+      // Given
+      const spec = {
+        openapi: '3.0.0',
+        paths: {
+          '/path/': {
+            get: {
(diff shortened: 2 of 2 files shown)


ilgrosso:
Thanks for your indications @shockey and clarifications about the matrix parameter support in OAS3 - the current work was produced under bad supposition, it seems.
I'll come soon with an updated PR, complying with what you say above.

shockey:
@ilgrosso no worries! there’s a lot of nuance in the code, it’s not hard to miss 😄

feel free to ping me here if you need any help.

ilgrosso:
Please take a look @shockey - now `npm run lint` also runs fine.

ilgrosso:
Ok, *now* it should be working, please @shockey have a look...

ilgrosso:
@shockey as you can see from Travis CI, the checks on Node.js 6.9 went green but the ones on Node.js 4.7 failed because the distribution could not be downloaded

shockey:
@ilgrosso thanks for the heads up, I just requeued the 4.7 one; looks like nodes.org was having some trouble yesterday.

ilgrosso:
thx @shockey - now all is green it seems :-)

shockey:
Thanks @ilgrosso! One last request - can you add some tests? See my review above for some pointers 😄

ilgrosso:
@shockey test added, sorry for overlooking this.
Thanks for your review.

ilgrosso:
Hi @shockey, what's missing to accept this PR?

shockey:
Hi @ilgrosso!

The test cases you added illuminated what you meant in your PR title: It's my fault for not digging into that earlier in the process.

From the 3.0.0 Definitions section:

> Path templating refers to the usage of curly braces ({}) to mark a section of a URL path as replaceable using path parameters.

and from the Parameter Object section:

> path - Used together with Path Templating, where the parameter value is actually part of the operation's URL. This does not include the host or base path of the API. For example, in /items/{itemId}, the path parameter is itemId.

I reached out to an OpenAPI TSC (the committee that writes the OpenAPI Specification) member, and they confirmed for me that path parameters must be referenced in the path key name in order to be used - path parameter values should not be concatenated to request URLs if the parameter is not referenced in the path itself. As a result, the changes you want to make would go against the specification.

Again, my apologies for not addressing this earlier in the PR process! All of that being said, I'd be happy to accept any improvements to how we generate matrix path parameters when the parameter _is_ being used in the path name.

ilgrosso:
I see: then it seems there is a bigger problem as, being constrained to be just a "style" of path parameters, matrix parameters do not have yet an adequate definition in OpenAPI v3.

As far as I can tell, in fact, matrix parameters are used by JAX-RS as an optional part of the URL (similarly to query parameters), while path parameters are a mandatory part of the URL.

When I saw https://github.com/OAI/OpenAPI-Specification/issues/69 closed, I though that finally the vexata quaestio was solved, and that people writing REST in Java could finally use freely matrix parameters; unfortunately, it seems we are not yet to that point.

Thanks for clarification, @shockey and sorry for the rant. I am closing this PR now.

PR #1225, opened 2018-01-15
Remove the unnecessary loop of all operations when new one added

### LINKED ISSUE #1221: Incorrect setting __originalOperationId when operationId was not defined
When processing the following swagger spec:

```
        const spec = {spec: {
          paths: {
            '/foo': {
              get: {
              },
              post: {
                operationId: 'myId2'
              }
            },
          }
        }}
```

The `normalizeSwagger` method in `src/helpers.js` would set `__originalOperationId` of both operations to 'myId2'. This causes the Swagger Editor to incorrectly report validation error of "Operations must have unique operationIds." (see https://github.com/swagger-api/swagger-editor/issues/1632)

Problem seems to be caused by the method looping through all operations when one being added to the map at [line 149](https://github.com/swagger-api/swagger-js/blob/master/src/helpers.js#L149).

Happy to submit a pull request.

jemerald:
### Description

The previous code was looping through all known operations and update their __originalOperationId and operationId, after each one operation is added to the known list.

This is not only inefficient, but introduces a bug at the following line

```
obj.__originalOperationId = obj.__originalOperationId || operation.operationId
```

Where `operation` is the one that just got added and `obj` is the current operation in the loop.

This means for the following spec:

```
        const spec = {spec: {
          paths: {
            '/foo': {
              get: {
              },
              post: {
                operationId: 'myId2'
              }
            },
          }
        }}
```

When operationId is not defined for the `get` operation, the __originalOperationId incorrectly gets set to 'myId2' when the post operation got added to the list.

### Motivation and Context

Fixes #1221 

### How Has This Been Tested?

Added extra unit test to cover the above scenario

### Types of changes

### Checklist:

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -146,22 +146,21 @@ export function normalizeSwagger(parsedSpec) {
           map[oid] = [operation]
         }
 
-        Object.keys(map).forEach((op) => {
-          if (map[op].length > 1) {
-            map[op].forEach((o, i) => {
-              o.__originalOperationId = o.__originalOperationId || o.operationId
-              o.operationId = `${op}${i + 1}`
-            })
-          }
-          else if (typeof operation.operationId !== 'undefined') {
-            // Ensure we always add the normalized operation ID if one already exists
-            // ( potentially different, given that we normalize our IDs)
-            // ... _back_ to the spec. Otherwise, they might not line up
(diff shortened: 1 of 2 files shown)


shockey:
Good catch. Thanks for the PR!

PR #1227, opened 2018-01-17
Fix: XSLX file operations

denis-s-svsg:
FIX: operations with XSLS file

### Description
Fix issue about download XSLS files 

### Motivation and Context
Root of issue that content type of XSLX files is 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet' According actual version of application it will be interpreted as text, but should be as blob, to fix it we need to fix RegExp pattern used to detect text files 

### Types of changes

### Checklist:

### DIFF
--- a/src/http.js
+++ b/src/http.js
@@ -65,7 +65,7 @@ export default function http(url, request = {}) {
 }
 
 // exported for testing
-export const shouldDownloadAsText = (contentType = '') => /json|xml|yaml|text/.test(contentType)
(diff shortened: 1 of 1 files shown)


shockey:
Hi @denis-s-svsg! Thanks for this PR.

A question - are you sending a `Content-Disposition: attachment` header in your XLSX file response? I ask because we're planning to prioritize Content-Disposition as a trigger for file download in https://github.com/swagger-api/swagger-ui/issues/4115.

denis-s-svsg:
Hi @shockey, yes we are using `Content-Disposition: attachment`

PR #1247, opened 2018-02-20
resolve  $refs for operation requestBody

RomanGotsiy:
### Description
Operation object's `requestBody` may be a reference to `components/requestBodies` object but this is not supported by this lib which results in `Content-Type` header not being sent (see tests change)

### How Has This Been Tested?
I have tested it using test OAS3 spec from this repo (https://github.com/swagger-api/swagger-js/blob/master/test/oas3/data/petstore-oas3.yaml)

### Types of changes

### Checklist:

### DIFF
--- a/src/execute/oas3/build-request.js
+++ b/src/execute/oas3/build-request.js
@@ -18,7 +18,10 @@ export default function (options, req) {
 
   req = applySecurities({request: req, securities, operation, spec})
 
-  const requestBodyDef = operation.requestBody || {}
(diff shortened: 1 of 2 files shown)


shockey:
Hey @RomanGotsiy, I'm going to pass on this PR since we do have a resolver interface (https://github.com/swagger-api/swagger-js#swagger-specification-resolver) already - if you need to resolve a definition before you use execute/buildRequest directly, you can use that 😄 

Regardless, thanks for the PR!

Issue #1249, opened 2018-02-23
Support asynchronous requestInterceptor


shockey:
_From @dinvlad on February 19, 2018 22:4_

| Q                               | A
| ------------------------------- | -------
| Bug or feature request?         |  feature
| Which Swagger/OpenAPI version?  | 2.0
| Which Swagger-UI version?       | 3.10.0
| How did you install Swagger-UI? | Yarn
| Which browser & version?        | Chrome
| Which operating system?         | Linux/MacOS

### Expected Behavior

`requestInterceptor` should allow its output to be a `Promise` 

### Current Behavior

`requestInterceptor` is currently defined as `req => req`, meaning that all request pre-processing must happen synchronously inside the interceptor.  However, there may be use cases for asynchronous pre-processing (see below).

### Possible Solution

Perhaps patching the current use of `requestInterceptor` to `await`/`then` its result? This way, both types of requests (sync and async) would be handled identically.

### Context

The use case that affects me the most is grabbing an auth token (JWT or OAuth) from user's browser context and attaching it to requests submitted through "Try it out". My application uses Firebase to generate idToken (JWT), which happens asynchronously because the tokens are valid only for 1 hour, and after that the Firebase library would automatically refresh the token through a network request when we try to access it next time.

Currently, to submit my requests through Swagger UI, I use an (async) `getIdToken()` through Firebase library and assign its resolved result as a static string available in `requestInterceptor` context when I configure the SwaggerUI object. However, if a user has stayed on the Swagger UI page for >1 hour, I have to force-refresh their page to re-initialize `requestInterceptor` with the new token (and thus lose the UI state). It would be nicer if we could silently refresh the token in the background using the same `getIdToken()` but inside the `requestInterceptor` itself.

`responseInterceptor` could also be useful in an asynchronous context, although I haven't found a use case for it yet.

_Copied from original issue: swagger-api/swagger-ui#4235_

PR #1261, opened 2018-03-14
Support asynchronous responseInterceptor

rg1:
### Description

Analogous to Issue #1249, I would need the responseInterceptor to be able to modify the response asynchronously.

### Motivation and Context

The idea is for the responseInterceptor to make another request and use the result in place of the previously received one. In particular I need this, because I need to handle redirect requests myself, because I need to receive cookies set on the redirect response. 

### How Has This Been Tested?

I've added a test to the test suite that performs a new request from the responseInterceptor (although from an error response, not a redirect).

### Types of changes

### Checklist:
- [x] I have updated the documentation accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ Swagger.http(request)
 // Interceptors
 Swagger.http({
   requestInterceptor: (req: Request) => Request
-  responseInterceptor: (res: Response) => Response
+  responseInterceptor: (res: Response) => Response | Promise
 })
 
 // Custom Fetch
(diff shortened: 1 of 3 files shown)


rg1:
I've updated the code to use async/await as requested. Sorry that I didn't realize it was supported.

I guess fixing  #1249 is now simply a matter of adding another await (and a test). I could add that as well if you like or submit another pull request once this is merged.

shockey:
@rg1 please do! might as well knock both out at once 😄

rg1:
@shockey I've pushed the change Monday. Is anything else needed?

PR #1278, opened 2018-04-07
Restore swagger2 support for x-tokenName extension in OAuth2 security

ngraef:
### Description

The previously supported `x-tokenName` can be used in order to select another field from the token JSON as bearer token.

### Motivation and Context

As noted in [`swagger-ui`#4084](https://github.com/swagger-api/swagger-ui/issues/4084), v2 supported the `x-tokenName` spec extension to allow for tokens other than `access_token` to be used for "Try it Out" requests with OAuth2 security. Most notably, this allows for the selection of the `id_token` for service authorization. This PR restores that functionality for swagger2 specs.

### How Has This Been Tested?

Added a test case that uses `id_token` for authentication.

### Types of changes

### Checklist:

### DIFF
--- a/src/execute/swagger2/build-request.js
+++ b/src/execute/swagger2/build-request.js
@@ -65,7 +65,8 @@ export function applySecurities({request, securities = {}, operation = {}, spec}
       const value = auth.value || auth
       const schema = securityDef[key]
       const {type} = schema
-      const accessToken = token && token.access_token
+      const tokenName = schema['x-tokenName'] || 'access_token'
+      const oauthToken = token && token[tokenName]
       let tokenType = token && token.token_type
 
       if (auth) {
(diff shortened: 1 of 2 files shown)


shockey:
Code changes look good - handing off to @webron for consideration of the feature as a whole

ngraef:
Thanks for reviewing! Any thoughts, @webron?

webron:
I don't object to bringing back functionality that was supported in previous versions of Swagger-UI but...

- We should consider adding extension functionality support in plugins (and yes, @shockey, that would mean a plugin repository).
- We should consider renaming the extension name based on some work the OAI is doing right now on formalizing extensions, but I'd need to check the status of that.

ngraef:
> We should consider renaming the extension name based on some work the OAI is doing right now on formalizing extensions

This PR adds the `x-tokenName` extension support exclusively for swagger 2 docs. OAI work would only impact the 3.x spec, right?

shockey:
> This PR adds the x-tokenName extension support exclusively for swagger 2 docs. OAI work would only impact the 3.x spec, right?

@ngraef, yep, fair point. Let's move forward with this one.

shockey:
thanks @ngraef!

PR #1281, opened 2018-04-12
WIP: return own promise from `execute`

### LINKED ISSUE #1271: Execute is not throwing error when the parameter expected is not provided
Execute is currently not throwing any error  when the parameter expected is not provided, which causes the code to hang and time out and we get "UnhandledPromiseRejectionWarning: Unhandled promise rejection "

shockey:
# don't merge this!!!

Fixes #1271.

Holding on this, because of a breaking change: the main client `catch` statement will no longer catch thrown errors instance execute calls. This could be viewed as "previously broken", but the change makes catching all errors in one place more difficult.

The workaround for this problem is relatively simple: just return your `client.execute` call in your `then` statement, and the execute promise will flow into your top-level catch.


shockey:
please build

JonathanParrilla:
please build

JonathanParrilla:
please build

JonathanParrilla:
please build

JonathanParrilla:
please build

JonathanParrilla:
please build

JonathanParrilla:
please build

PR #1297, opened 2018-04-23
fix: normalization regression

shockey:
Addresses: 
- https://github.com/swagger-api/swagger-ui/issues/4466
- https://github.com/swagger-api/swagger-ui/issues/4467


shockey:
The issue here was that $ref'd parameters were always failing the existential test in the code, which led to an infinite loop of pushing onto `operation[inheritName]`. 

- This code really should be rewritten, the current implementation feels messy. It's hard to see what's going on.
- I was surprised to find that the current tests didn't catch this. In light of that I added the example from https://github.com/swagger-api/swagger-ui/issues/4467 as a specific test for this bug. The normalizer is version-agnostic, so there isn't a need for a sister OAS3 test.

PR #1314, opened 2018-05-05
v3.8.1-2

shockey:


shockey:
Fixes an issue in the Travis script that was preventing an npm deployment!

PR #1318, opened 2018-05-08
Don't mix import and module.exports

yesmeck:
### Description

According to https://github.com/webpack/webpack/issues/4039#issuecomment-273804003

> You can't mix import and module.exports.

### Motivation and Context

Since swagger-client only pushed compressed file to npm, it's hard to debug. So I install swagger-client from github directly then I got error as described above issue.

### Types of changes

### Checklist:


shockey:
As is, this PR changes the library output: Swagger-Client is now available under `default` when using require:

```js
var client = require("swagger-client")

typeof client === "function" // false
typeof client.default === "function" // true
```

shockey:
> After change module.exports to ES module export, we need to import it by ES module import.

As an isomorphic module, we do our best to support all relevant loader formats, so saying "just use ES module import syntax" isn't a workable solution here. 

> This change won't affect end users, because of we've told webpack export the default property.

@yesmeck, unless I'm missing something here, this _does_ affect folks using `require` - while using ES import syntax, the `default` property is reached into, but CommonJS is not aware of this.

The end result is that ES module users get `function SwaggerClient`, and CommonJS (and probably others) users get `{ default: function SwaggerClient }` - so changing the tests to use ES modules do not solve the underlying problem.

Webpack's `output.libraryExport` option _should_ be exporting default, but it doesn't appear to be doing that correctly, since the tests using CommonJS modules were failing. That's why it's important that we test it 😄

shockey:
thanks for working with me here, @yesmeck!

Issue #1326, opened 2018-05-26
fresh of swagger has unmet dependencies


srl295:
### Q&A (please complete the following information)
 - OS: macOS
 - Environment: Node 8.11.2
 - Method of installation: npm
 - Swagger-Client version: 3.8.5
 - Swagger/OpenAPI version: n/a

### Swagger-Client usage:
```js
require('swagger-client')
```

### Describe the bug you're encountering
```
Error: Cannot find module 'base64-js'
```

Manually adding that module caused a "cannot find module `ieee754`.

### To reproduce...

```
mkdir /tmp/junk
cd /tmp/junk
nvm i 8.11.2
npm init -q -y
npm install swagger-client
node -p "require('swagger-client')"
```
### Expected behavior
probably nothing, no error message

### Additional context or thoughts
I hit this by removing my `package-lock.json` and trying to rebuild it, chasing a lodash dependency issue.

srl295:
gist here https://gist.github.com/srl295/782b21a24f70f963fefb541e23373984

shockey:
Thanks @srl295, going to look into this now.

vrsbrazil:
Is there a swagger-client version where this is not happening?

vrsbrazil:
# FOUND DEPENDENCY ERROR

It is the deep-extend 0.6.0 version that is breaking the code, changed the package.json version to fixed 0.5.1 at the swagger-client and it worked.

shockey:
Okay, I've been chasing this for a while and haven't nailed it down yet. This appears to be new in 3.8.5, I can require prior versions just fine.

Firstly, I haven't been able to reproduce this in my local environment yet - it looks like Swagger-Client's development dependencies do pull in `base64-js`, so it's hard to run tests or build locally for this. 

Second... only one production dependency mentions `base64-js` at all:

```
➜  js git:(master) ag -l "base64-js" ./node_modules     
node_modules/fast-json-patch/package-lock.json
```

This is a `package-lock.json`, which shouldn't be in published packages at all - but beyond that, [`fast-json-patch`](https://www.npmjs.com/package/fast-json-patch) hasn't released in 7 months, and we didn't change any usage of it last week, so it's very unlikely that this is the cause.

---



In fact, we changed very little in Swagger-Client last week ([diff](https://github.com/swagger-api/swagger-js/compare/v3.8.4...master)). 

[...]

PR #1327, opened 2018-05-28
Locked deep-extend version to 0.5.1

vrsbrazil:
Locked deep-extend version to 0.5.1

### Description
https://github.com/swagger-api/swagger-js/issues/1326

### Motivation and Context
It is the deep-extend 0.6.0 version that is breaking the code, changed the package.json version to fixed 0.5.1 at the swagger-client and it worked.

### How Has This Been Tested?
Locally tested, I can send evidences if needed

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "btoa": "1.1.2",
     "cookie": "^0.3.1",
     "cross-fetch": "0.0.8",
-    "deep-extend": "^0.5.1",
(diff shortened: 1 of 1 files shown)


shockey:
@vrsbrazil, did this end up not being the source of the issue?

vrsbrazil:
It wasn't, however I noticed that if I go to node_modules/swagger-client and run npm install in there, it works too

PR #1331, opened 2018-06-22
URL encode parameters for OpenApiSpec 3

artemyarulin:
URL encode parameters for OpenApiSpec 3

### Description
Fixes this issue in swagger-ui https://github.com/swagger-api/swagger-ui/issues/4025

### Motivation and Context

I guess it was disabled by accident in https://github.com/swagger-api/swagger-js/pull/1181 which was about allowing to skip url encode for query params if `Paramter.allowReserved` is set to true. But [specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md) says clearly: 

`
Determines whether the parameter value SHOULD allow reserved characters, as defined by RFC3986 :/?#[]@!$&'()*+,;= to be included without percent-encoding. This property only applies to parameters with an in value of query. The default value is false.
`. So it should be applied only to query params and path parameters should use encoding all the time

### How Has This Been Tested?

I'm not familiar with your testing strategy, but I've recompiled your library and checked that it works correctly using swagger-ui:

Before:
image
After:
image

### Types of changes

### Checklist:

### DIFF
--- a/src/execute/oas3/parameter-builders.js
+++ b/src/execute/oas3/parameter-builders.js
@@ -14,7 +14,7 @@ function path({req, value, parameter}) {
     value,
     style: style || 'simple',
     explode: explode || false,
-    escape: false,
+    escape: true,
   })
 
   req.url = req.url.replace(`{${name}}`, styledValue)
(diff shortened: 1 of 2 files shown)


shockey:
Note to self: the fact that changing `false` to `true` here didn't break any tests is concerning - may need to add additional unit tests along with this PR

artemyarulin:
Anything I can do to push it forward? Tests, docs, etc.?

shockey:
@artemyarulin, yes! I was planning to add some tests myself, but some cases that fail prior to this change would be great, if you're up to the task 😄

PR #1333, opened 2018-06-29
Allow omission of `allowEmptyValue: true` parameters

shockey:
This PR paves the way for a UI-based resolution of https://github.com/swagger-api/swagger-ui/issues/4223.

### Rationale

TLDR: `allowEmptyValue` is different from `required`.

Previously, there was no way to express to Swagger Client that an `allowEmptyValue` parameter should not be present. 

### Previous behavior 

Given:
```yaml
parameters:
- name: myParam
  in: query
  allowEmptyValue: true
```

All of these execute `parameters` values would result in `/?myParam=`:

```js
{}
```

```js
{
  myParam: undefined
}
```

```js
{
  myParam: ""
}
```
### New behavior 

Given:
```yaml
parameters:
- name: myParam
  in: query
  allowEmptyValue: true
```

....these execute `parameters` values result in the following:

```js
{} 
```

yields `/`.

```js
{
  myParam: undefined
}
```

yields `/`.

```js
{
  myParam: ""
}
```

yields `/?myParam=`.

### TODO

- [x] Review parameter types covered by the tests

### Request for review

_@webron_

This is, strictly speaking, a breaking change - but the behavior is due to a nonstandard implementation present since `3.0.0`. I'd appreciate your read on whether this should be: 
1. shipped as a bugfix, 
2. held back as a breaking change, or 
3. modified so that `{}`'s behavior doesn't change, but the more esoteric `{ myParam: undefined }` form does change.

### DIFF
--- a/src/execute/oas3/parameter-builders.js
+++ b/src/execute/oas3/parameter-builders.js
@@ -85,7 +85,7 @@ function query({req, value, parameter}) {
       }
     }
   }
-  else if (parameter.allowEmptyValue) {
+  else if (parameter.allowEmptyValue && value !== undefined) {
     const paramName = parameter.name
     req.query[paramName] = req.query[paramName] || {}
     req.query[paramName].allowEmptyValue = true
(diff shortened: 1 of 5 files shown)


webron:
I think the breaking change is acceptable and should be treated as a bug fix as you suggested.
There are currently still discussions on what the implementation of `allowEmptyValue` means, and this may change even further depending on that decision.

Issue #1330, opened 2018-06-08
Swagger.resolve gives error when parsing/resolving response of definition URL


brahmdev:
### Q&A (please complete the following information)
 - OS: Windows
 - Environment: Version 66.0.3359.181 (Official Build) (64-bit), Node version: 8.9.1
 - Method of installation: npm
 - Swagger-Client version: 3.5.2
 - Swagger/OpenAPI version:  OpenAPI 3.0

### Content & configuration

Swagger/OpenAPI definition:
```yaml
https://raw.githubusercontent.com/Bungie-net/api/master/openapi.json
```

Swagger-Client usage:
```js
Swagger.resolve({spec: parsedSwagger }).then((resolved) => {
        resolved.errors; // resolution errors, if any
        resolved.spec;   // the resolved spec
        if(resolved.errors.length === 0) {
            //do something
        } else {
            //do something
        }
    });
```

### To reproduce...

1. Try to parse/resolve the content of  "https://raw.githubusercontent.com/Bungie-net/api/master/openapi.json" using the "Swagger.resolve" method.
2. It gives "Cannot read property 0 of undefined".

### Expected behavior
**Swagger.resolve** should parse/resolve the definition content properly without any errors.
### Additional context or thoughts
When looked into the Definition then there are lots of cyclic references are present so it could be the problem but not sure exactly.

shockey:
This is probably caused by allof/$ref interaction in the resolver

ponelat:
Seeing similar behaviour, created a test case... tried to go down the rabbit whole, and kept finding places where a patch was being used, even though it was dead ( it pointed to a location that doesn't exist anymore ). Tried several things before running out of time :/

Failing test:
https://github.com/swagger-api/swagger-js/pull/1339

shockey:
Closing, looks like this is fixed: https://runkit.com/shockey/5b9adbf43a2b3b00126cb55d

PR #1339, opened 2018-07-04
add failing test for edge case resolver issue. allOf + items + $refs

ponelat:
Adds a failing test to support https://github.com/swagger-api/swagger-js/issues/1330


shockey:
Fixed, presumably by #1365.

This class of issue isn't completely fixed yet (see the label above), but we're thinning it out 😄

PR #1374, opened 2018-08-26
Add simple check to make `isFile` compatible with React Native

simonheys:
### Description
This change adds an extra simple check to the `isFile` method which is part of serializing form data e.g. with `multipart/form-data` requests. If running in React Native, the method checks for the shape of the React Native file object by checking for its `uri` parameter. The object is not of type `File` in React Native so the original code did not work.

### Motivation and Context
I am using the library in React Native projects and needed the ability to upload files.

### How Has This Been Tested?
Manually testing in production apps, added test case.

n/a

### Types of changes

### Checklist:


simonheys:
Thank you @shockey for the clear review. I have added initial test implementation to cover the points you made, looking forward to hearing what you think so far.

shockey:
Merged - the test cases you wrote are just what I was looking for. Thanks @simonheys!

PR #1386, opened 2018-09-15
Fix lack of leading zero on low unicode values

adjenks:
### Description

Low value unicode characters were not encoded properly. For example, 'LINE FEED (LF)' (U+000A) was encoded as '%A' instead of '%0A' which cannot be decoded properly.

### Motivation and Context

Low value unicode characters were not encoded properly. For example, 'LINE FEED (LF)' (U+000A) was encoded as '%A' instead of '%0A' which cannot be decoded properly.

### How Has This Been Tested?

I tested the output directly in the browser

image

### Types of changes


shockey:
thanks @adjenks!

Issue #1418, opened 2019-01-09
Content-Type `application/json; charset=utf-8` is not parsed with `JSON.parse`


extempl:
It uses `jsYaml.safeLoad(body)` (which is a bit longer) instead, because of strict comparison `contentType === 'application/json'`.

Is it possible to use `includes()` entry check or something like this? Will it affect parse result (especially with different charsets)?

I see there is an issue alike: https://github.com/swagger-api/swagger-js/issues/153, but for a different state of the code

extempl:
The reason it should be supported even that charset `utf-8` is the default one for the JSON is security concerns: https://portswigger.net/blog/json-hijacking-for-the-modern-web

extempl:
@shockey Any progress on this? As I understand, there should not be any issues with parsing for different charsets, as there should not be any for json. Can the MR be merged or is there any issue with it? 
I've come here while refactoring of the app for performance reasons, and curious, is it can be fixed in short time or should I just include fork with the fix? `JSON.parse` works twice faster than the `jsYaml.safeLoad` and for 20+ requests in total, one or two of which is big enough, performance affection may be quite big.

shockey:
@extempl, I'll take a look at the PR this week 😄

extempl:
@shockey Great, thank you!
But for being able to update on this version we need to create a new release, could you do it please or is not enough for a new release?

shockey:
@extempl, new release (`v3.20.6`) is now live on npm 👍

shockey:
Closing as resolved!

extempl:
`3.8.23` btw, right? :)

shockey:
@extempl indeed... sorry, `v3.20.6` was the new Swagger UI version 🤦‍♂️

PR #1419, opened 2019-01-21
Use `JSON.parse` method for all types and encodings of `json`

extempl:
https://github.com/swagger-api/swagger-js/issues/1418


shockey:
thanks @extempl!

PR #1423, opened 2019-02-15
fix: block resolution of item-level `example`

shockey:
Fixes https://github.com/swagger-api/swagger-ui/issues/4931.


shockey:
New behavior:

image

PR #1437, opened 2019-06-04
[WIP Proposal] eliminate Specmap's ContextTree

shockey:
### Description

This PR eliminates Specmap's notion of a `ContextTree` for storing remote document contexts. 

Instead, `$ref`s found within newly-loaded remote content will be rewritten themselves as remote references with absolute URLs.

### Motivation and Context

This PR builds off of #1435's work, using the same strategy of rewriting `$ref` values... but instead of doing so only when needed to work around ContextTree's limitations, this PR eliminates ContextTree entirely.

### How Has This Been Tested?

No new tests as this is mostly a subtractive PR -- user-facing functionality is well covered by existing tests.

### Types of changes

### Checklist:


shockey:
comment that was just pulled out of #1435:

```
// ## Problem
// If an allOf member's content is provided by remote $ref, any remaining $ref
// values within the content will still be expressed from the point of view of
// the remote document. Usually this is handled by Specmap's ContextTree, but
// allOf causes the ContextTree to be out of sync, because the ContextTree can't
// track the original source of each member's contents once it has all been
// collapsed into one object.
//
// ## Example
//
// (a.yaml)
// myObj:
//   allOf:
//   - $ref: "./b.yaml#one"
//
// (b.yaml)
// one:
//   $ref: "#/two"
// two:
//   hello: "world"
//
// When the $ref in `a.yaml` is resolved, ContextTree knows that ["myObj",
// "allOf", 0] is coming from `b.yaml`, but once the allOf plugin collapses the
// allOf array into a single result object at ["myObj"], that context is lost.
//
// As a result, the resolver incorrectly reads the `#/two` $ref as a local ref
// within `a.yaml`, and an error is generated.
//
// Furthermore, even if ContextTree could keep up with each $ref, we'd still be
// *possibly* handing back invalid $ref values to the user, because the resolver
// is sometimes throttled by path restriction and depth limitations.
//
// ## Solution
// Generate a set of patches that will rewrite any $refs within an allOf member
// to be valid within their new position in the root document, using the
// existing ContentTree data as a basis for rewriting each reference.
//
// Essentially: persist the ContextTree information by writing it into the $ref.


```

PR #1445, opened 2019-06-09
housekeeping: adopt Prettier

shockey:
This PR:
* introduces Prettier to our codebase
* makes an initial run of Prettier on all JavaScript files in our codebase
* scales back our ESLint to not enforce formatting rules (see ["Prettier vs Linters"](https://prettier.io/docs/en/comparison.html))
* adds a test step that fails if any files are not in line with Prettier's output

### TODO
- [ ] install a pre-commit hook
- [ ] review all changes for code correctness


char0n:
Closing for now as the code changed significantly since this PR.

Issue #1385, opened 2018-09-13
Support deep objects for query parameters with deepObject style


bajtos:
### Content & configuration

 Swagger/OpenAPI definition:

 ```yaml
parameters:
  filterParam:
    in: query
    name: filter
    schema:
      type: object
    style: deepObject
    explode: true
    description: Options for filtering the results
    required: false
```

 Swagger-Client usage:
 ```js
 SwaggerClient({
   // no special configuration is needed
 })
 ```

### Is your feature request related to a problem?

Many applications expect deeply nested objects in input parameters, see the discussion in swagger-ui starting from this comment: https://github.com/swagger-api/swagger-ui/issues/4064#issuecomment-357417513 In [LoopBack](https://loopback.io), we are running into this problem too, see https://github.com/strongloop/loopback-next/pull/1679.

Consider the `filter` parameter described above (`in:query style:deepObject explode:true`).

Let's say the user provides the following value, for example by entering a JSON into the text area rendered by swagger-ui:

```json
{
  "name": "ivan",
  "birth-date": {
    "gte": "1970-01-01"
  }
}
```

At the moment, swagger-js (and swagger-ui) silently ignores the value, produces an empty query and does not even let the user know that something went wrong.

### Describe the solution you'd like

The following query string should be created by swagger-js client for the input value shown above.

```
filter[name]=ivan&filter[birth-date][qte]=1970-01-01
```

### Describe alternatives you've considered

There are no alternatives as far as I can tell.

In the past, when Swagger 2.x did not support deepObject style, we were describing our `filter` argument as `string` and allowing clients to send the deep object in a JSON string. This was our custom extension that was of course not supported by swagger-js and swagger-ui, creating a suboptimal user experience.

### Additional context

We are using `swagger-ui-dist@3.17.0`, which is the latest version published to npm. Unfortunately I am not able to tell which version of swagger-js is used by this version of swagger-ui-dist :(

The proposed serialization style is supported by https://www.npmjs.com/package/qs, which is used by http://expressjs.com as the default query parser, which means that a vast number of Node.js API servers are already expecting this serialization style.

I found the following two older comments that may be relevant:

https://github.com/swagger-api/swagger-js/pull/1140
> Limitations:
> deepObject does not handle nested objects. The specification and swagger.io documentation does not provide an example for serializing deep objects. Flat objects will be serialized into the deepObject style just fine.

https://github.com/swagger-api/swagger-js/pull/1140#issuecomment-333017825
[...]

PR #1450, opened 2019-07-06
Adding nested object support for deepObject style

DarkSorrow:
Adding deepObject query parameters serialisation and fix error when a value was null in the object

### Description
Created a function to transform the object pass as deepObject stype to query parameters, it uses the same style serialisation as before.

### Motivation and Context
My company is using this style of GET method to make filtered query, it seems that many REST API use this method as well to benefit from cache and such. I've read the post such as #1385 and i know that you are talking about asking people from the draft to change this but i know a lot of people are already using this style so i thought i'll just contribute the code that worked for me.
Just like people use nutella to make ice cream instead of just their bread like the picture i suppose it is fine to let them use nested object in their get if they want this :x

### How Has This Been Tested?
Started using it in our production environement

### Types of changes

- [x] No code changes (changes to documentation, CI, metadata, etc)

### Checklist:


ethlo:
Any progress in this area?

DarkSorrow:
Well i'm using my code for deep objects since last year and it seems to be working pretty well for the different use case we had

ethlo:
Thank you. Any reason why this is not merged yet? Having PRs hanging around for a year seems quite counter-productive :-)

char0n:
Hi @DarkSorrow,

First of all thank you for contributing this change!

As mentioned in https://github.com/swagger-api/swagger-js/issues/1385 and subsequent discussion in https://github.com/OAI/OpenAPI-Specification/issues/1706 with no clear decision, at this point even with [OAS 3.1](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.1.0.md) it is still not specified how to serialize deep nested structures using `deepObject`. 

This PR would force everybody using swagger-client to use the style of serialization that your company users. You can understand we are reserved to merge this PR. Ultimately, things like `deepObject` should be controlled via extensions, then plugins to implement the extended functionality, and should not be part of the core library.

At this point of time we will not proceed with merging this PR.

Issue #1426, opened 2019-03-15
[OpenAPI 3.0]: Fileupload with multiple files using multipart/form-data


timaschew:
### Q&A (please complete the following information)
 - OS: macOS
 - Environment: Node.js v8
 - Method of installation: npm
 - Swagger-Client version: 3.8.23
 - Swagger/OpenAPI version: OpenAPI 3.0

### Content & configuration

Swagger/OpenAPI definition:
```yaml
openapi: 3.0.0
paths:
  /services:
    post:
      description: Add a new service
      operationId: addService
      requestBody:
        $ref: '#/components/requestBodies/Service'

  requestBodies:
    Service:
      content:
        multipart/form-data:
          schema:
            type: object
            required:
              - manifest
              - configuration
            properties:
              manifest:
                type: string
                format: binary
                description: Defines the service and data interfaces. In YAML format.
              configuration:
                type: string
                format: binary
                description: Configuration file. In YAML format.
          encoding:
            manifest:
              contentType: text/yaml; charset=utf-8
            configuration:
              contentType: text/yaml; charset=utf-8
```

Swagger-Client usage:
```js
const fs = require('fs')
const path = require('path')
const Swagger = require('swagger-client')
const yaml = require('js-yaml')

const content = fs.readFileSync(path.join(process.cwd(), 'openapi-spec.yaml'), 'utf8')
const options = {
	spec: yaml.safeLoad(content)
}

const file1 = fs.readFileSync(path.join(process.cwd(), 'manifest.yml'), 'utf8')
const file2 = fs.readFileSync(path.join(process.cwd(), 'config.yml'), 'utf8')

new Swagger(options).then(client => {
  return client.apis.default.addService({
    manifest: file1,
    configuration: file2
  })
})
```

### Actual Output
This is the output of the server
```
rg.springframework.web.multipart.MultipartException: Failed to parse multipart servlet request; nested exception is javax.servlet.ServletException: org.apache.tomcat.util.http.fileupload.FileUploadBase$InvalidContentTypeException: the request doesn't contain a multipart/form-data or multipart/mixed stream, content type header is null
```

### Expected Output
Works without any error.  
The server works fine when using **Try it out** in the Swagger editor.

Lukas-Kullmann:
I'm kind of the same issue. For me, the backend server complains, that a required request part is not present.

[...]

PR #1452, opened 2019-07-10
Send streams directly when using OAS3

Lukas-Kullmann:
### Description

It basically fixes the problem reported in [this comment](https://github.com/swagger-api/swagger-js/issues/1426#issuecomment-509713866).

The problem is, that streams are serialized by the swagger client to a string (via JSON.stringify), which should not be done. They should be passed on to the `FormData` as is.

### How Has This Been Tested?

I tested uploading files via stream. Before it did not work, after it did.

### Types of changes

I'm not sure about the breaking change. I removed the Buffer handling from the code. But it was not working anyway. So I'm not sure if you want to treat this as breaking.

The problem is, that for buffers, a file name has to be set since it cannot be inferred by FormData. (See form-data's [alternative submission methods section](https://github.com/form-data/form-data#alternative-submission-methods)):

> Form-Data can recognize and fetch all the required information from common types of streams (`fs.readStream`, `http.response` and `mikeal`'s request), for some other types of streams you'd need to provide "file"-related information manually

The problem is, that we cannot just set `buffer` or some constant. The server can then not infer the file type since there is no mime type and no file extension for the file.

### Checklist:

I'm not sure about code style. The linter does not compalin, but I don't know if you have any additional code style rules.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # Swagger Client <img src="https://raw.githubusercontent.com/swagger-api/swagger.io/wordpress/images/assets/SW-logo-clr.png" height="50" align="right">
 
 [![Build Status](https://travis-ci.org/swagger-api/swagger-js.svg?branch=master)](https://travis-ci.org/swagger-api/swagger-js)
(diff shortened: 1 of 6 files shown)


tim-lai:
@Lukas-Kullmann Thanks for submitting this PR. The root cause of referenced issue is fixed with https://github.com/swagger-api/swagger-js/pull/1527. The use of streams is for tests, or in non-browser environments.

Issue #1461, opened 2019-08-08
How do I run update swagger.js locally and test against swagger-ui


markdemich:
### Q&A (please complete the following information)
 - OS: Windows 10
 - Environment: N/A
 - Method of installation: NPM
 - Swagger-Client version: Latest
 - Swagger/OpenAPI version: OpenAPI 3.0

### Question
This question is more about updating the swagger.js.  I found a bug in swagger.js that causes external yaml references to not work.

The problem is in the file src/resolver.js in the function makeFetchJSON, we need to add application/yaml to the accept header of the request.

I've done this locally, but I can't figure out how to test this against a local installation of swagger-ui.  I tried npm link but that gives errors when I try to run swagger-ui in dev mode.

I searched online but can't seem to find instructions on how to do this.  Can anyone share how they accomplish something like this?

char0n:
Hi @markdemich,

As this is rather common question, I've prepared a [PR](https://github.com/swagger-api/swagger-js/pull/1630) with documentation how to do local testing of `swagger-client` in `swagger-ui`.

PR #1487, opened 2019-11-25
URL encode the query parameter name

john3300:
### Description

URL encode the query parameter name which can contain restricted characters like square brackets: []

### Motivation and Context

JSON:API suggests using query parameters to specify pagination controls like this: `page[offset]=0`. When using OAS 3.0 the square brackets in this parameter name are not being URL encoded, which causes Try It Out attempts that specify these parameters to fail. This was not a problem with Swagger 2.0 files. The URL encoding did happen there.

This has already been reported in swagger-api/swagger-ui#5304

### How Has This Been Tested?

I have written a unit test for this change which is included in the PR.

I have not been able to test this fully integrated into Swagger UI however.  I'm just not clear on how to do that honestly.  Looks like I'm not the only one: #1461. I would be happy to do that testing if someone can explain how.

### Types of changes

### Checklist:

### DIFF
--- a/src/execute/oas3/parameter-builders.js
+++ b/src/execute/oas3/parameter-builders.js
@@ -87,9 +87,10 @@ export function query({req, value, parameter}) {
       })
     }
     else {
-      req.query[parameter.name] = {
+      const paramName = encodeURIComponent(parameter.name)
+      req.query[paramName] = {
         value: stylize({
-          key: parameter.name,
+          key: paramName,
           value,
           style: parameter.style || 'form',
           explode: typeof parameter.explode === 'undefined' ? true : parameter.explode,
--- a/src/execute/oas3/parameter-builders.js
+++ b/src/execute/oas3/parameter-builders.js
@@ -87,10 +87,10 @@ export function query({req, value, parameter}) {
       })
     }
     else {
-      const paramName = encodeURIComponent(parameter.name)
-      req.query[paramName] = {
+      const encodedParamName = encodeURIComponent(parameter.name)
+      req.query[encodedParamName] = {
         value: stylize({
-          key: paramName,
+          key: encodedParamName,
           value,
           style: parameter.style || 'form',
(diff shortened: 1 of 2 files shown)


john3300:
@shockey Can I get some feedback on this please?

john3300:
@shockey It's been a couple weeks now.  Still looking for your feedback.

webron:
@john3300 thanks for the PR. Due to some circumstances, we're running behind reviewing and merging PRs, but it is not being ignored. Thanks for understanding.

src/execute/oas3/parameter-builders.js
```diff
@@ -37,6 +37,9 @@ export function query({req, value, parameter}) {
     return
   }
 
+  // Encode query parameter name
+  parameter.name = encodeURIComponent(parameter.name)
```
shockey: Instead of mutating the parameter's name here (which would probably cause some unintended consequences elsewhere in this library), I'd like to instead see `encodeURIComponent` being used to encode `parameter.name` where we're passing that value to `stylize()` further down in this function.

shockey:
@john3300:

> I have not been able to test this fully integrated into Swagger UI however. I'm just not clear on how to do that honestly. Looks like I'm not the only one: #1461. I would be happy to do that testing if someone can explain how.

It's not trivial — you'd need to set up a watch build of Swagger Client, `npm link` your Swagger Client build to a local Swagger UI, then run a development build of Swagger UI. For this PR, which only touches Swagger Client's request builder, you don't need to test-drive in the UI since the builder interface is straightforwardly used in the UI.

src/execute/oas3/parameter-builders.js
john3300: I have moved the `encodeURIComponent` call closer to where the value is actually used.

shockey:
thanks @john3300! 🎉

john3300:
> thanks @john3300! 🎉

My pleasure.  Any idea when the next release will happen so this gets picked up by the UI?

PR #1495, opened 2020-01-15
making baseUrl method available for consumption

shah-vidhi:
Exposing **baseUrl** method

### Description
Making **baseUrl** method available for consumption.

### Motivation and Context
We have a requirement where our users desire route URLs on **swagger-ui** to have a copy-to-clipboard functionality. To do so, we need the output of the **baseUrl** method to be appended with a specific path sub URL to obtain the absolute URL for a route. Currently, the **baseUrl** method is not exposed for consumption. 
We are aware of the **buildRequest** method. Unfortunately, it does not work without the mandatory route parameters. However, the functionality that we are looking for just requires fetching the route path without appending the parameter values. 

### How Has This Been Tested?
The baseUrl method has been tested for in the existing test cases. As part of this change, the method has been exposed. Hence, it is not expected to break any existing functionality.

### Types of changes

### Checklist:

**Note**: existing test cases already cover the validations

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -6,7 +6,7 @@ import Http, {makeHttp, serializeRes, serializeHeaders} from './http'
 import Resolver, {clearCache} from './resolver'
 import resolveSubtree from './subtree-resolver'
 import {makeApisTagOperation} from './interfaces'
-import {execute, buildRequest} from './execute'
+import {execute, buildRequest, baseUrl} from './execute'
 import {opId} from './helpers'
 
 Swagger.http = Http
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -20,6 +20,7 @@ Swagger.clearCache = clearCache
 Swagger.makeApisTagOperation = makeApisTagOperation
 Swagger.buildRequest = buildRequest
 Swagger.helpers = {opId}
+Swagger.baseUrl = baseUrl
```
shockey: I'd like to rename the public-facing name for this function to `getBaseUrl`, for consistency with the existing public methods
```suggestion
Swagger.getBaseUrl = baseUrl
```

shah-vidhi:
@shockey I have implemented your suggestions.

shockey:
merged. thanks @shah-vidhi!

PR #1497, opened 2020-01-30
Basic auth: Support empty passwords

Simran-B:
### Description

Don't cast non-string passwords to string (e.g. undefined to "undefined").

I'm not sure if `value.password instanceof String` is needed or if a test for plain strings would be sufficient. Or should it explicitly test for `undefined`? Because people expect whatever they pass to be converted to string implicitly, except the blank password case?

Should the username also be checked?

### Motivation and Context

Empty passwords for basic HTTP authorization is a valid use case (see https://github.com/swagger-api/swagger-ui/issues/5410) and I opened a PR to allow this in the frontend (https://github.com/swagger-api/swagger-ui/pull/5812), but it turned out that the swagger-client casts the undefined `password` to a string, leading to an incorrect base64-encoded password.

- `"root:undefined"` → `"cm9vdDp1bmRlZmluZWQ="` ❌
- `"root:"` → `"cm9vdDo="` ✔

### How Has This Been Tested?

```
webpack --config webpack\test_webpack_build.babel.js
jest --runInBand -t "swagger2 - execute - applySecurities"
jest --runInBand -t "Authorization - OpenAPI Specification 3.0"
```

### Types of changes

### Checklist:

- [x] All new ~and existing~ tests passed.

### DIFF
--- a/src/execute/oas3/build-request.js
+++ b/src/execute/oas3/build-request.js
@@ -148,7 +148,8 @@ export function applySecurities({request, securities = {}, operation = {}, spec}
         }
         else if (type === 'http') {
           if (schema.scheme === 'basic') {
-            const {username, password} = value
+            const username = value.username
+            const password = typeof value.password === "string" || value.password instanceof String ? value.password : ""
             const encoded = btoa(`${username}:${password}`)
             result.headers.Authorization = `Basic ${encoded}`
           }
--- a/src/execute/oas3/build-request.js
+++ b/src/execute/oas3/build-request.js
@@ -148,8 +148,8 @@ export function applySecurities({request, securities = {}, operation = {}, spec}
         }
         else if (type === 'http') {
           if (schema.scheme === 'basic') {
-            const username = value.username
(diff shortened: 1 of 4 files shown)


src/execute/oas3/build-request.js
```diff
@@ -148,7 +148,8 @@ export function applySecurities({request, securities = {}, operation = {}, spec}
         }
         else if (type === 'http') {
           if (schema.scheme === 'basic') {
-            const {username, password} = value
+            const username = value.username
+            const password = typeof value.password === "string" || value.password instanceof String ? value.password : ""
```
shockey: if the concern here is falsy (eg `null` and `undefined`) values, we can just use logical OR to default to an empty string:

```suggestion
            const password = value.password || ""
```

Simran-B: The main concern is undefined, possibly also null, which a user may pass to express the absence of a value. Other problematic cases are unlikely to occur.

Simran-B:
@shockey Should be good to go. Unsure about the Swagger-UI PR however https://github.com/swagger-api/swagger-ui/pull/5812

PR #1515, opened 2020-04-29
trying to avoid usage of `Function` as it requires `unsafe-eval` Cont…

vmukhachev:
…ent Security Policy

Note: I do not know if the old variant does something useful and if the error should not be ignored...

### Types of changes

### Checklist:

### DIFF
--- a/src/specmap/index.js
+++ b/src/specmap/index.js
@@ -350,7 +350,7 @@ class SpecMap {
 
       // Waits for all to settle instead of Promise.all which stops on rejection
       return Promise.all(promises.map((promise) => {
-        return promise.then(Function, Function)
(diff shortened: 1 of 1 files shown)


src/specmap/index.js
```diff
@@ -350,7 +350,7 @@ class SpecMap {
 
       // Waits for all to settle instead of Promise.all which stops on rejection
       return Promise.all(promises.map((promise) => {
-        return promise.then(Function, Function)
+        return promise.then(function () {}, function () {})
```
char0n, line 353: This looks like a noop operation, so let's use one from `lodash`.
```js
import noop from 'lodash/noop';

return promise.then(noop, noop)

Issue #1508, opened 2020-04-22
Headers serialization


char0n:
### Q&A (please complete the following information)
 - OS: [e.g. linux]
 - Environment: [e.g. Chrome 59, Node.js v13.3.0]
 - Method of installation: [e.g. npm, unpkg]
 - Swagger-Client version: dev (commit: 0f6e38837b5d247928d9668e6985bd2507ff4f1b)
 - Swagger/OpenAPI version: [*]

### Content & configuration

Swagger-Client usage:
```js
SwaggerClient({
  // your config options here
})
```

### Describe the bug you're encountering

HTTP Client Header serialization algorithm is expected to produce following results when dealing with multiple headers with the same name:

```js
// eg: Cookie: one
//     Cookie: two
//  =  { Cookie: [ "one", "two" ]
```
This will never happen, as the [Headers](https://developer.mozilla.org/en-US/docs/Web/API/Headers/append) interface already merges headers with the same name to one header field and separating them with `, ` characters.

### To reproduce...

1. Create a simple HTTP node server

```js
const http = require('http');

const requestListener = function (req, res) {
  res.setHeader('Content-Type', ['text/plain', 'application/json']);
  res.writeHead(200);
  res.end('Hello, World!');
}
```
2. Save it as `server.js` and run it
```sh
$ node server.js
```

3. create file `test.js` with following content and run it

```js
'use strict';

const SwaggerClient = require('./index');

const request = {
  url: 'http://localhost:8080',
  method: 'GET',
};

SwaggerClient.http(request).then(res => res.headers).then(console.log)
```
```sh
 $ node test.js
```

Result is: 

```js
{
  'content-type': 'text/plain',
  date: 'Wed, 22 Apr 2020 15:52:38 GMT',
  connection: 'close',
  'transfer-encoding': 'chunked'
}
```

### Expected behavior

Result is:
```js
{
  'content-type': [ 'text/plain', 'application/json' ],
  date: 'Wed, 22 Apr 2020 15:52:38 GMT',
  connection: 'close',
  'transfer-encoding': 'chunked'
}
```

### Additional context or thoughts

[...]

PR #1523, opened 2020-05-01
fix: amend header serialization algorithm

char0n:
### Description

I needed to update very old cross-fetch@0.0.8 to cross-fetch@3.0.4
which uses 2.x branch of node-fetch instead of 1.x branch.
node-fetch@1.x is no longer compatible with fetch API in browsers
and is incompatible with the fetch specification.

Ref #1508

### Motivation and Context

Header serialization issues. More in #1508

### How Has This Been Tested?

Existing tests have been amended to compenseta for fetch API spec changes.

### Types of changes

### Checklist:

### DIFF
--- a/test/http.js
+++ b/test/http.js
@@ -310,9 +310,8 @@ describe('http', () => {
       expect(encodeFormOrQuery(req.query)).toEqual('id=1,2')
     })
 
-    test('should handle custom array serilization', () => {
-      // Given
-      fetchMock.get('*', {hello: 'world'})
+    test('should handle custom array serialization', () => {
+      fetchMock.get('*', {body: 'response body'})
       const req = {
         url: 'http://example.com',
         method: 'GET',
@@ -336,6 +335,9 @@ describe('http', () => {
       // Given
       fetchMock.get('*', (url, opts) => {
         return {
+          headers: {
+            'Content-Type': 'text/plain',
+          },
           body: opts.headers.WilliamCWilliamsHeader
         }
       })
@@ -360,23 +362,32 @@ describe('http', () => {
     test(
       'should serialize fetch-like response and call serializeHeaders',
       () => {
-        const headers = {
-          Authorization: ['Basic hoop-la', 'Advanced hoop-la']
-        }
+        // cross-fetch exposes FetchAPI methods onto global
+        require('cross-fetch/polyfill')
+        const response = new Response('data', { // eslint-disable-line no-undef
+          status: 200,
+          headers: {
+            Authorization: 'Basic hoop-la, Advanced hoop-la',
+            'Content-Type': 'text/plain',
+          }
+        })
 
-        const res = fetchMock.mock('http://swagger.io', {headers})
-
-        return fetch('http://swagger.io').then((_res) => { // eslint-disable-line no-undef
-          return serializeRes(_res, 'https://swagger.io')
-        }).then((resSerialize) => {
-          expect(resSerialize.headers).toEqual({authorization: ['Basic hoop-la', 'Advanced hoop-la']})
-        }).then(fetchMock.restore)
+        return serializeRes(response, 'https://swagger.io')
+          .then((serializedResponse) => {
+            expect(serializedResponse.headers).toEqual({
+              authorization: ['Basic hoop-la', 'Advanced hoop-la'],
+              'content-type': 'text/plain'
+            })
+          })
       }
     )
 
     test(
       'should set .text and .data to body Blob or Buffer for binary response',
       () => {
+        // cross-fetch exposes FetchAPI methods onto global
+        require('cross-fetch/polyfill')
+
         const headers = {
           'Content-Type': 'application/octet-stream'
         }
@@ -392,11 +403,11 @@ describe('http', () => {
         }).then((resSerialize) => {
           expect(resSerialize.data).toBe(resSerialize.text)
           if (originalRes.blob) {
-            expect(resSerialize.data).toBeInstanceOf(Blob) // eslint-disable-line no-undef
+            expect(Object.prototype.toString.call(resSerialize.data)).toBe('[object Blob]')
           }
           else {
             expect(resSerialize.data).toBeInstanceOf(Buffer)
-            expect(resSerialize.data).toEqual(new Buffer(body))
+            expect(resSerialize.data).toEqual(Buffer.from(body))
           }
         }).then(fetchMock.restore)
       }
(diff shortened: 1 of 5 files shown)


test/http.js
```diff
@@ -175,7 +175,7 @@ describe('http', () => {
       // cross-fetch exposes FetchAPI methods onto global
       require('cross-fetch/polyfill')
       expect(global.Headers).toBeInstanceOf(Function)
-      const headers = new Headers() // eslint-disable-line no-undef
+      const headers = new Headers()
```
tim-lai, line 178: 2 things. First, I think we should move  `require('cross-fetch/polyfill')` to global top of test file. Second, we should not remove the `eslint-disable` because it will otherwise still be a linting problem, at least until we are able to import { Headers } directly.

test/http.js
```diff
@@ -360,64 +359,55 @@ describe('http', () => {
     test(
       'should serialize fetch-like response and call serializeHeaders',
       () => {
-        const headers = {
-          Authorization: ['Basic hoop-la', 'Advanced hoop-la']
-        }
-
-        const res = fetchMock.mock('http://swagger.io', {headers})
+        const response = new Response('data', {
```
tim-lai, line 362: I suspect `new Response` will require an eslint-disable. same reason as `new Headers` above

test/http.js
```diff
@@ -360,64 +359,55 @@ describe('http', () => {
...
+            'Content-Type': 'text/plain',
+          }
+        })
 
-        return fetch('http://swagger.io').then((_res) => { // eslint-disable-line no-undef
-          return serializeRes(_res, 'https://swagger.io')
-        }).then((resSerialize) => {
-          expect(resSerialize.headers).toEqual({authorization: ['Basic hoop-la', 'Advanced hoop-la']})
-        }).then(fetchMock.restore)
+        return serializeRes(response, 'https://swagger.io')
```
tim-lai, line 370: keep the eslint-disable comment.

test/http.js
```diff
@@ -360,64 +359,55 @@ describe('http', () => {
...
+          .then((serializedResponse) => {
+            expect(serializedResponse.headers).toEqual({
+              authorization: ['Basic hoop-la', 'Advanced hoop-la'],
+              'content-type': 'text/plain'
+            })
+          })
```
tim-lai, line 376: we need `.catch` statements for all of these promise chains in test file. Node will fail missing catch in future

test/http.js
```diff
@@ -360,64 +359,55 @@ describe('http', () => {
...
-
-        const body = 'body data'
-        const res = fetchMock.mock('http://swagger.io', {body, headers})
+        const response = new Response('body data', {
+          headers: {
+            'Content-Type': 'application/json'
+          }
+        })
 
-        return fetch('http://swagger.io').then((_res) => { // eslint-disable-line no-undef
```
tim-lai, line 415: I think we should remove the fetchMock test, and add a separate non-fetchMock test.

test/http.js
char0n: Regarding eslint-disable:
I've created the the global override for tests [here](https://github.com/swagger-api/swagger-js/pull/1523/files#diff-799dd7ade334108d9d2c335b978f7521R5). As we refer to `Header`, `Request` and `Response` global symbols routinely and I had to write `eslint-disable` more than 4 times it just made sense to make global override. But...I agree that change like this breaks current convention and should be systematic and reflected  everywhere instead having it locally in one file only. So I'll fall back to `eslint-disable` on a line basis and do global symbol allowance in separate PR and that

char0n: Regarding `require('cross-fetch/polyfill')`

I do understand why applying the polyfill is made locally. The original author was trying to achieve complete isolation which I think makes total sense. But the complete isolation is not finalized because [afterEach hook](https://github.com/swagger-api/swagger-js/blob/b9189befb43294ebbdf3c384a8d94c8bfb41e080/test/http.js#L17) doesn't contain code to cleanup the polyfill symbols from global object (Request, Response, fetch, Headers...).

So we have following options:
**1.)** achieve complete isolation by applying polyfill in test bodies but finalize proper clearing of global object
**2.)** as you suggest move polyfilling to top of every file
**3.)** Import node-fetch symbols explicitly from cross-fetch

I'm more inclined for  **3.)** or **1.)** (in that order). Second options just doesn't feel right in context of unitests.

test/http.js
char0n: Nope, I've made a global [override](https://github.com/swagger-api/swagger-js/pull/1523/files#diff-799dd7ade334108d9d2c335b978f7521R5). If it would require the build would fail as the linting is part of build process. Anyway I'll put the eslint disables on line basis instead of global override as discussed [here](https://github.com/swagger-api/swagger-js/pull/1523/files#r419903324)

test/http.js
char0n: No need for disable comment here. `serializeRes` is explicitly imported symbol in global scope.

test/http.js
char0n: What is say is true for `node.js`, but not true in our current context. We're node running our tests in pure `node.js`. We use testing framework [jest](https://jestjs.io/). And jest handles promises for us:

Excerpt from documentation:

> If your code uses promises, there is a more straightforward way to handle asynchronous tests. Return a promise from your test, and Jest will wait for that promise to resolve. If the promise is rejected, the test will automatically fail.

More [here](https://jestjs.io/docs/en/asynchronous#promises)

char0n:
@tim-lai thanks for you CR (code review)! Can you please look at my comments?

test/http.js
char0n: Agree, returned back the tests and fixed their underlying fundamental issue of fetchMock using different version of node-fetch as is used in production build. More [here](https://github.com/swagger-api/swagger-js/pull/1523/files#diff-bd8228415b2a4b7ae1c42e201e63da4b).

test/jest.setup.js
```diff
@@ -0,0 +1,9 @@
+import fetchMock from 'fetch-mock';
+import { Headers, Request, Response } from 'cross-fetch';
+
+fetchMock.setImplementations({
```
char0n, line 4: Here we explicitly make sure that `fetchMock` is using expected version of `node-fetch` and not other `node-fetch` present in our dependency treee.

test/http.js
char0n: Actual header serialization test on multiple values is [here](https://github.com/swagger-api/swagger-js/blob/80f9bf1bb933b60949b8829d94dd3f827d9c40f9/test/http.js#L210).

test/http.js
tim-lai: I believe using `import fetch from cross-fetch` will work, but haven't had a chance to fully confirm or examine other possible artifacts. So let's try that for now.

test/http.js
tim-lai: ok for now. revisit later as needed.

test/http.js
char0n: It will, but that option 3.) and as I said I don't want to make local changes to conventions but rather change the convention for entire codebase in one PR.

PR #1531, opened 2020-05-06
Parse OAS3 HTTP-Auth schemes case-insensitively

asazernik:
According to the authors of the OAI spec (https://github.com/OAI/OpenAPI-Specification/pull/1880#issuecomment-584730191), schemes are case-insensitive. Even if they were not, the current checks against lowercase versions of scheme names do not match the IANA registry's canonical versions (https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml#table-authschemes), which are "Basic" and "Bearer".

### Motivation and Context

If users are overly-adhering to the spec, they may copy-paste HTTP auth scheme names directly from the IANA registry. If they do, the request builder will omit the necessary headers, because it performs a case-sensitive comparison of the scheme name to a lower-case version.

### How Has This Been Tested?

I have not yet tested this change.

### Types of changes

### Checklist:

### DIFF
--- a/src/execute/oas3/build-request.js
+++ b/src/execute/oas3/build-request.js
@@ -147,14 +147,15 @@ export function applySecurities({request, securities = {}, operation = {}, spec}
           }
         }
         else if (type === 'http') {
-          if (schema.scheme === 'basic') {
+          const scheme = schema.scheme && schema.scheme.toLowerCase()
+          if (scheme === 'basic') {
             const username = value.username || ''
             const password = value.password || ''
             const encoded = btoa(`${username}:${password}`)
             result.headers.Authorization = `Basic ${encoded}`
           }
 
-          if (schema.scheme === 'bearer') {
+          if (scheme === 'bearer') {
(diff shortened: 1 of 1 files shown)


asazernik:
This is work-in-progress created from the GitHub UI; I will fix test errors locally.

src/execute/oas3/build-request.js
```diff
@@ -147,14 +147,15 @@ export function applySecurities({request, securities = {}, operation = {}, spec}
           }
         }
         else if (type === 'http') {
-          if (schema.scheme === 'basic') {
+          const scheme = schema.scheme && schema.scheme.toLowerCase()
```
char0n, line 150: We should be able to use lodash [toLower](https://lodash.com/docs/4.17.15#toLower) which returns and empty string of when provided a non-string argument.

```js
import 'toLower' from 'lodash/toLower';
```
```suggestion
          const scheme = toLower(schema.scheme)
```

char0n:
Closing in favor of https://github.com/swagger-api/swagger-js/pull/1562

PR #1547, opened 2020-05-19
Form data monkey patch

tim-lai:
### Description

Replace `formdata-node` with an internal patch of `form-data`. Update tests to reflect dependency change.

### Motivation and Context

commit  b35d36b introduced a dependency change from `form-data` to `formdata-node`. The reason was that `form-data` is missing matching methods for FormData. However, `formdata-node` also required Node 10, which is technically a breaking semver change for `swagger-client`, but not noted as a breaking semver change in the release v3.10.3.

We are choosing to maintain `swagger-client` in a non-breaking semver state. As a result, we are applying an internal patch to `form-data` and removing `formdata-node`.

Note: Node 8 has been EOL for some time now, and remains technically unsupported by `swagger-client`.

### How Has This Been Tested?

Existing tests updated to reflect change between `formdata-node` and `form-data` + patch

### Types of changes

### Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,7 +98,6 @@
   "dependencies": {
     "@babel/runtime-corejs2": "^7.0.0",
     "@kyleshockey/object-assign-deep": "^0.4.0",
-    "@tim-lai/isomorphic-form-data": "^1.0.0",
     "btoa": "1.1.2",
     "buffer": "^5.1.0",
     "cookie": "^0.3.1",
@@ -106,6 +105,7 @@
     "deep-extend": "^0.5.1",
     "encode-3986": "^1.0.0",
     "fast-json-patch": "~2.1.0",
+    "isomorphic-form-data": "^2.0.0",
     "js-yaml": "^3.13.1",
     "lodash": "^4.17.14",
     "qs": "^6.3.0",
--- a/src/internal/form-data-monkey-patch.js
+++ b/src/internal/form-data-monkey-patch.js
@@ -1,0 +1,74 @@
+import isFunction from 'lodash/isFunction'
+import IsomorphicFormData from 'isomorphic-form-data'
+
+// patches FormData type by mutating it.
+// patch :: FormData -> PatchedFormData
+export const patch = (FormData) => {
+  const createEntry = (field, value) => ({name: field, value})
+  /** We return original type if prototype already contains one of methods we're trying to patch.
+   * Reasoning: if one of the methods already exists, it would access data in other
+   * property than our `_entryList`. That could potentially create nasty
+   * hardly detectable bugs if `form-data` library implements only couple of
+   * methods that it misses, instead of implementing all of them.
+   * Current solution will fail the tests to let us know that form-data library
+   * already implements some of the methods that we try to monkey-patch, and our
+   * monkey-patch code should then compensate the library changes easily.
+   */
+  if (
+    isFunction(FormData.prototype.set)
+    || isFunction(FormData.prototype.get)
+    || isFunction(FormData.prototype.getAll)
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -106,6 +106,7 @@
     "deep-extend": "^0.5.1",
     "encode-3986": "^1.0.0",
     "fast-json-patch": "~2.1.0",
+    "isomorphic-form-data": "^2.0.0",
```
char0n, line 109: We have to remove `@tim-lai/isomorphic-form-data` with
```sh
$ npm uninstall --save @tim-lai/isomorphic-form-data
```

test/execute/main.js
```diff
@@ -1,4 +1,4 @@
-import FormData from '@tim-lai/isomorphic-form-data'
+import FormData from 'isomorphic-form-data'
```
char0n, line 1: We'd better use patched FormData here as well.
```js
import FormData from './internal/form-data-monkey-patch'
```

test/http-multipart.js
```diff
@@ -1,4 +1,4 @@
-import FormData from '@tim-lai/isomorphic-form-data'
+import FormData from 'isomorphic-form-data'
```
char0n, line 1: We'd better use patched FormData here as well.
```js
import FormData from './internal/form-data-monkey-patch'
```

src/internal/form-data-monkey-patch.js
```diff
@@ -0,0 +1,69 @@
+import IsomorphicFormData from 'isomorphic-form-data'
```
char0n, line 1: I propose the following change here:
```js
import isFunction from 'lodash/isFunction';
import IsomorphicFormData from 'isomorphic-form-data'

// patches FormData type by mutating it.
// patch :: FormData -> PatchedFormData
export const patch = (FormData) => {
  const createEntry = (field, value) => ({name: field, value})

  // We return original type if prototype already contains one of methods we're trying to patch.
  // Reasoning: if one of the methods already exists, it would access data in other
  // property than our `_entryList`. That could potentially create nasty
  // hardly detectable bugs if `form-data` library implements only couple of
  // methods that it misses, instead of implementing all of them.
  // Current solution will fail the tests to let us know that form-data library
  // already implements some of the methods that we try to monkey-patch, and our
  // monkey-patch code should then compensate the library changes easily.
  if (
    isFunction(FormData.prototype.set)
    || isFunction(FormData.prototype.get)
    || isFunction(FormData.prototype.getAll)
    || isFunction(FormData.prototype.has)
  ) {
    return FormData
  }

  class PatchedFormData extends FormData {
    constructor(form) {
      super(form)
      this._entryList = []
    }

    append(field, value, options) {
      this._entryList.push(createEntry(field, value))
      return super.append(field, value, options)
    }

    set(field, value) {
      const newEntry = createEntry(field, value)

      this._entryList = this._entryList.filter((entry) => {
        return entry.name !== field
      })

      this._entryList.push(newEntry)
    }

    get(field) {
      const foundEntry = this._entryList.find((entry) => {
        return entry.name === field
      })

      return foundEntry === undefined ? null : foundEntry
    }

    getAll(field) {
      return this._entryList
        .filter((entry) => {
          return entry.name === field
        })
        .map((entry) => {
          return entry.value
        })
    }

    has(field) {
      return this._entryList.some((entry) => {
        return entry.name === field
      })
    }
  }

  return PatchedFormData
}

export default patch(IsomorphicFormData)
```
The reasoning is embedded in JavaScript comment directly in code.

tim-lai:
> Looks good! Please have look at my CRC (code review comments) that address multiple aspects of the PR we need to take care of before merge.

resolved

PR #1550, opened 2020-05-26
fix: oas2 collectionFormat when not 'multi'

tim-lai:
### Description

Fix OAS2 collectionFormat handling for non-multipart cases
```
collectionFormat: csv
collectionFormat: tsv
collectionFormat: ssv
collectionFormat: pipes
```

### Motivation and Context

Previous OAS2 multipart fixes did not cover non-multi cases.

### Types of changes

### Checklist:

### DIFF
--- a/src/http.js
+++ b/src/http.js
@@ -3,6 +3,7 @@ import qs from 'qs'
 import jsYaml from 'js-yaml'
 import isString from 'lodash/isString'
 import isFunction from 'lodash/isFunction'
+import isNil from 'lodash/isNil'
 import FormData from './internal/form-data-monkey-patch'
 
 // For testing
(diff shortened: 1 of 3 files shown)


test/http-multipart.js
```diff
@@ -42,6 +47,22 @@ describe('buildRequest - openapi 2.0', () => {
       expect(itemEntries.length).toEqual(2)
       expect(itemEntries[0]).toEqual('person1')
       expect(itemEntries[1]).toEqual('person2')
+      // set of non-multi collectionFormat cases below
```
char0n, line 50: I'd suggest to return to my original comment https://github.com/swagger-api/swagger-js/pull/1527#discussion_r423757494 about how we approach building tests here. We're just adding stuff into the test and making it bigger and more complex. IMHO building tests in this manner will not scale for future and maintainability.

src/http.js
```diff
@@ -214,7 +221,7 @@ function buildFormData(reqForm) {
    * @return {FormData} - new FormData instance
    */
   return Object.entries(reqForm).reduce((formData, [name, input]) => {
-    if (!input.collectionFormat && !input.isOAS3formatArray) {
+    if (!(input.collectionFormat && input.collectionFormat === 'multi') && !input.isOAS3formatArray) {
```
char0n, line 224: More readable suggestion
```suggestion
    if (has(input, 'collectionFormat') && input.collectionFormat !== 'multi' && !input.isOAS3formatArray) {
```
[has](https://lodash.com/docs/4.17.15#has) function comens from `lodash`

test/http-multipart.js
tim-lai: done. separated and categorized 'expects"

src/http.js
char0n: I think my suggestion is logically wrong because I misunderstood the negated conjunction.

char0n, line 224:
```suggestion
    if ((isNil(input.collectionFormat) || input.collectionFormat !== 'multi') && !input.isOAS3formatArray) {
```

PR #1640, opened 2020-07-14
Inject missing ponyfills for `cross-fetch` in UMD build

### LINKED ISSUE #1633: Internet Explorer 11 support
The `browserslist` configuration used to build this project already resolves to IE11, and all the necessary polyfills are injected by `@babel/plugin-transform-runtime`. However, the `cross-fetch` dependency does not ship with a `Promise` polyfill: https://github.com/lquixada/cross-fetch#usage (see the warning at the bottom of the usage documentation).

As a result, the browser bundle contains references to both the "pure" `Promise` ponyfill supplied by `@babel/runtime-corejs2`, and native `new Promise()` calls made by `cross-fetch`.

Could we either document that `swagger-js` will only work in an environment supporting promises, or ensure that the usages in `cross-fetch` are polyfilled? The latter isn't ideal, because it would require choosing between:
1. Bundling `cross-fetch` inside `swagger-js` (i.e. moving it to `devDependencies`), and letting `@babel/plugin-transform-runtime` transform its `Promise` calls.
2. Keeping `cross-fetch` as a dependency, but adding the `Promise` _polyfill_ (i.e. global, what you get when you set `useBuiltIns: "usage"` in `@babel/preset-env`) instead of the ponyfill provided by `@babel/runtime-corejs2`.
3. A hybrid, where the ES module build still has a dependency on an external `cross-fetch` (and is documented as requiring a global polyfill), but the bundled browser build properly ponyfills its calls to `Promise`.

vvanpo:
### Description

Altered the Babel configuration to use a project-wide
`babel.config.json` instead of `.babelrc`, which is not only recommended
by Babel but is require for applying configuration to `node_modules`.

Updated the Webpack configuration used in the UMD build to apply an
exception for `cross-fetch` in the `node_modules` exclusion criteria.

### Motivation and Context

Fixes #1633 

As discussed in #1633, the dependency on `cross-fetch` requires a global
polyfill for `Promise` and others (if I inspect the intermediate Babel
output, I see that `@babel/plugin-transform-runtime` injects ponyfills
for `Object.create`, `Promise`, `Object.getOwnPropertyNames`,
`Array.isArray`, `Symbol.iterator`, and `Symbol`). Leaving it up the
consumer to polyfill makes some of the ponyfilling already done in
`swagger-js`' build redundant, and it is unclear to the consumer exactly
which polyfills are required and necessitates diggin into transitive
dependencies.

For the bundled browser-targeting build, since we already bundle
`cross-fetch`, it only makes sense to let Babel ponyfill what it can.

### How Has This Been Tested?

I added a test to the `test:artifact:umd:browser` suite, that runs a
regex over the output bundle to look for `new Promise\s*\(`. This test
can be verified by running the test suite in 13c115712b3b (I wrote the
test before implementing the changes), and seeing that the test fails.
In the next commit (ba6f85025951) it passes.

### Types of changes

### Checklist:


babel.config.json
```diff
@@ -49,12 +49,7 @@
     },
     "browser": {
       "presets": [
-        [
-          "@babel/preset-env",
-          {
-            "debug": false
-          }
-        ]
+        "@babel/preset-env"
```
char0n, line 52: Can we please return back the extended syntax here? We'll have to rewrite it in future if we want to turn the debugging on.

test/build-artifacts/umd.js
```diff
@@ -1,3 +1,6 @@
+import { readFileSync } from 'fs';
```
char0n, line 1:
```suggestion
import fs from 'fs';
```

test/build-artifacts/umd.js
```diff
@@ -54,6 +57,13 @@ describe('webpack browser umd build', () => {
       },
     });
   });
+
+  test('should not contain references to `new Promise`', () => {
+    const path = join(__dirname, '../../dist/swagger-client.browser.js');
+    const src = readFileSync(path).toString();
```
char0n, line 63:
```suggestion
    const src = fs.readFileSync(path).toString();
```

char0n: Just a convention nitpick

config/webpack/browser.config.babel.js
```diff
@@ -6,6 +6,20 @@ import { WebpackBundleSizeAnalyzerPlugin } from 'webpack-bundle-size-analyzer';
...
 
+const module = {
+  rules: [
+    {
+      test: /\.js$/,
+      // `cross-fetch` requires a number of polyfills (like Promise), so we
+      // let `@babel/plugin-transform-runtime` inject them.
+      exclude: /node_modules\/(?!cross-fetch\/)/,
+      use: {
+        loader: 'babel-loader',
```
char0n, line 17: Are we sure that by removing `babelrc: true`, the webpack is using the `babel.config.json`?

babel.config.json
```diff
@@ -64,6 +59,7 @@
             "version": "^7.10.4"
           }
         ],
+        "@babel/plugin-transform-modules-commonjs",
```
char0n, line 62: Why adding `@babel/plugin-transform-modules-commonjs` here?

config/webpack/browser.config.babel.js
vvanpo: `babelrc` defaults to `true` when a filename is set: https://babeljs.io/docs/en/options#babelrc
`babel-loader` sets the `filename`: https://github.com/babel/babel-loader/blob/master/src/index.js#L131

However, I think the `babelrc` option is only relevant for `.babelrc` files, hence why it requires a filename (as `.babelrc` options are merged on top of one another based on where each file is in the project tree. This does not apply to `babel.config.js` files, which just apply over the entire project and even `node_modules`).

babel.config.json
vvanpo: I'm not sure I fully understand why it's needed, but without it I get a warning:
```
    WARNING in ./src/specmap/lib/refs.js 332:9-14
    "export 'fetch' was not found in 'cross-fetch'
     @ ./src/specmap/index.js
     @ ./src/resolver.js
     @ ./src/index.js
     @ multi ./src/index.js
```
and the resulting bundle breaks with `TypeError: Cannot assign to read only property 'exports' of object '#<Object>'`.

`cross-fetch` uses kind of an odd module definition that I don't really understand, it tries to be compatible with both CommonJS and ESM: https://github.com/lquixada/cross-fetch/blob/master/dist/browser-ponyfill.js#L545

Because `cross-fetch` itself does not use ESM imports, `@babel/plugin-transform-modules-commonjs` does not alter its transpilation. However, it does transform all of _our_ source code's imports, including the imports of `cross-fetch`. For some reason, `webpack` had no problem with this import when it wasn't transpiled. But now that we're running Babel over it, `webpack` can't seem to make sense of it anymore as an ESM import, but doesn't have a problem with it as a CommonJS import. So by running the CommonJS transform over our source files we work around this problem.

config/webpack/browser.config.babel.js
char0n: Thanks for explanation.

babel.config.json
char0n: Thanks for explanation.

char0n: I did some investigation and the answer it documented really well here: https://github.com/webpack/webpack/issues/4039#issuecomment-419284940. I'm suggesting a changes that will fix this problem and don't use `@babel/plugin-transform-modules-commonjs`.

babel.config.json
```diff
@@ -64,6 +64,7 @@
             "version": "^7.10.4"
```
char0n, line 64: Let's transform this file to `babel.config.js` so we can add line comments.

babel.config.json
```diff
@@ -64,6 +64,7 @@
...
           }
         ],
+        "@babel/plugin-transform-modules-commonjs",
```
char0n, line 67:
```suggestion
```

babel.config.json
char0n: Let's amend browser env like this:
```js
"browser": {
  "sourceType": "unambiguous", // https://github.com/webpack/webpack/issues/4039#issuecomment-419284940
  ...rest...
}

char0n: That should solve the issue.

babel.config.json
vvanpo: Nice find, that does explain it succinctly.

char0n:
Hi @vvanpo,

There's been a development in a topic this PR was handling.

More info here: https://github.com/swagger-api/swagger-js/pull/2622

PR #1731, opened 2020-09-24
Create swagger-attach-cookie.js

YaakovHatam:
### Description
An example of creating swagger client object with cookie attached credentials (`include` instead of `same-origin`)

### Motivation and Context
if your api based on cookie authorization you should use it

### Todo:
add this to swagger options, for example
```javascript
 SwaggerClient({
            url: process.env.REACT_APP_SWAGGER_URI,
            **withCredentials: true**
        })
```

### DIFF
--- a/examples/swagger-attach-cookie.js
+++ b/examples/swagger-attach-cookie.js
@@ -1,0 +1,8 @@
+// this line will set the credentials to "include" instead of "same-origin", as a result a cookie will be attached to the request
+SwaggerClient.http.withCredentials = true;
+
+SwaggerClient({
+    url: process.env.REACT_APP_SWAGGER_URI.
+}).then(_swag => {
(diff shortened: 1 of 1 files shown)


char0n:
@YaakovHatam thank you very much for your effort, but IMHO our [CORS documentation](https://github.com/swagger-api/swagger-js/blob/master/docs/usage/http-client.md#cors) covers this pretty well. Please check it out and let me know if your example contains any more info that the documentation.

char0n:
@YaakovHatam I'm going to close this PR. Thanks for the effort thought. We aim to eliminate the `example` directory in near feature and not to extend it. IMHO our [CORS documentation](https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials) covers cookie usecase enough.

PR #2093, opened 2021-05-31
fix: add failback error message in case if HTTP/2 in Chrome

davidyuk:
### Description

I have found that at least in Chrome@90, `fetch` returns an empty string in `statusText` field of the response object. According to MDN:

> The default value is "". Note that HTTP/2 does not support status messages.

https://developer.mozilla.org/en-US/docs/Web/API/Response/statusText#value

`cross-fetch` behaves the same because I think it uses `fetch` as the underlying implementation in this case.

`statusText` is the only source of error messages thrown by `swagger-client`. The lack of `statusText` makes errors look unobvious, like:
```
Error
    at _callee$ (http.js?0c7c:170)
    ...
```

I'm proposing to improve it by adding a fallback message containing HTTP status code in case if `statusText` is empty.

```
Error: response status is 404
    at _callee$ (http.js?0c7c:170)
    ...
```

### How Has This Been Tested?
I've connected this version to the existing project using `npm link`. It behaved as expected.

### Types of changes

### Checklist:

### DIFF
--- a/src/http.js
+++ b/src/http.js
@@ -71,14 +71,14 @@ export default async function http(url, request = {}) {
       // so we'll just throw the error we got
       throw resError;
     }
-    const error = new Error(res.statusText);
+    const error = new Error(res.statusText || `response status is ${res.status}`);
     error.status = res.status;
     error.statusCode = res.status;
     error.responseError = resError;
     throw error;
   }
   if (!res.ok) {
-    const error = new Error(res.statusText);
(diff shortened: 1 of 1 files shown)


char0n (approved):
Thanks! Having a default custom message is better then having a cryptic error nobody understands.

char0n:
@davidyuk would you please rebase this PR on top of current master?

char0n:
This PR was succeeded by https://github.com/swagger-api/swagger-js/pull/2136

PR #2094, opened 2021-05-31
refactor: extract FetchResponseError, expose it for instanceof

davidyuk:
### Description

In our project, we have a global error handler that may catch an exception generated by `swagger-client`. We can detect such exceptions by the persistence of specific fields (`status`, `statusText`, `request`), but would be more robust to be able to use `instanceof` to do that.

We can't do it now because errors generated by `swagger-client` are just instances of `Error` with some additional fields. I'm proposing to add a new error class to use when a response is not ok or failed to process.

### How Has This Been Tested?
I've connected this version to the existing project using npm link. It behaved as expected.

### Types of changes

### Checklist:

### DIFF
--- a/src/http.js
+++ b/src/http.js
@@ -7,12 +7,21 @@ import { Buffer } from 'buffer';
 
 import FormData from './internal/form-data-monkey-patch';
 import { encodeDisallowedCharacters } from './execute/oas3/style-serializer';
+import createError from './specmap/lib/create-error';
 
 // For testing
 export const self = {
   serializeRes,
   mergeInQueryOrForm,
 };
+
+export const FetchResponseError = createError('Error', function cb(response, responseError) {
+  this.message = response.statusText;
+  this.status = response.status;
+  this.statusCode = response.status;
+  if (responseError) this.responseError = responseError;
+  else this.response = response;
+});
 
 // Handles fetch-like syntax and the case where there is only one object passed-in
 // (which will have the URL as a property). Also serilizes the response.
@@ -71,18 +80,10 @@ export default async function http(url, request = {}) {
       // so we'll just throw the error we got
       throw resError;
     }
-    const error = new Error(res.statusText);
-    error.status = res.status;
-    error.statusCode = res.status;
-    error.responseError = resError;
-    throw error;
+    throw new FetchResponseError(res, resError);
   }
   if (!res.ok) {
-    const error = new Error(res.statusText);
-    error.status = res.status;
-    error.statusCode = res.status;
-    error.response = res;
-    throw error;
+    throw new FetchResponseError(res);
   }
   return res;
 }
--- a/src/index.js
+++ b/src/index.js
@@ -9,6 +9,8 @@ import { makeApisTagOperation } from './interfaces';
 import { execute, buildRequest, baseUrl } from './execute';
 import { opId } from './helpers';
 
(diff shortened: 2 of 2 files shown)


char0n:
Hi @davidyuk,

Sure I can see this part of swagger-client. Let's look at the code..

src/http.js
```diff
@@ -14,6 +14,17 @@ export const self = {
   mergeInQueryOrForm,
 };
 
+export class FetchResponseError extends Error {
```
char0n, line 17: Please use this util (just for consistency) to create new Error Type: https://github.com/swagger-api/swagger-js/blob/3b5cc18fd7c738f3954609b0820459ef57e15c11/src/specmap/lib/create-error.js#L5

Here is an example how new Error types are created using this util: https://github.com/swagger-api/swagger-js/blob/57a2436a352997e9426ee795f2a21d2075ae251d/src/execute/index.js#L17-L16

The util handles `Error.captureStackTrace` static method of V8. More about it here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#custom_error_types

src/http.js
```diff
@@ -14,6 +14,17 @@ export const self = {
...
+  constructor(response, responseError) {
+    super(response.statusText);
+    this.name = 'FetchResponseError';
```
char0n, line 20: I'm a little bit worried about changing the name from `Error` to `FetchResponseError` . By setting a name to a new one, people that may be asserting the error name for `Error` will suddenly get something different:  `FetchResponseError`. That makes it kind backward incompatible. What do you think?

char0n (changes requested):
Let's look at the failing tests and address my Core Review comments. We can go from there..

src/http.js
davidyuk: I agree, maybe the name can be changed later in the next major release

davidyuk:
I have update PR and checked it locally:
- `error.toString()` looks like `'Error: Bad Request'`
- `error instanceof require('swagger-client/es/index.js').FetchResponseError` returns true

but it doesn't work with `require('swagger-client/es/index.js')` because of [this](https://github.com/swagger-api/swagger-js/blob/22da4ad9bbe9ea742ad8a20b15f7c10160bb3043/src/commonjs.js#L5-L6), should FetchResponseError be exported as `Swagger.FetchResponseError`? Should we do the same for other Errors created with `createError`?

src/http.js
char0n: Sounds like a good compromise.

char0n:
> error.toString() looks like 'Error: Bad Request'
:+1: 

> error instanceof require('swagger-client/es/index.js').FetchResponseError returns true
> but it doesn't work with require('swagger-client/es/index.js') because of this

To get around this, we can export it as static property of Swagger class as you already suggest. That is the current modus operandi to keep up compatibility with older SwaggerUI versions.

> should FetchResponseError be exported as Swagger.FetchResponseError?
Yes

> Should we do the same for other Errors created with createError?

Depends. When the custom error is propagated to outside world, then we should. When custom error is only used for internal purposes, then we should keep it private.

char0n:
@davidyuk, how are you?  Are you still willing to continue on this PR?

char0n:
Closing this as there's been no activity and responses for some time. If you'd like to continue with the effort, please just reopen the PR and continue.

Issue #2008, opened 2021-04-01
How to use SwaggerClient in web worker


tmkasun:
I'm getting 

```js
browser.js:1 Uncaught ReferenceError: window is not defined
```

This exception is coming from 
`swagger-client -> HTTP -> form-data-monkey-patch -> isomorphic-form-data)`

Exception when trying to use SwaggerClient in a web worker. I did a little bit of a search and found that there is an [issue created](https://github.com/form-data/isomorphic-form-data/issues/5) in `isomorphic-form-data` repo asking for a solution.

I tried setting [outputglobalobject](https://webpack.js.org/configuration/output/#outputglobalobject) in webpack output also, But that didn't work.

### Q&A (please complete the following information)
 - OS: macOS
 - Environment: Chrome Version 89.0.4389.90 
 - Method of installation: npm
 - Swagger-Client version: 3.12.0
 - Swagger/OpenAPI version: OpenAPI 3.0.1

### Content & configuration

Swagger/OpenAPI definition:

[Link to definition](https://github.com/wso2/carbon-apimgt/blob/master/components/apimgt/org.wso2.carbon.apimgt.rest.api.publisher.v1/src/main/resources/publisher-api.yaml)

Swagger-Client usage:
```js
import SwaggerClient from 'swagger-client';

SwaggerClient.resolve({
    url: '/api/am/publisher/v2/swagger.yaml',
    requestInterceptor: (request) => {
        request.headers.Accept = 'text/yaml';
    },
}).then((spec) => {
    self.postMessage(spec);
});
```

### Screenshots

![image](https://user-images.githubusercontent.com/3313885/113243244-472e4680-92d0-11eb-84b4-0b85b9d02749.png)

### How can we help?

Can you please help me in using the SwaggerClient in the work environment? Is SwaggerClient support it or am I doing anything wrong here?

char0n:
swagger-js should work properly in Web Worker environment. The problem is (as you already mentioned correctly in this issue) `isomorphic-form-data` library that we use. Let's continue the discussion there, becase this problem needs to be fix upstream before it's fixed in swagger-js.

char0n:
Added a [comment](https://github.com/form-data/isomorphic-form-data/issues/5#issuecomment-811963897) in `isomorphic-form-data`. If it's not going to be resolved there in some reasonable time, we'll have to find an alternative solution to move things forward.

char0n:
@tmkasun,

[...]

PR #2154, opened 2021-07-25
feat(formdata): use formdata-node to handle form data

char0n:
### Notable changes in this PR

- removed use of isomorphic-form-data library
- added [formdata-node](https://www.npmjs.com/package/formdata-node) library to handle FormData bodies
- added [form-data-encoder](https://www.npmjs.com/package/form-data-encoder) to provide compatibility between formdata-node and node-fetch@2.x
- introduced browser field mapping for isomorphic use and avoiding bundling unnecessary code
- installation runtime now requires Node.js >= 12.4.x
- runtime now requires Node.js >= 12.4.x
- all these changes will allow to use swagger-client in various additional envs like WebWorker or Realm

> Note to runtime requirement: when we migrate to node-fetch@3.x the runtime requirement will be Node.js >= 12.20.x.

This PR is a preparation for using node-fetch@3.x which uses formdata-node for FormData API operations and not the obscure [form-data](https://www.npmjs.com/package/form-data) library which is not spec-compliant and causes various issues.

Refs #2008


char0n:
This PR will fail unless new releases of formdata-node is released with [my change](https://github.com/octet-stream/form-data/commit/53f67a284906428e895b6b0f884de5ce57fbee48).

tim-lai (commented):
@char0n Was this tested in (npm link) SwaggerUI and with "Try It Out"?  We previously tried to migrate to `formdata-node` and deprecate use of `isomorphic-fetch`, but had to roll it back. That said, this PR seems cover additional considerations.

- Node 12.4 requirement: Ok, no issue from CI perspective.
- What is preventing use of `node-fetch@3` in this PR?

.github/workflows/nodejs.yml
```diff
@@ -62,7 +62,7 @@ jobs:
     - name: Use Node.js 8.x
```
tim-lai, line 62:
```suggestion
    - name: Use Node.js 12.4
```

char0n:
> @char0n Was this tested in (npm link) SwaggerUI and with "Try It Out"? 

Yes it has. I've tested manually and run SwaggerUI tests on locally linked swagger-client. Technically this PR brings no change to the SwaggerUI. This is due to following reasons:

- [my change](https://github.com/octet-stream/form-data/commit/53f67a284906428e895b6b0f884de5ce57fbee48) that adds isomorphic support for formdata-node is already present in the version of the library this PR introduces
- browser override in package.json make sure that native FormData instance is passed to native fetch function

> We previously tried to migrate to `formdata-node` and deprecate use of `isomorphic-fetch`, but had to roll it back. That said, this PR seems cover additional considerations.

Yes additional things needed to be done to successfully integrate formdata-node. Author of the library helped me understand the complexities and reasons between formdata-node vs form-data vs node-fetch (v2 vs v3) 

> * Node 12.4 requirement: Ok, no issue from CI perspective.

No issue, minimal version in workflows haves been bumped to 12.4.x

> * What is preventing use of `node-fetch@3` in this PR?

node-fetch@3 is still in beta and ready for production use. I assume they release first stable during one or two months.

octet-stream:
> node-fetch@3 is still in beta and not ready for production use. I assume they'll release first stable during one or two months.

Not to mention that since 3.0.0-beta.10 node-fetch requires native ESM support, which means that you can't use it unless you get rid of CJS in your package, or unless you can use dynamic `import()`.

char0n:
> Not to mention that since 3.0.0-beta.10 node-fetch requires native ESM support, which means that you can't use it unless you get rid of CJS in your package, or unless you can use dynamic `import()`.

Yeah, that'll be fun ;] We already use ESM in 99% places and transpile to CJS, so it shouldn't be that much of a hassle.

char0n:
Ok let's go for a merge and release!

PR #2241, opened 2021-09-15
use extension everywhere

jimmywarting:
### Description

I changed all imports to be explicit so the extension is required.
this can make some stuff work outside of NodeJS where ESM are in use

### Motivation and Context

If you ever decide to add `"type": "module"` in the feature (and potentially remove any build step) then this will help out
full path is the ESM standard spec way of doing things.

### How Has This Been Tested?

npm run test

### Types of changes

- [x] Refactor

### Checklist:

(i had to change one lint rule to fix it...)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -25,6 +25,7 @@
       ],
       "newlines-between": "always"
     }],
+    "import/extensions": ["error", "always", { "ignorePackages": true }],
     "prettier/prettier": "error",
     "no-param-reassign": 0, // needs to be eliminated in future
     "no-use-before-define": [2, "nofunc"] // needs to be eliminated in future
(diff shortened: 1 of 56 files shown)


char0n:
@jimmywarting can you please drop changes for file extensions? If something like that will be introduced it needs to have a separate issue and discussion (It's a major change to all files). Let's keep the scope of this PR to lodash only changes please and not mix two different things.

Does this sound reasonable?

jimmywarting:
> (It's a major change to all files)

it's not a major/breaking change to include the extension...

But i know i should have made a issue first. i just did this while i was waiting for the other PR to get merged first...

char0n:
> it's not a major/breaking change to include the extension...

Right, what I meant was that it's a big PR changing a lot of files

> But i know i should have made a issue first. i just did this while i was waiting for the other PR to get merged first...

Well it's better before any significant effort. Let's create an issue where we draft a plan for ESM conversion, which is something that we'll need to do anyway because node-fetch@3 is already out and we'd like to use it. I'll create a ticket and this PR can be the first thing that's go inside as a preparation for it.

Issue #2187, opened 2021-08-12
Try to remove lodash?


jimmywarting:
https://github.com/swagger-api/swagger-js/blob/3937607d80879c4627a780e7b3a94c1c13ebed48/src/http/index.js#L4-L5
```js
// Native pick
const { a, c } = object;
const result = { a, c };

// native is fn
typeof x === 'function'
```

https://github.com/swagger-api/swagger-js/blob/22da4ad9bbe9ea742ad8a20b15f7c10160bb3043/src/execute/oas3/build-request.js#L3-L4
```js
Object.assign(...)
const securityDef = spec?.components?.securitySchemes || {}
```

char0n:
Hi,

Using lodash over native JS and vice versa is highly debated topic lately. Some of the articles I read lately on the topic includes:

- https://dev.to/jrdev_/the-battle-of-the-array-titans-lodash-vs-vanilla-an-experiment-b6g
- https://areknawo.com/lodash-and-usefulness-of-utility-libraries/
- https://thejs.dev/jmitchell/its-time-to-let-go-of-lodash-nqc
- https://codeburst.io/why-you-shouldnt-use-lodash-anymore-and-use-pure-javascript-instead-c397df51a66
- https://blog.bjoern-ruberg.de/2018/10/06/why-you-should-still-embrace-lodash/

**My take on this topic is:**

It depends on the context...If there is a native JS alternative in this code-base we should probably use it. Code in this code-base is highly imperative anyway, so converting a declarative approach of using utility function from lodash into imperative native JS counterparts wouldn't do any harm. Feel free to issue PRs.


**Note on predicates:**

I would still prefer the use of predicate function, such as `isFunction`, `isString`, `isNumber` and so on. They either come from lodash or are defined locally in utility module. The reason I prefer them is that they are 

- self-contained
- already predefined, tested and cover corner cases
- low risk

By using `typeof x === 'function'` there is too much going on here IMHO, and too much things can go wrong here as well. We can make a simple typo in `function` string very easily and get an unexpected program results.

`typeof x === 'function'` vs `isFunction(x)`

char0n:
@jimmywarting would you like to own this issue and implement the proposed changes?

jimmywarting:
sure thing

char0n:
As part of this issue, `pick` has been replaced with object destructuring. @jimmywarting nice job!

jimmywarting:
[...]

PR #2242, opened 2021-09-15
removed more lodash stuff

jimmywarting:
### Description
got rid of more lodash stuff

### Motivation and Context
part of #2187 

### How Has This Been Tested?

- npm run test
- npm run lint

### Types of changes

- [x] Refactor

### Checklist:

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -1,5 +1,4 @@
 import isObject from 'lodash/isObject';
-import startsWith from 'lodash/startsWith';
 
 const toLower = (str) => String.prototype.toLowerCase.call(str);
 const escapeString = (str) => str.replace(/[^\w]/gi, '_');
@@ -11,7 +10,7 @@ export function isOAS3(spec) {
     return false;
   }
 
-  return startsWith(oasVersion, '3');
+  return oasVersion[0] === '3';
 }
 
 export function isSwagger2(spec) {
@@ -20,7 +19,7 @@ export function isSwagger2(spec) {
     return false;
   }
 
-  return startsWith(swaggerVersion, '2');
+  return swaggerVersion[0] === '2';
 }
 
 // Strategy for determining operationId
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -10,7 +10,7 @@ export function isOAS3(spec) {
     return false;
   }
 
-  return oasVersion[0] === '3';
+  return oasVersion.startsWith('3');
 }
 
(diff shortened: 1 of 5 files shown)


src/helpers.js
```diff
@@ -11,7 +10,7 @@ export function isOAS3(spec) {
     return false;
   }
 
-  return startsWith(oasVersion, '3');
+  return oasVersion[0] === '3';
```
char0n, line 13: wouldn't following form be more readable in this case?

```js
return oasVersion.charAt(0) === '3';
```

By looking at above `charAt` I guess I'd still prefer a `startsWith` method here:

```js
return oasVersion.startsWith('3')
```
It's declarative and it does exactly what it says. What's your take on this?

src/helpers.js
```diff
@@ -20,7 +19,7 @@ export function isSwagger2(spec) {
     return false;
   }
 
-  return startsWith(swaggerVersion, '2');
+  return swaggerVersion[0] === '2';
```
char0n, line 22: Duplicate of https://github.com/swagger-api/swagger-js/pull/2242/files#r709543609

src/http/index.js
```diff
@@ -142,7 +141,7 @@ function serializeHeaderValue(value) {
 //     Cookie: two
 //  =  { Cookie: [ "one", "two" ]
 export function serializeHeaders(headers = {}) {
-  if (!isFunction(headers.entries)) return {};
+  if (typeof headers.entries !== 'function') return {};
```
char0n, line 144: I had a brief note on the predicates like this in my original comment: https://github.com/swagger-api/swagger-js/issues/2187

Given that we have `valid-typeof` eslint rule active, the correctness of these assertions is guaranteed; so let's go for it.

src/helpers.js
jimmywarting: They all solves the issue. I Just took the most easiest, low level, fastest, and shortest method available. 

I don't really care about which of them are being used.
charAt has better support then newer startswith

My method could work for both string and array of it where splited

char0n: This library minimum required versions of Node.js is now at 12.4. So as long as we use features that falls withing this minimum required version we're fine. 

> I don't really care about which of them are being used.

If you don't have preference let use `startsWith`. It increases the readability a bit.

jimmywarting: fixed

char0n (approved):
LGTM. Thank you for contributing!

PR #2247, opened 2021-09-17
build(babel): set commonjs target for Node.js 12.4

char0n:


char0n:
We not require minimum Node.js version of 12.4.

PR #2256, opened 2021-09-24
Inherited parameters should be before operation method level parameters

### LINKED ISSUE #7482: Path-level parameters should be listed before operation-level parameters
### Q&A (please complete the following information)
 - OS: Windows 10
 - Browser: Chrome, Firefox latest 
 - Method of installation: https://editor.swagger.io
 - Swagger-UI version: 3.52.0
 - Swagger/OpenAPI version: any

### Content & configuration
Example Swagger/OpenAPI definition:
```yaml
openapi: 3.0.0
info:
  title: test
  version: 0.0.0
paths:
  /something/{id}:
    parameters:
      - in: path
        name: id
        required: true
        schema:
          type: string
    get:
      parameters:
        - in: query
          name: foo
          schema:
            type: string
      responses:
        '200':
          description: ok
```

### Describe the bug you're encountering
In the API definition provided above, the `/something/{id}` path has path-level parameter `id`, plus the GET operation defines an additional parameter `foo`.

Swagger UI renders the path-level parameter `id` AFTER the operation-level parameters `foo`. This is often undesirable because path-level parameter list often contains `in: path` parameters (which are required), so it makes more sense to render path-level parameters first.

### To reproduce...

1. Paste the sample API definition into https://editor.swagger.io.
2. Expand the operation.

### Expected behavior
Parameter order is: `id` first, then `foo`.

### Actual behavior
Parameter order is: `foo`, `id`.

### Screenshots
image

mathis-m:
### Motivation and Context

Fixes swagger-api/swagger-ui#7482

### How Has This Been Tested?

Adjusted given tests to test new order

### Types of changes

- [x] Improvement

### Checklist:

- [x] I have edited tests to cover my changes.

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -205,7 +205,7 @@ export function normalizeSwagger(parsedSpec) {
                 operation[inheritName] = inherits[inheritName];
               } else if (inheritName === 'parameters') {
                 // eslint-disable-next-line no-restricted-syntax
-                for (const param of inherits[inheritName]) {
+                for (const param of inherits[inheritName].slice(0).reverse()) {
                   const exists = operation[inheritName].some(
                     (opParam) =>
                       (opParam.name && opParam.name === param.name) ||
@@ -215,7 +215,7 @@ export function normalizeSwagger(parsedSpec) {
                   );
 
                   if (!exists) {
-                    operation[inheritName].push(param);
+                    operation[inheritName].unshift(param);
                   }
                 }
               }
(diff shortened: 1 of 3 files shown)


src/helpers.js
```diff
@@ -205,7 +205,7 @@ export function normalizeSwagger(parsedSpec) {
                 operation[inheritName] = inherits[inheritName];
               } else if (inheritName === 'parameters') {
                 // eslint-disable-next-line no-restricted-syntax
-                for (const param of inherits[inheritName]) {
+                for (const param of inherits[inheritName].slice(0).reverse()) {
```
sauravazad, line 208: Why not use a vanilla for loop and iterate in reverse order , that would save one extra loop that happens with **.reverse**  and  **.slice()**

PR #2289, opened 2021-10-13
clean code 

gabrielsimongianotti:
### Description

I made some changes to the code to make it cleaner and more readable


jimmywarting:
☝️ agree with you there, i hate linters that require one curtain way of doing things, i just want a linter to fix errors for not change code style - it's so annoying. one thing doe... i don't like 1 letter variable, annoying do search & replace....

char0n:
@gabrielsimongianotti why closing it? We had couple of things to discuss here, and the codebase would benefit from this.

PR #2309, opened 2021-10-31
Native functions

gabrielsimongianotti:
### Description

i make two alters of https://github.com/swagger-api/swagger-js/issues/2187#issuecomment-920392081, removed lodash/isPlainObject, lodash/isObject and replace per native function

### How Has This Been Tested?

I used your tests

### Types of changes

- [x] remove lodash/isPlainObject
- [x] remove lodash/isObject

### Checklist:

- [x] isObject
- [x] sPlainObject


src/execute/index.js
```diff
@@ -76,7 +75,14 @@ export function execute({
     ...extras,
   });
 
-  if (request.body && (isPlainObject(request.body) || Array.isArray(request.body))) {
```
char0n, line 79: Comparing this change with original, I'd say we'll stick with `isPlainObject` for the time being. Determining if something is POJO is a hard job, and having a well tested utility function that abstract the complexity just make sense. 

I think we have two options with `isPlainObject`
1.) remove the `isPlainObject` from the list and keep using lodash version
2.) create our own utility function that does example the same as `isPlainObject` from lodash


The scope of #2187 (as I understood it) is to replace lodash functions with modern & native simple javascript constructs. IMHO this doesn't fall within that scope. 

// cc @jimmywarting

src/helpers.js
```diff
@@ -136,7 +134,7 @@ export function normalizeSwagger(parsedSpec) {
   for (const pathName in paths) {
     const path = paths[pathName];
 
-    if (!isObject(path)) {
+    if (typeof path !== 'object') {
```
char0n, line 137: `isObject` cannot be simply replaced with `typeof path === 'object'`; this is not functionally equivalent.

The source code of `isObject` is as follows:

```js
    function isObject(value) {
      var type = typeof value;
      return value != null && (type == 'object' || type == 'function');
    }
```

Which means that anything that reports typeof `object` || `function` except `null` and `undefined`.

char0n, line 137:
```suggestion
    if (path == null || !['object', 'function'].includes(typeof path)) {
```

src/helpers.js
```diff
@@ -145,7 +143,7 @@ export function normalizeSwagger(parsedSpec) {
     // eslint-disable-next-line no-restricted-syntax, guard-for-in
     for (const method in path) {
       const operation = path[method];
-      if (!isObject(operation)) {
+      if (typeof path !== 'object') {
```
char0n, line 146:
```suggestion
      if (path == null || !['object', 'function'].includes(typeof path)) {
```

char0n (changes requested):
Thanks for contributing! I've added couple of code review comments. Please have a look at them.

src/execute/index.js
jimmywarting: What @char0n have mention about `isPlainObject` not being equivalent to just a simple `typeof` is true... need to check more things such as function, null etc

in some senarios it's fine to use just `typeof x === object` but we need to know for sure if it's okey to use it.

the point of my issue is not to use any lodash at all, even if you just include one simple function then lodash will at least load some internal core functions that are pretty big in itself and is deeply dependent on many other stuff to make it compatible with es3 or 5 (whatever it is).

Right now there is so few lodash fn that we use that i don't see any point in keeping it, so i suggest # 2 
`create our own utility function that does the same thing`

gabrielsimongianotti: where can i  create function isPlainObject? In which project folder?

char0n: @gabrielsimongianotti `src/helpers.js` contains already two helper functions at the top of the file. Le'ts put `isPlainObject`

char0n: Another options is to use https://www.npmjs.com/package/is-plain-object

Contains the logic that we need and handles corner cases. Has no dependencies as well.

src/execute/oas3/build-request.js
```diff
@@ -41,7 +41,7 @@ export default function buildRequest(options, req) {
     const mediaTypes = Object.entries(operation.responses)
       .filter(([key, value]) => {
         const code = parseInt(key, 10);
-        return code >= 200 && code < 300 && isPlainObject(value.content);
+        return code >= 200 && code < 300 && isPlainObject(value);
```
char0n, line 44:
```suggestion
        return code >= 200 && code < 300 && isPlainObject(value.content);
```

char0n:
Made one change, but the PR is missing changes in `package-lock.json` file, after we've introduced new library.

gabrielsimongianotti:
> Made one change, but the PR is missing changes in `package-lock.json` file, after we've introduced new library.

thank you to remember :sweat_smile:

char0n (changes requested):
As we've introduced new library, we need the `package-lock.json` file change in this PR as well.

gabrielsimongianotti:
> As we've introduced new library, we need the `package-lock.json` file change in this PR as well.

I maked merge in 'upstream/master' , is something missing? you can make approve?

char0n:
> > As we've introduced new library, we need the `package-lock.json` file change in this PR as well.
> 
> I maked merge in 'upstream/master' , is something missing? you can make approve?

You need to run following command in repo fork:

```sh
 $ npm i --save is-plain-object
 ```
 
This will modify two files: `package.json` and `package-lock.json`. These files needs to be part of this PR. We have changes in `package.json` but you didn't submitted changed in `package-lock.json`. 

More info about package-lock files here: https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json

gabrielsimongianotti:
I found this package https://www.npmjs.com/package/is-empty, do you think we can replace lodash/isEmpty with it?

gabrielsimongianotti:
> > > As we've introduced new library, we need the `package-lock.json` file change in this PR as well.
> > 
> > 
> > I maked merge in 'upstream/master' , is something missing? you can make approve?
> 
> You need to run following command in repo fork:
> 
> ```shell
>  $ npm i --save is-plain-object
> ```
> 
> This will modify two files: `package.json` and `package-lock.json`. These files needs to be part of this PR. We have changes in `package.json` but you didn't submitted changed in `package-lock.json`.
> 
> More info about package-lock files here: https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json

sorry i don't speak english very well, now i understand what you were trying to say

char0n (approved):

char0n:
I've made some final changes to the PR, regarding package-lock.json file. We're using `lockfileVersion: 1`, you've generated it for `lockfileVersion: 2`. No big deal, changed it back to v1.

char0n:
Merging and thanks for contributing!

PR #2334, opened 2021-11-10
replace lodash/isEmpty with vanilla

jimmywarting:
what title says 👆

### DIFF
--- a/src/specmap/lib/all-of.js
+++ b/src/specmap/lib/all-of.js
@@ -1,5 +1,3 @@
-import isEmpty from 'lodash/isEmpty';
-
 import { isFreelyNamed, generateAbsoluteRefPatches } from '../helpers.js';
 
 export default {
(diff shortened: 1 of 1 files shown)


src/specmap/lib/all-of.js
```diff
@@ -35,7 +33,7 @@ export default {
     originalDefinitionObj = { ...originalDefinitionObj };
 
     // when we've lost sight, interrupt prematurely
-    if (isEmpty(originalDefinitionObj)) {
+    if (!Object.keys(originalDefinitionObj).length) {
```
char0n, line 36: Nitpick: can we be more explicit here?

```suggestion
    if (Object.keys(originalDefinitionObj).length === 0) {
```

jimmywarting: done

char0n (approved):
LGTM, thanks again!

PR #2431, opened 2022-01-25
fix(security): update cross-fetch to >=3.1.5

char0n:
Refs https://github.com/advisories/GHSA-r683-j2x4-v87g


char0n:
Regression was introduces into node-fetch in version [v2.6.3](https://github.com/node-fetch/node-fetch/releases/tag/v2.6.3)

PR #2437, opened 2022-01-29
fix(specmap): suppress merging examples 

cheremnov:
Fix the Swagger UI [Issue 4175](https://github.com/swagger-api/swagger-ui/issues/4175). 

### Description
Supply a custom merge rule to the `deepmerge` library function to prevent merging examples.

### How Has This Been Tested?
Tested on the example from the issue. Implemented it in `tests/bugs/ui-4175.js`.

### Types of changes

### Checklist:


char0n:
Thanks for the PR!

If I'm looking at [customMerge docs](https://github.com/TehShrike/deepmerge#custommerge) correctly, the function will be triggered for every property called `example`. It means that even for properties that are not OpenAPI `example` fields in top level Schema Objects, right?

src/specmap/lib/all-of.js
```diff
@@ -75,6 +75,12 @@ export default {
       return undefined;
     });
 
+    // If there was an example in the original definition,
+    // keep it instead of merging with examples from subschemas
+    if (originalDefinitionObj.example) {
+      // Delete subschemas examples
+      patches.push(specmap.remove([].concat(parent, 'example')));
```
char0n, line 82: What would you say if we simplify this statement using array literal instead?
```suggestion
      patches.push(specmap.remove([parent, 'example']));
```

I know there is this: https://github.com/swagger-api/swagger-js/blob/master/src/specmap/lib/all-of.js#L84 but it's unnecessary complex IMHO.

test/bugs/ui-4175.js
```diff
@@ -0,0 +1,68 @@
+// https://github.com/swagger-api/swagger-ui/issues/4175
```
char0n, line 1: We broke away from this pattern `/test/bugs/*.js` some time ago. Can you please incorporate tests into https://github.com/swagger-api/swagger-js/blob/master/test/specmap/all-of.js? 

Second note, can you please use minimal OpenAPI definition as a data for your tests? That way the test will verify that it can process the OpenAPI definitions as we expect. THanks!

char0n (changes requested):
Cool! Thanks for the change. Please have a look at my additional code review comments.

test/bugs/ui-4175.js
cheremnov: Thanks for your review! Now the test also verifies definitions, not only examples.

src/specmap/lib/all-of.js
cheremnov: If I understand correctly, the semantics differ. `concat` merges arrays, appending `'example'` to the `parent`. The literal creates the array nesting the `parent` inside.

For the `parent` equal to `['definitions', 'model']`, the array literal produces `[['definitions', 'model'], 'example']` instead of `['definitions', 'model', 'example']`.

The `remove` patch takes an array of strings as the `path` field. It may not handle the nested array correctly.

char0n: You're absolutely right, I didn't realize that parent is an array.

char0n (approved):
This is good to go! I've simplified the fixtures. I was having problem following the hal+link+model example. 

Thank you for contributing

src/specmap/lib/all-of.js
```diff
@@ -75,6 +75,12 @@ export default {
       return undefined;
     });
 
+    // If there was an example in the original definition,
+    // keep it instead of merging with examples from subschemas
```
char0n, line 79:
```suggestion
    // keep it instead of merging with examples from other schema
```

src/specmap/lib/all-of.js
```diff
@@ -75,6 +75,12 @@ export default {
...
+    if (originalDefinitionObj.example) {
+      // Delete subschemas examples
```
char0n, line 81:
```suggestion
      // Delete other schema examples
```

PR #2597, opened 2022-06-30
Support reserved chars in path params

derSascha:
### Description
Add `allowReserved` to path parameter to support reserved characters like the query parameter already does. Default stays the same.

See `allowReserved` in the parameter documentation: https://swagger.io/specification/#parameter-object

### Motivation and Context

We are using URL's with reserved characters like an asterix in the path: `http://example.com/items/*1234abcd`
Swagger always encode this characters.

### How Has This Been Tested?

New test case added.

### Types of changes

### Checklist:

### DIFF
--- a/src/execute/oas3/parameter-builders.js
+++ b/src/execute/oas3/parameter-builders.js
@@ -2,7 +2,7 @@ import stylize, { encodeDisallowedCharacters } from './style-serializer.js';
 import serialize from './content-serializer.js';
 
 export function path({ req, value, parameter }) {
-  const { name, style, explode, content } = parameter;
+  const { name, style, explode, content, allowReserved } = parameter;
 
   if (content) {
     const effectiveMediaType = Object.keys(content)[0];
@@ -18,7 +18,7 @@ export function path({ req, value, parameter }) {
     value,
     style: style || 'simple',
     explode: explode || false,
-    escape: true,
+    escape: allowReserved ? 'unsafe' : true,
   });
 
   req.url = req.url.split(`{${name}}`).join(styledValue);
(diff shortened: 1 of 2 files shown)


derSascha:
The OpenAPI Specification contains a line that needs an update: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md?plain=1#L1036
Is this the documentation file that needs to be updated?

char0n:
Hi @derSascha,

THanks for contributing. We're looking into your PR now.

> The OpenAPI Specification contains a line that needs an update: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md?plain=1#L1036 Is this the documentation file that needs to be updated?

Nope, this repo is just one of the implementations of client library for the OpenAPI 3.0.x. You have to submit your specification change in https://github.com/OAI/OpenAPI-Specification.

char0n:
[According to the OpenAPI 3.0.3](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md) specification, the `allowReserver` says following:

```
This property only applies to parameters with an `in` value of `query`. The default value is `false`.
```

image

This means that when `in` value is `path`, this parameter get's always encoded. IMHO behavior of swagger-js is now compliant with the spec and your PR is going outside of the spec and introducing behavior that is unspecified. 

---

**Getting back to your example**: `http://example.com/items/*1234abcd`

Swagger-js turns your url into: `http://example.com/items/%2A1234abcd`. Can you do url decoding at you server code? You should probably do it anyway as people may request your url in encoded fashion.

char0n:
I'm closing this for now as there's been no follow up from the @derSascha.

PR #2658, opened 2022-09-19
ci(ga): drop Node.js v12 from pipeline

char0n:
Jest no longer runs on Node.js 12. It needs to run tests on Node.js 14 now. We're still test build fragments in Node.js 12.4 though.


char0n:
We use `xmock` that prevents running our tests on Node.js 18.

Issue #2954, opened 2023-05-13
Missing individual parts content-types in `multipart/formdata` requests.


mateusbandeiraa:
### Q&A (please complete the following information)
 - OS: MacOS
 - Environment: Chrome 115.0.5762.4, Node v18.16.0
 - Method of installation: npm
 - Swagger-Client version: 3.19.7
 - Swagger/OpenAPI version: OpenAPI 3.0

### Content & configuration

Swagger/OpenAPI definition:
```yaml
openapi: 3.0.0
info:
  title: 'Test Swagger'
  description: 'Test Swagger'
  contact:
    email: test@swagger.com
  version: '1'
servers:
  - url: https://demo7917836.mockable.io/
paths:
  /api/v1/profile:
    post:
      tags:
        - Profile
      summary: 'profile of the user user'
      requestBody:
        description: 'Profile that needs to be stored'
        content:
          multipart/form-data:
            schema:
              type: object
              properties:
                avatar_image:
                  description: avatar_image
                  type: string
                  format: binary
                profile:
                  $ref: '#/components/schemas/Profile'
            encoding:
              profile:
                contentType: application/json
      responses:
        201:
          description: 'Profile that was registered.'
          content:
            application/json:
              schema:
                type: object
        422:
          description: 'Unprocessable Entity.'
          content:
            application/json:
              schema:
                type: object
components:
  schemas:
    Profile:
      title: Profile
      description: 'profile model'
      required:
        - email
        - telephone
      properties:
        email:
          description: 'Email of the profile'
          type: string
        name:
          description: 'Name of the profile'
          type: string
      type: object
tags:
  -
    name: Profile
    description: 'Operations for the user profile'
```

Swagger-Client usage:
```js
SwaggerClient({
  // your config options here
})
```

### To reproduce...

1. Execute the request `POST /api/v1/profile` with `avatar_image` as a png file, and `profile` as a JSON string

### Expected behavior
The generated request body should be:
```
------WebKitFormBoundaryCqXhiFiwHDcc41AS
Content-Disposition: form-data; name="avatar_image"; filename="Captura de Tela 2023-05-13 às 14.57.47.png"
Content-Type: image/png


------WebKitFormBoundaryCqXhiFiwHDcc41AS
Content-Disposition: form-data; name="profile"
Content-Type: application/json

{
  "email": "string",
  "name": "string"
}
------WebKitFormBoundaryCqXhiFiwHDcc41AS--
```
[...]

PR #2955, opened 2023-05-13
Add individual content-types for each multipart/formdata part

mateusbandeiraa:
### Description
Addresses #2954

### Types of changes

### Checklist:


mateusbandeiraa:
This is my first contribution to this project, so bear with me. I could use some help with making sure I'm not breaking anything with this change.

char0n:
Switching from draft to standard PR

char0n:
I have added `import { Blob } from 'formdata-node';` line as this library is isomorphic, which means it needs to run in Node.js and in browsers. This change make it possible to run in both environments. After this change the tests were run, but your code introduces a regression.

char0n:
@mateusbandeiraa  are you willing to continue with this PR?

mateusbandeiraa:
Yes. Will try to push something later this week.

mateusbandeiraa:
I spent more time studying the code base and I think the fix should be in a different place; `buildFormData()` seems more appropriate. I can even see that for some cases, `Blob` objects are already being used. It's a matter of adding the `{ type: contentType }` argument to the constructor.
https://github.com/swagger-api/swagger-js/blob/1df1de9068a93bcdc0aeecf87eccb55ad829b6b2/src/http/index.js#L333-L368

I'm trying to figure out the cases in which the content type argument should be included. For times where it's explicitly defined in the spec it's obvious, but I think there are default values that can be omitted. (Look for "Encoding Object" at https://swagger.io/specification/).

Still trying to figure this out.

mateusbandeiraa:
Still working on it, probably there's a tidier way to do this. Pushed to get the results from the status checks. Comments are welcome (maybe I'm spending too much of my time on something that wouldn't work?)

char0n:
@mateusbandeiraa thanks for the PR, but we need to replace it with https://github.com/swagger-api/swagger-js/pull/3078

**Justification:** swagger-client facilitates request building and execution for SwaggerUI so we have to be aware how the built requests are later used. With changes in this PR, it would not be possible to pull original data encoded in `File` or `Blob` instances in synchronous way. 

The [PR](https://github.com/swagger-api/swagger-js/pull/3078) that replaces this one, introduced a sub-class of `File` which makes it possible to extract original data from `File` instances in synchronous way to support for example building the `curl` commands for SwaggerUI.

char0n:
Addressed in SwaggerUI by https://github.com/swagger-api/swagger-ui/pull/9105

PR #3137, opened 2023-09-11
feat: replace node-fetch with undici

### LINKED ISSUE #1220: escaping inconsistent
We ran into an issue https://github.com/IBM-Cloud/gp-js-client/issues/113 where a path component (part of a parameter) was for example `a '` (A, space, quote).  Going into the URL Interceptor so that we can add authorization the path was escaped to a URL such as `http://example.com/api/a%20'` - so far so good. But then, when the request was actually sent, the Fetch module ran the URL through `url.parse` again, resulting in `http://example.com/api/a%20%27`.  This results in the same logical behavior, however, the exact URL string is different causing our HMAC calculation to be wrong.

I worked around this in https://github.com/IBM-Cloud/gp-js-client/pull/114/files#diff-38aee4b3886a621f8daddb4104803472R91 by having an interceptor run `url.parse()` on the URL pre-emptively before HMAC calculation, which makes the later parse a no-op. 

I was a little surprised that quote (%27) gets escaped as it is acceptable per HTTP.  Probably the larger issue is that the escaping that happens within swagger-js should match the escaping that actually goes out on the wire.

### LINKED ISSUE #2187: Try to remove lodash?

### LINKED ISSUE #2291: replace node:url with native url?
https://github.com/swagger-api/swagger-js/blob/ba0db610cfbd6b716206e2eefe8b46760606b30c/src/index.js#L86-L102

### LINKED ISSUE #2381: SwaggerClient and data:url's
Hi folks!

I tried to handle data:url for spec instead and faced an exception.

The documentation said 

`var specUrl = 'http://petstore.swagger.io/v2/swagger.json'; // data urls are OK too 'data:application/json;base64,abc...'
`

but it does not work.

Example
```
  const swaggerSpecBase64 = fs.readFileSync(__dirname + '/../swagger.yml').toString('base64');
  client = await SwaggerClient(`data:text/yaml;base64,${swaggerSpecBase64}`);
```

throws 

```
 /source/node_modules/node-fetch/lib/index.js:1309
	throw new TypeError('Only HTTP(S) protocols are supported');
		      ^
  TypeError: Only HTTP(S) protocols are supported

```

What is wrong here?

### LINKED ISSUE #2415: Update node-fetch to v3
`node-fetch@3` is out for some it. Unfortunately it's pure ESM so we cannot use it. We need to convert this library to ESM to use it (https://github.com/swagger-api/swagger-js/issues/2414).

### LINKED ISSUE #2736: querystring is deprecated now, is there any plan to move away from url or to replace url
In swagger-client url is pulling querystring and querystring is deprecated now, is there any plan to move away from url or to replace url.

└─┬ swagger-client@3.18.5
  └─┬ url@0.11.0
    └── querystring@0.2.0

char0n:
`swagger-client` requires Node.js `>=12.20.0` and uses different `fetch` implementation depending
on Node.js version.

- `>=12.20.0 <16.8` - [node-fetch@3](https://www.npmjs.com/package/node-fetch)
- `>=16.8 <18` - [undici](https://www.npmjs.com/package/undici) 
- `>=18` - [native Node.js fetch](https://nodejs.org/dist/latest-v18.x/docs/api/globals.html#fetch)

Closes #1220
Closes #2736
Closes #2415
Closes #2381
Closes #2187
Closes #2291


test/execute/main.js
```diff
@@ -1,4 +1,4 @@
-import { Readable } from 'stream';
+import { FormData } from 'formdata-node';
```
jimmywarting, line 1: node-fetch warns users about the usage of formdata-node cuz it's not spec compatible. 
node-fetch would like to one day remove support for this non-speced variant.

node-fetch, undici and node-fetch-cjs all comes with a spec'ed FormData implementation built right into the package itself. so you do not need to install any additional formdata package

```js
import { FormData } from 'node-fetch'
import { FormData } from 'node-fetch-cjs'
import { FormData } from 'undici'
```

jimmywarting: ups, never mind. read to fast... `form-data` package are discouraged. `formdata-node` is fine to be used. still, unnecessary dependency...

test/http/http-multipart.js
```diff
@@ -1,22 +1,25 @@
-import path from 'path';
-import { Readable } from 'stream';
-import fetchMock from 'fetch-mock';
-import { File, Blob } from 'formdata-node';
+import path from 'node:path';
+import * as undici from 'undici';
+import { FormData, File, Blob } from 'formdata-node';
```
jimmywarting, line 3: i belive this can come from undici as well
`undici.FormData`

char0n:
@jimmywarting thanks for opening this up. I've addressed it in https://github.com/swagger-api/swagger-js/pull/3139

PR #3145, opened 2023-09-14
Update installation.md

tomchiverton:
This section made no sense, referring to a different project. Update it to something that seems to work.

I couldn't find the constructor definition documentation anywhere to link to

### DIFF
--- a/docs/usage/installation.md
+++ b/docs/usage/installation.md
@@ -86,5 +86,9 @@ See unpkg's main page for more information on how to use [unpkg](https://unpkg.c
 
 ### Static files without HTTP or HTML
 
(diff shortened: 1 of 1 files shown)


char0n:
Hi @tomchiverton,

Thanks for the PR. This looks like a fragment that got here during creating the docs by coping. 

Can you transform your PR and actually delete the entire `### Static files without HTTP or HTML` section? This section just doesn't make any sense, and shouldn't be there at all.

Thanks!

char0n:
I've update the PR to align with https://github.com/swagger-api/swagger-js/pull/3145#issuecomment-1734066238.

Thanks for opening this topic up.

PR #3367, opened 2024-02-09
fix(execute): empty field added to multi-value parameters

glowcloud:
Refs https://github.com/swagger-api/swagger-ui/issues/5176

Fixes the issue for when no value is selected, sending an empty string in an array after selecting the '--' option is a separate issue in Swagger UI.

### DIFF
--- a/src/execute/oas3/parameter-builders.js
+++ b/src/execute/oas3/parameter-builders.js
@@ -50,7 +50,7 @@ export function query({ req, value, parameter }) {
     value = '0';
   }
 
-  if (value) {
+  if (value && (!Array.isArray(value) || value.length > 0)) {
     const { style, explode, allowReserved } = parameter;
     req.query[parameter.name] = {
       value,
@@ -79,7 +79,7 @@ export function header({ req, parameter, value }) {
     return;
   }
 
-  if (typeof value !== 'undefined') {
+  if (typeof value !== 'undefined' && (!Array.isArray(value) || value.length > 0)) {
     req.headers[parameter.name] = stylize({
       key: parameter.name,
       value,
@@ -101,7 +101,7 @@ export function cookie({ req, parameter, value }) {
     return;
   }
 
-  if (type !== 'undefined') {
+  if (type !== 'undefined' && (!Array.isArray(value) || value.length > 0)) {
     const prefix =
       type === 'object' && !Array.isArray(value) && parameter.explode ? '' : `${parameter.name}=`;
 
--- a/src/execute/oas3/parameter-builders.js
+++ b/src/execute/oas3/parameter-builders.js
@@ -50,7 +50,7 @@ export function query({ req, value, parameter }) {
     value = '0';
   }
 
-  if (value && (!Array.isArray(value) || value.length > 0)) {
+  if (value) {
     const { style, explode, allowReserved } = parameter;
     req.query[parameter.name] = {
       value,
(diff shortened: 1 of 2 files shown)


src/execute/oas3/parameter-builders.js
```diff
@@ -50,7 +50,7 @@ export function query({ req, value, parameter }) {
     value = '0';
   }
 
-  if (value) {
+  if (value && (!Array.isArray(value) || value.length > 0)) {
```
glowcloud, line 53: Added the same check for `query` despite it working without it - when we had an empty array, it would go inside the if block but we somehow wouldn't have the empty field. Not sure if this change is needed.

glowcloud:
Let me know if the same changes should also be made for `query` and `header` here https://github.com/swagger-api/swagger-js/blob/fix-empty-field-for-multi-value-parameters/src/execute/swagger2/parameter-builders.js

src/execute/oas3/parameter-builders.js
```diff
@@ -79,7 +79,7 @@ export function header({ req, parameter, value }) {
     return;
   }
 
-  if (typeof value !== 'undefined') {
+  if (typeof value !== 'undefined' && (!Array.isArray(value) || value.length > 0)) {
```
char0n, line 82: Can we make it a bit clearer? Now I can read it as sentence, instead of analyzing the boolean logic.

```suggestion
  if (typeof value !== 'undefined' || (Array.isArray(value) && value.length > 0)) {
```

src/execute/oas3/parameter-builders.js
char0n, line 53:
```suggestion
  if (value !== undefined && !(Array.isArray(value) && value.length === 0)) {
```

src/execute/oas3/parameter-builders.js
```diff
@@ -101,7 +101,7 @@ export function cookie({ req, parameter, value }) {
     return;
   }
 
-  if (type !== 'undefined') {
+  if (type !== 'undefined' && (!Array.isArray(value) || value.length > 0)) {
```
char0n, line 104:
```suggestion
  if (type !== 'undefined' && !(Array.isArray(value) && value.length === 0)) {
```

char0n:
Please add a test as well for the cases we're trying to fix  with this change.

char0n:
> Added the same check for query despite it working without it - when we had an empty array, it would go inside the if block but we somehow wouldn't have the empty field. Not sure if this change is needed

We need to know for sure, why this happens.

> Let me know if the same changes should also be made for `query` and `header` here https://github.com/swagger-api/swagger-js/blob/fix-empty-field-for-multi-value-parameters/src/execute/swagger2/parameter-builders.js

I think it will be related to the result of analysis related to the previous quote

glowcloud:
The `if (value)` check is fine for query and OAS3, because of how we format queries later. At some point we use `formatKeyValue`, which in case of OAS3 will also go into `formatKeyValueBySerializationOption`, where here https://github.com/swagger-api/swagger-js/blob/e0c96b0831e410c4ef23dbf4ba42e33895a6cbd6/src/http/index.js#L331 it will format the params as `[ 'fields', [] ]`.

In case of OAS2, we will go in here https://github.com/swagger-api/swagger-js/blob/e0c96b0831e410c4ef23dbf4ba42e33895a6cbd6/src/http/index.js#L296 where the params will become `[ 'fields', '' ]`.

Because of this, the string that we get to add to the base URL in `encodeFormOrQuery`, the result of this line https://github.com/swagger-api/swagger-js/blob/e0c96b0831e410c4ef23dbf4ba42e33895a6cbd6/src/http/index.js#L419 in case of OAS3 will be just an empty string, while for OAS2 it will be `fields=`

PR #3380, opened 2024-02-21
fix(specmap): limit depth of traversal

glowcloud:
Refs https://github.com/swagger-api/swagger-ui/issues/9513

Fixes the issue by limiting the depth of traversal but leaves some of the references unresolved.


test/specmap/complex.js
```diff
@@ -49,4 +50,17 @@ describe('complex', () => {
...
+    // Given
+    const spec = globalThis.loadJsonFile(
+      path.join(__dirname, 'data', 'specs', 'complex-example.json')
+    );
+
+    // When
+    const result = await Swagger.resolve({ spec });
+
+    // Then
+    expect(result).not.toEqual(null);
```
glowcloud, line 64: Not sure if this is fine or if perhaps we should check some schema here.

glowcloud: Matching a snapshot here seems like a bad idea to me as the generated snapshot has a size over a 100 MB.

glowcloud: Perhaps we could do a check for `$ref` and `$$ref` to show that the spec is only partially resolved? And the test name should also be changed to `'should partially resolve...'`

```
expect(
      result.spec.components.schemas[
        'com.sap.ctsm.backend.core.api.study.v1.StudyAPIv1.StudyTreatments-create'
      ].properties.scenario.allOf[0].$ref
    ).toEqual(
      '#/components/schemas/com.sap.ctsm.backend.core.api.study.v1.StudyAPIv1.Scenarios-create'
    );
```

 ```
   expect(
      result.spec.components.schemas[
        'com.sap.ctsm.backend.core.api.study.v1.StudyAPIv1.BlindingGroups'
      ].properties.study.properties.scenarios.items.$$ref
    ).toEqual('#/components/schemas/com.sap.ctsm.backend.core.api.study.v1.StudyAPIv1.Scenarios');
```

char0n (approved):
LGTM. Nice work!

PR #3381, opened 2024-02-22
chore(deps): update unidici to v5.28.3

char0n:
Refs https://github.com/swagger-api/swagger-js/issues/3382


char0n:
undici@5.28.3 introduced [Nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) which is not supported in Node.js@12.

PR #3432, opened 2024-03-22
fix(execute): skip undefined values serialization

glowcloud:
Refs https://github.com/swagger-api/swagger-ui/issues/9550

### DIFF
--- a/src/execute/oas3/parameter-builders.js
+++ b/src/execute/oas3/parameter-builders.js
@@ -4,24 +4,26 @@ import serialize from './content-serializer.js';
 export function path({ req, value, parameter }) {
   const { name, style, explode, content } = parameter;
 
-  if (value !== undefined) {
-    if (content) {
-      const effectiveMediaType = Object.keys(content)[0];
+  if (value === undefined) {
+    return;
+  }
 
-      req.url = req.url
-        .split(`{${name}}`)
-        .join(encodeDisallowedCharacters(serialize(value, effectiveMediaType), { escape: true }));
-    } else {
-      const styledValue = stylize({
-        key: parameter.name,
-        value,
-        style: style || 'simple',
-        explode: explode || false,
-        escape: true,
-      });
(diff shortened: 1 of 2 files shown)


src/execute/oas3/parameter-builders.js
```diff
@@ -21,19 +21,23 @@ export function path({ req, value, parameter }) {
     escape: true,
   });
 
-  req.url = req.url.split(`{${name}}`).join(styledValue);
+  if (styledValue !== undefined) {
+    req.url = req.url.split(`{${name}}`).join(styledValue);
+  } else {
+    req.url = req.url.split(`{${name}}`).join('');
```
glowcloud, line 27: Added this as otherwise URLs for undefined parameters would have a comma: `http://localhost:8080/,`, which I noticed when testing the fix.

src/execute/oas3/parameter-builders.js
```diff
@@ -72,7 +76,7 @@ export function header({ req, parameter, value }) {
     return;
   }
 
-  if (parameter.content) {
+  if (value !== undefined && parameter.content) {
     const effectiveMediaType = Object.keys(parameter.content)[0];
 
     req.headers[parameter.name] = serialize(value, effectiveMediaType);
```
glowcloud, line 82: We could also add something like this
```
if (value === undefined) {
   return value
}
```
to the `serialize` function
https://github.com/swagger-api/swagger-js/blob/b1e84804ee835eab7fe5671b9beefff61f871608/src/execute/oas3/content-serializer.js#L6

and then we would have to do
```
const serializedValue = serialize(value, effectiveMediaType)

if (serializedValue !== undefined) {
   req.headers[parameter.name] = serializedValue
}
```

char0n (approved):
Looks legit. The flow logic is sound as well. Nice work!

As discussed, please refactor the path builder by unnesting the if statements and returning as easly as we can.

Issue #3438, opened 2024-03-26
OpenAPI 2.x: undefined body and path parameters are included in requests


glowcloud:
### Content & configuration

Swagger/OpenAPI definition:
```yaml
{
  host: 'swagger.io',
  basePath: '/v1',
  paths: {
    '/pets/find/{pathParam}': {
      get: {
        operationId: 'getMe',
        parameters: [
          {
            in: 'path',
            name: 'pathParam',
            type: 'string',
            required: false,
          },
          {
            in: 'body',
            name: 'bodyParam',
            type: 'string',
            required: false,
          },
        ],
        responses: {
          200: {
            description: 'ok',
          },
        },
      },
    },
  },
}
```

Swagger-Client usage:
```js
SwaggerClient.buildRequest({
  spec,
  operationId: 'getMe',
  parameters: { bodyParam: undefined, pathParam: undefined },
})
```

### Describe the bug you're encountering

Swagger Client adds `undefined` values to path and body parameters. 
```
{
  url: 'http://swagger.io/v1/pets/find/undefined',
  credentials: 'same-origin',
  headers: {},
  method: 'GET',
  body: undefined
}
```

### Expected behavior

Swagger Client should not add anything to the body when provided with `undefined` values and should not replace the parameter with `undefined` string in the path.
```
{
  url: 'http://swagger.io/v1/pets/find/{pathParam}',
  credentials: 'same-origin',
  headers: {},
  method: 'GET'
}
```

char0n:
Addressed in https://github.com/swagger-api/swagger-js/pull/3439

PR #3439, opened 2024-03-26
fix(execute): skip undefined parameters in body and path request builders for OpenAPI 2.0

glowcloud:
Refs #3438

### DIFF
--- a/src/execute/swagger2/parameter-builders.js
+++ b/src/execute/swagger2/parameter-builders.js
@@ -11,7 +11,9 @@ export default {
 
 // Add the body to the request
 function bodyBuilder({ req, value }) {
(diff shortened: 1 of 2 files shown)


char0n:
Note related to `bodyBuilder` - it seems that according to the `fetch API` spec, when `body` is either `undefined` or `""` empty string, it is not being sent at all. 

But given that we currently use 3 Fetch API implementations (browser Fetch API, Node.js Fetch API and node-fetch) it's probably better to remove the body instead of passing it with `undefined` value.

char0n (approved):
LGTM! Added one note: https://github.com/swagger-api/swagger-js/pull/3439#issuecomment-2019885415

PR #1235, opened 2018-01-29
fix deepObject serialization

championswimmer:
fix deepObject serialization

### Description

This will fix the problem described in might fix  https://github.com/swagger-api/swagger-ui/issues/4064 

### Motivation and Context

The deepObject style does not work as per described in teh spec

https://github.com/swagger-api/swagger-ui/issues/4064 

### Types of changes

- [ ] No code changes (changes to documentation, CI, metadata, etc)
- [ ] Dependency changes (any modification to dependencies in `package.json`)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #3446, opened 2024-04-03
fix(execute): fix encoding of complex nested type values in OpenAPI 3.x request builders

glowcloud:
Refs https://github.com/swagger-api/swagger-ui/issues/9755

From the test example, when encoding these parameters:
```
query: {
  a: {
    b: {
      c: 'd',
    },
  },
},
pathPartial: {
  foo: {
    bar: { baz: 'qux' },
  },
  a: {
    b: {
      c: 'd',
    },
  },
}
```
we will now get the following URL:
```
/foo=%7B%22bar%22%3A%7B%22baz%22%3A%22qux%22%7D%7D,a=%7B%22b%22%3A%7B%22c%22%3A%22d%22%7D%7D?a=%7B%22b%22%3A%7B%22c%22%3A%22d%22%7D%7D
```

which will decode to: 
```
/foo={"bar":{"baz":"qux"}},a={"b":{"c":"d"}}?a={"b":{"c":"d"}}
```

The cURL result for the 2.0 specification is:
```
curl -X 'GET' \
'https://petstore.swagger.io/v2/pet/findByStatus/%7B%22test%22%3A%20%7B%22test2%22%3A%20%22test3%22%20%7D%20%7D?queryParam=%7B%22test%22%3A%20%7B%22test2%22%3A%20%22test3%22%20%7D%20%7D' \
  -H 'accept: application/xml' \
  -H 'headerParam: {"test": {"test2": "test3" } }'
```
and the URL decodes to: 
```
https://petstore.swagger.io/v2/pet/findByStatus/{"test": {"test2": "test3" } }?queryParam={"test": {"test2": "test3" } }
```
so it seems to me like it doesn't need any changes.

### DIFF
--- a/src/execute/oas3/style-serializer.js
+++ b/src/execute/oas3/style-serializer.js
@@ -6,6 +6,11 @@ export function encodeDisallowedCharacters(str, { escape } = {}, parse) {
   if (typeof str === 'number') {
     str = str.toString();
   }
+
+  if (typeof str === 'object' && !Array.isArray(str)) {
+    str = JSON.stringify(str);
+  }
+
   if (typeof str !== 'string' || !str.length) {
     return str;
   }
--- a/src/execute/oas3/style-serializer.js
+++ b/src/execute/oas3/style-serializer.js
@@ -7,10 +7,6 @@ export function encodeDisallowedCharacters(str, { escape } = {}, parse) {
     str = str.toString();
   }
 
-  if (str !== null && typeof str === 'object' && !Array.isArray(str)) {
-    str = JSON.stringify(str);
-  }
-
   if (typeof str !== 'string' || !str.length) {
     return str;
   }
@@ -60,23 +56,29 @@ export default function stylize(config) {
   return encodePrimitive(config);
 }
 
-function encodeArray({ key, value, style, explode, escape }) {
-  const valueEncoder = (str) =>
-    encodeDisallowedCharacters(str, {
-      escape,
-    });
+function valueEncoder(value, escape) {
+  if (value !== null && typeof value === 'object') {
+    value = JSON.stringify(value);
+  } else if (typeof val === 'number') {
+    value = value.toString();
+  }
+  return encodeDisallowedCharacters(value, {
+    escape,
+  });
+}
 
+function encodeArray({ key, value, style, explode, escape }) {
   if (style === 'simple') {
-    return value.map((val) => valueEncoder(val)).join(',');
+    return value.map((val) => valueEncoder(val, escape)).join(',');
   }
 
   if (style === 'label') {
-    return `.${value.map((val) => valueEncoder(val)).join('.')}`;
+    return `.${value.map((val) => valueEncoder(val, escape)).join('.')}`;
   }
 
   if (style === 'matrix') {
     return value
-      .map((val) => valueEncoder(val))
+      .map((val) => valueEncoder(val, escape))
       .reduce((prev, curr) => {
         if (!prev || explode) {
           return `${prev || ''};${key}=${curr}`;
@@ -87,33 +89,28 @@ function encodeArray({ key, value, style, explode, escape }) {
 
   if (style === 'form') {
     const after = explode ? `&${key}=` : ',';
-    return value.map((val) => valueEncoder(val)).join(after);
+    return value.map((val) => valueEncoder(val, escape)).join(after);
   }
 
   if (style === 'spaceDelimited') {
     const after = explode ? `${key}=` : '';
-    return value.map((val) => valueEncoder(val)).join(` ${after}`);
+    return value.map((val) => valueEncoder(val, escape)).join(` ${after}`);
   }
 
   if (style === 'pipeDelimited') {
     const after = explode ? `${key}=` : '';
-    return value.map((val) => valueEncoder(val)).join(`|${after}`);
+    return value.map((val) => valueEncoder(val, escape)).join(`|${after}`);
   }
 
   return undefined;
 }
 
 function encodeObject({ key, value, style, explode, escape }) {
-  const valueEncoder = (str) =>
-    encodeDisallowedCharacters(str, {
-      escape,
-    });
-
   const valueKeys = Object.keys(value);
 
   if (style === 'simple') {
     return valueKeys.reduce((prev, curr) => {
-      const val = valueEncoder(value[curr]);
+      const val = valueEncoder(value[curr], escape);
       const middleChar = explode ? '=' : ',';
       const prefix = prev ? `${prev},` : '';
 
@@ -123,7 +120,7 @@ function encodeObject({ key, value, style, explode, escape }) {
 
   if (style === 'label') {
     return valueKeys.reduce((prev, curr) => {
-      const val = valueEncoder(value[curr]);
+      const val = valueEncoder(value[curr], escape);
       const middleChar = explode ? '=' : '.';
       const prefix = prev ? `${prev}.` : '.';
 
@@ -133,7 +130,7 @@ function encodeObject({ key, value, style, explode, escape }) {
 
   if (style === 'matrix' && explode) {
     return valueKeys.reduce((prev, curr) => {
-      const val = valueEncoder(value[curr]);
+      const val = valueEncoder(value[curr], escape);
       const prefix = prev ? `${prev};` : ';';
 
       return `${prefix}${curr}=${val}`;
@@ -143,7 +140,7 @@ function encodeObject({ key, value, style, explode, escape }) {
   if (style === 'matrix') {
     // no explode
     return valueKeys.reduce((prev, curr) => {
-      const val = valueEncoder(value[curr]);
+      const val = valueEncoder(value[curr], escape);
       const prefix = prev ? `${prev},` : `;${key}=`;
 
       return `${prefix}${curr},${val}`;
@@ -152,7 +149,7 @@ function encodeObject({ key, value, style, explode, escape }) {
 
   if (style === 'form') {
     return valueKeys.reduce((prev, curr) => {
-      const val = valueEncoder(value[curr]);
+      const val = valueEncoder(value[curr], escape);
       const prefix = prev ? `${prev}${explode ? '&' : ','}` : '';
       const separator = explode ? '=' : ',';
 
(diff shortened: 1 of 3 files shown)


src/execute/oas3/style-serializer.js
```diff
@@ -6,6 +6,11 @@ export function encodeDisallowedCharacters(str, { escape } = {}, parse) {
   if (typeof str === 'number') {
     str = str.toString();
   }
+
+  if (typeof str === 'object' && !Array.isArray(str)) {
```
char0n, line 10: If you really want to just support POJOs here (Plain Old JavaScript Objects), and not `null`, you have to do the followig.

```suggestion
  if (str !== null && typeof str === 'object' && !Array.isArray(str)) {
```

test/oas3/execute/main.js
```diff
@@ -524,6 +524,97 @@ describe('buildRequest - OpenAPI Specification 3.0', () => {
         method: 'GET',
       });
     });
+
+    it('should encode JSON values provided as objects', () => {
```
char0n, line 528: Can we create this test just on top of https://github.com/swagger-api/swagger-js/blob/e943f35aa94745f9b7e89bec5a029a89e1789951/test/oas3/execute/main.js#L666 like this?

```js
  describe('`schema` parameters', () => {
    it('should serialize JSON values provided as objects', () => {
....
```

char0n:
One additional question: how is it possible that it worked for Parameter Objects with `content` field, but didn't work for ones with `schema` field? Do we know?

This test guarantees that it works for Parameter Objects with `content` field: https://github.com/swagger-api/swagger-js/blob/e943f35aa94745f9b7e89bec5a029a89e1789951/test/oas3/execute/main.js#L666

glowcloud:
> One additional question: how is it possible that it worked for Parameter Objects with `content` field, but didn't work for ones with `schema` field? Do we know?
> 
> This test guarantees that it works for Parameter Objects with `content` field:
> 
> https://github.com/swagger-api/swagger-js/blob/e943f35aa94745f9b7e89bec5a029a89e1789951/test/oas3/execute/main.js#L666

In this case, we use the `serialize` function
https://github.com/swagger-api/swagger-js/blob/12f249aee1a372bae9de0d69c00b50712835ae61/src/execute/oas3/parameter-builders.js#L31-L33
which stringifies the value 
https://github.com/swagger-api/swagger-js/blob/12f249aee1a372bae9de0d69c00b50712835ae61/src/execute/oas3/content-serializer.js#L6-L16

glowcloud:
With our current findings, we decided not to apply the styling properties from OpenAPI specification to the nested objects. They will be processed by using `JSON.stringify` function.

We should do that before using `encodeDisallowedCharacters` for objects and arrays in 
https://github.com/swagger-api/swagger-js/blob/12f249aee1a372bae9de0d69c00b50712835ae61/src/execute/oas3/style-serializer.js#L59-L62
and accordingly in the other `valueEncoder` functions, and in https://github.com/swagger-api/swagger-js/blob/12f249aee1a372bae9de0d69c00b50712835ae61/src/http/index.js#L322

`encodeDisallowedCharacters` should also be refactored (in a separate issue/PR) so that it actually only encodes given strings, instead of having additional logic like this:
https://github.com/swagger-api/swagger-js/blob/12f249aee1a372bae9de0d69c00b50712835ae61/src/execute/oas3/style-serializer.js#L6-L8

We should handle these cases in the `valueEncoder` functions and `encodeDisallowedCharacters`  should only be getting strings.

It also looks like the `parse` parameter is not set anywhere when using this function, so this logic might be removed with the refactor:
https://github.com/swagger-api/swagger-js/blob/12f249aee1a372bae9de0d69c00b50712835ae61/src/execute/oas3/style-serializer.js#L17-L19

There's also the issue of using the `escape` parameter where sometimes it's a `string` and sometimes a `boolean` - we need to research its usage and whether it should stay in its current form.

src/execute/oas3/style-serializer.js
```diff
@@ -159,29 +161,24 @@ function encodeObject({ key, value, style, explode, escape }) {
...
   }
 
   if (style === 'form') {
-    return valueEncoder(value);
+    return valueEncoder(value, escape);
   }
 
   if (style === 'deepObject') {
-    return valueEncoder(value, {}, true);
+    return valueEncoder(value, escape);
```
glowcloud, line 181: While refactoring `valueEncoder` as a separate function, I noticed that what was there before 
```
return valueEncoder(value, {}, true);
``` 
shouldn't actually pass these additional parameters to the function, as there's only one defined. It looks to me like `{}` was supposed to be a falsy `encode` parameter and `true` was to be the `parse` parameter in `encodeDisallowedCharacters`.

I found the PR for this: https://github.com/swagger-api/swagger-js/pull/1235/files and it looks to me like these values were never actually used.

Looking further into this.

glowcloud:
For the parameters in the new test:
```
arrayOfObjects: [
  {
    a: {
      b: 'c',
    },
  },
  {
    d: {
      e: 'f',
    },
  },
],
arrayOfArrays: [
  [
    {
      a: {
        b: 'c',
      },
    },
  ],
]
```

we will get the following URL:
```
/?arrayOfObjects=%7B%22a%22%3A%7B%22b%22%3A%22c%22%7D%7D,%7B%22d%22%3A%7B%22e%22%3A%22f%22%7D%7D&arrayOfArrays=%5B%7B%22a%22%3A%7B%22b%22%3A%22c%22%7D%7D%5D
```
which should decode to:
```
/?arrayOfObjects={"a":{"b":"c"}},{"d":{"e":"f"}}&arrayOfArrays=[{"a":{"b":"c"}}]
```

glowcloud:
More on `deepObject` style for query parameters:

For this spec:
```
{
  openapi: '3.0.1',
  info: {
    title: 'Test',
    version: 'v0',
  },
  servers: [
    {
      url: 'http://localhost:8080',
      description: 'Test',
    },
  ],
  paths: {
    '/': {
      post: {
        operationId: 'params',
        parameters: [
          {
            name: 'queryParam',
            in: 'query',
            schema: {
              type: 'object',
            },
            style: 'deepObject',
            explode: true,
          },
        ],
      },
    },
  },
}
```

with these parameters:
```
{
  queryParam: {
    first: {
      firstKey: 'firstValue',
    },
    second: {
      secondKey: 'secondValue',
    },
    third: {
      thirdKey: 'thirdValue',
    },
  },
}
```
With the current changes, the parameters will be encoded as:
```
http://localhost:8080/?queryParam%5Bfirst%5D=%7B%22firstKey%22%3A%22firstValue%22%7D&queryParam%5Bsecond%5D=%7B%22secondKey%22%3A%22secondValue%22%7D&queryParam%5Bthird%5D=%7B%22thirdKey%22%3A%22thirdValue%22%7D
```
which should decode to:
```
http://localhost:8080/?queryParam[first]={"firstKey":"firstValue"}&queryParam[second]={"secondKey":"secondValue"}&queryParam[third]={"thirdKey":"thirdValue"}
```
so only the first nested key/value pair would be in the `deepObject` style.

Previously we would have this url:
```
http://localhost:8080/?queryParam%5Bfirst%5D[firstKey]=firstValue&queryParam%5Bsecond%5D[secondKey]=secondValue&queryParam%5Bthird%5D[thirdKey]=thirdValue
```
which would be decoded to:
```
http://localhost:8080/?queryParam[first][firstKey]=firstValue&queryParam[second][secondKey]=secondValue&queryParam[third][thirdKey]=thirdValue
```

so we had the style but not the encoding for nested values.

I did further research and the URL for nested object parameters if we didn't use `deepObject` style would be:
```
http://localhost:8080/?first[firstKey]=firstValue&second[secondKey]=secondValue&third[thirdKey]=thirdValue
```
In case of an array of parameters:
```
http://localhost:8080/?queryParam[firstKey]=firstValue&queryParam[secondKey]=secondValue&queryParam[thirdKey]=thirdValue
```

so there wasn't an issue with getting `[object Object]` in query but we didn't do any encoding.

As it turns out, it's because we're using the [qs library](https://www.npmjs.com/package/qs) here
https://github.com/swagger-api/swagger-js/blob/8128a0903f9cb397fa1581bd6a6ef6c176abb9cb/src/http/index.js#L419
which has support for the `deepObject` style:

Screenshot 2024-04-04 at 16 19 23

it can even do its own encoding, which we set to `false`.

With these in mind, I'm wondering if we should stay with defaulting to `deepObject` for nested parameters in case of query or if we should align the behaviour with header, cookie and path where we stringify nested objects and arrays (also works like this for query in the current PR). 
If we would want to stay with the `deepObject` behaviour, perhaps we could use the encoding option from `qs` so that the whole URL would be encoded. It might also be a general idea in case of specifying the `deepObject` style.

src/execute/oas3/style-serializer.js
```diff
@@ -55,23 +56,29 @@ export default function stylize(config) {
   return encodePrimitive(config);
 }
 
-function encodeArray({ key, value, style, explode, escape }) {
-  const valueEncoder = (str) =>
-    encodeDisallowedCharacters(str, {
-      escape,
-    });
+function valueEncoder(value, escape) {
+  if (value !== null && typeof value === 'object') {
```
char0n, line 60: NOTE: extend the explicit assertion to the array

src/http/index.js
```diff
@@ -319,8 +319,19 @@ function formatKeyValueBySerializationOption(key, value, skipEncoding, serializa
     : serializationOption && serializationOption.allowReserved
       ? 'unsafe'
       : 'reserved';
-  const encodeFn = (v) => encodeDisallowedCharacters(v, { escape });
-  const encodeKeyFn = skipEncoding ? (k) => k : (k) => encodeDisallowedCharacters(k, { escape });
+
+  const encodeFn = (v) => {
```
char0n, line 323: We use exported from https://github.com/swagger-api/swagger-js/pull/3446/files#diff-5f9165415636dea264548b1e96c83949a93c5a3b6ae9af3951cfb1fe63443c61R59

src/execute/oas3/style-serializer.js
```diff
@@ -55,23 +56,29 @@ export default function stylize(config) {
...
+    value = JSON.stringify(value);
+  } else if (typeof val === 'number') {
```
char0n, line 62: For now let's not do the number encoding here.

Issue #3462, opened 2024-04-08
Rework encodeDisallowedCharacters function for OpenAPI 3.x request builders


glowcloud:
`encodeDisallowedCharacters` function 
https://github.com/swagger-api/swagger-js/blob/f327d4bd28aa5d3ed646b412f8ec9422fe55441a/src/execute/oas3/style-serializer.js#L5
 contains logic that should be moved out of it - the function should only encode characters.
We can move the logic for converting numbers to strings and returning empty strings or non-string type values to the `valueEncoder` function. It should take care of converting all values to strings as it does now for objects and arrays.
https://github.com/swagger-api/swagger-js/blob/f327d4bd28aa5d3ed646b412f8ec9422fe55441a/src/execute/oas3/style-serializer.js#L59

We can also remove the `parse` parameter as it is not used anywhere. 

The signature of the function should be `encodeCharacters(str, { characterSet = 'reserved' } = {}` and the logic for returning non-encoded strings for when the `escape` value is set to `false` should also be moved to `valueEncoder`.

char0n:
Addressed in https://github.com/swagger-api/swagger-js/pull/3463

PR #3463, opened 2024-04-08
fix(execute): refactor encodeDisallowedCharacters function

glowcloud:
Refs #3462

### DIFF
--- a/src/execute/oas3/style-serializer.js
+++ b/src/execute/oas3/style-serializer.js
@@ -1,35 +1,19 @@
 const isRfc3986Reserved = (char) => ":/?#[]@!$&'()*+,;=".indexOf(char) > -1;
-const isRrc3986Unreserved = (char) => /^[a-z0-9\-._~]+$/i.test(char);
+const isRfc3986Unreserved = (char) => /^[a-z0-9\-._~]+$/i.test(char);
 
 // eslint-disable-next-line default-param-last
-export function encodeDisallowedCharacters(str, { escape } = {}, parse) {
-  if (typeof str === 'number') {
-    str = str.toString();
-  }
-
-  if (typeof str !== 'string' || !str.length) {
-    return str;
-  }
-
-  if (!escape) {
-    return str;
-  }
-
-  if (parse) {
-    return JSON.parse(str);
-  }
-
+export function encodeCharacters(str, { characterSet = 'reserved' } = {}) {
   // In ES6 you can do this quite easily by using the new ... spread operator.
   // This causes the string iterator (another new ES6 feature) to be used internally,
   // and because that iterator is designed to deal with
   // code points rather than UCS-2/UTF-16 code units.
   return [...str]
     .map((char) => {
-      if (isRrc3986Unreserved(char)) {
+      if (isRfc3986Unreserved(char)) {
         return char;
       }
 
-      if (isRfc3986Reserved(char) && escape === 'unsafe') {
+      if (isRfc3986Reserved(char) && characterSet === 'unsafe') {
         return char;
       }
 
@@ -59,10 +43,14 @@ export default function stylize(config) {
 export function valueEncoder(value, escape) {
   if (Array.isArray(value) || (value !== null && typeof value === 'object')) {
     value = JSON.stringify(value);
+  } else if (typeof value === 'number' || typeof value === 'boolean') {
+    value = value.toString();
   }
-  return encodeDisallowedCharacters(value, {
-    escape,
-  });
+
+  if (escape && value.length > 0) {
+    return encodeCharacters(value, escape === true ? {} : { characterSet: escape });
+  }
+  return value;
 }
 
 function encodeArray({ key, value, style, explode, escape }) {
--- a/src/execute/oas3/style-serializer.js
+++ b/src/execute/oas3/style-serializer.js
@@ -2,7 +2,7 @@ const isRfc3986Reserved = (char) => ":/?#[]@!$&'()*+,;=".indexOf(char) > -1;
 const isRfc3986Unreserved = (char) => /^[a-z0-9\-._~]+$/i.test(char);
 
 // eslint-disable-next-line default-param-last
-export function encodeCharacters(str, { characterSet = 'reserved' } = {}) {
+export function encodeCharacters(str, characterSet = 'reserved') {
   // In ES6 you can do this quite easily by using the new ... spread operator.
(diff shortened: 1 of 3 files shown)


src/execute/oas3/style-serializer.js
```diff
@@ -59,10 +43,14 @@ export default function stylize(config) {
 export function valueEncoder(value, escape) {
   if (Array.isArray(value) || (value !== null && typeof value === 'object')) {
     value = JSON.stringify(value);
+  } else if (typeof value === 'number' || typeof value === 'boolean') {
```
glowcloud, line 46: Added a check for `boolean` type values as I've noticed they were not being converted for headers if just using Swagger Client - the values should already be stringified when getting them from Swagger UI. For header these values will be returned without any encoding, so now the result for the `buildRequest` function will be something like:
```
{
  url: 'http://localhost:8080/',
  credentials: 'same-origin',
  headers: { headerNum: '1', headerBool: 'false' },
  method: 'POST'
}
```
and before this change it was:
```
{
  url: 'http://localhost:8080/',
  credentials: 'same-origin',
  headers: { headerNum: '1', headerBool: false },
  method: 'POST'
}
```

src/execute/oas3/style-serializer.js
```diff
@@ -1,35 +1,19 @@
...
-
-  if (!escape) {
-    return str;
-  }
-
-  if (parse) {
-    return JSON.parse(str);
-  }
-
+export function encodeCharacters(str, { characterSet = 'reserved' } = {}) {
```
glowcloud, line 5: I'm wondering if we could have the signature for this as just 
```
encodeCharacters(str, characterSet = 'reserved')
```

src/execute/oas3/style-serializer.js
char0n: Yep, as discussed last week, `number` and `boolean` types along with `null` were affected. So the check is warranted.

src/execute/oas3/style-serializer.js
```diff
@@ -59,10 +43,14 @@ export default function stylize(config) {
 export function valueEncoder(value, escape) {
   if (Array.isArray(value) || (value !== null && typeof value === 'object')) {
     value = JSON.stringify(value);
+  } else if (typeof value === 'number' || typeof value === 'boolean') {
+    value = value.toString();
```
char0n, line 47: I'd suggest to use this notation instead of `toString()`, as `isString()` is called by `String` function as of of the mechanism of coercion (not the only one).
```suggestion
    value = String(value);
```

test/oas3/execute/style-serializer.js
```diff
@@ -1,9 +1,9 @@
-import { encodeDisallowedCharacters } from '../../../src/execute/oas3/style-serializer.js';
+import { valueEncoder } from '../../../src/execute/oas3/style-serializer.js';
 
 describe('OAS3 style serializer', () => {
```
char0n, line 3: Why converting this test to testing `valueEncoder` instead of `encodeCharacters`? This test is actually testing only the `encodeCharacters` as it assumes that the input is always a string. If you want to test `valueEncoder` I'd create a separate test suite that tests it's aspects of coercion.

src/execute/oas3/style-serializer.js
```diff
@@ -59,10 +43,14 @@ export default function stylize(config) {
 export function valueEncoder(value, escape) {
```
char0n, line 43: If we look at the current code of value encoder:

```js
export function valueEncoder(value, escape) {
  if (Array.isArray(value) || (value !== null && typeof value === 'object')) {
    value = JSON.stringify(value);
  } else if (typeof value === 'number' || typeof value === 'boolean') {
    value = value.toString();
  }

  if (escape && value.length > 0) {
    return encodeCharacters(value, escape === true ? {} : { characterSet: escape });
  }
  return value;
}
```

I can conclude multiple observations:

### 1. the valueEncoder doesn't encode characters unless explicitly requested

I would suggest to change the signature to 

```js
export function valueEncoder(value, escape = false) {
```

### 2. value encoder assumes that the `escape` can be `true`

We've established that the allowed values for the `escape` parameter can be: `false`, `reserved` and `unreserved`

```js
return encodeCharacters(value, escape === true ? {} : { characterSet: escape });
```

So the above code passes `true` to the `encodeCharacters`, which is wrong. It should just omit the `characterSet` option so that default characterSet will kick in. 

Still valueEncoder itself should not receive `true` and `escape` value.

src/execute/oas3/style-serializer.js
char0n: We would. I guess the original author of the function expected that the second argument the be an `Options` objects and that it would eventually support more options. But as there is only one option currently  and given the function is internal, you can (if you wish) transform the signature to standard default parameter.

test/oas3/execute/style-serializer.js
glowcloud: This was done because `characterSet` was supposed to be only `unsafe` or `reserved` in `encodeCharacters` while the `escape` parameter should be handled by `valueEncoder`. Before in this test we were passing a `null` value as escape - the current `characterSet` would be the default `reserved` after the changes and the result would be encoded. We only return unencoded characters if `escape` is `false` in the `valueEncoder`. Before, this line took care of this case:
https://github.com/swagger-api/swagger-js/blob/f327d4bd28aa5d3ed646b412f8ec9422fe55441a/src/execute/oas3/style-serializer.js#L14-L16

test/oas3/execute/style-serializer.js
```diff
@@ -37,9 +39,105 @@ describe('OAS3 style serializer', () => {
...
+      expect(tested({ a: { b: { c: 'd' } } })).toEqual(escape('{"a":{"b":{"c":"d"}}}'));
+      expect(tested({ a: 123 })).toEqual(escape('{"a":123}'));
+      expect(tested({ a: false })).toEqual(escape('{"a":false}'));
+    });
+
+    test('should correctly encode objects with escape set to `unsafe`', () => {
+      const tested = (value) => valueEncoder(value, 'unsafe');
+
+      expect(tested({ a: 'a#b$c%d[e]_~1' })).toEqual(
+        `${escape('{"a"')}:${escape('"')}a#b$c${escape('%')}d[e]_~1${escape('"}')}`
```
char0n, line 74: Is it possible to rewrite this to something more sensible? I cannot really read this or tell if this is right or not (as a human).

PR #3476, opened 2024-04-22
fix(execute): fix encoding for stringified request body

glowcloud:
Refs https://github.com/swagger-api/swagger-ui/issues/7810

With these changes we should now parse the string and then iterate through it only if it's actually an object.

The issue is that we are sending a stringified object and then iterating through it as if it were just an object. It happened even if I updated the examples in the specification provided in the issue, which seemed wronged to me,  to:
```yaml
examples:
  authorization_code grant type:
    description: authorization_code grant type
    value:
      grant_type: authorization_code
      code: xxxxxxxxxx
      client_id: xxxxxxxxxx
  client_credentials grant type:
    description: client_credentials grant type
    value: 
      grant_type: client_credentials
      client_id: xxxxxxxxxx
      client_secret: xxxxxxxxxx
  refresh_token grant type:
    description: refresh_token grant type
    value: 
      grant_type: refresh_token
      refresh_token: xxxxxxxxxx
```
so that it shows as an object in Swagger UI:

Screenshot 2024-04-22 at 14 38 52

### DIFF
--- a/src/http/index.js
+++ b/src/http/index.js
@@ -405,6 +405,19 @@ export function encodeFormOrQuery(data) {
    * @param {string} parameterName - Parameter name
    * @return {object} encoded parameter names and values
    */
+  if (typeof data === 'string') {
+    try {
+      data = JSON.parse(data);
+      Object.entries(data).forEach(([key, value]) => {
+        if (typeof value === 'object' && !Array.isArray(value)) {
+          data[key] = JSON.stringify(value);
+        }
+      });
+    } catch {
+      return valueEncoder(data, 'reserved');
+    }
+  }
+
   const encodedQuery = Object.keys(data).reduce((result, parameterName) => {
     // eslint-disable-next-line no-restricted-syntax
     for (const [key, value] of formatKeyValue(parameterName, data[parameterName])) {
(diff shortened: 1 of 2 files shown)


src/http/index.js
```diff
@@ -405,6 +405,19 @@ export function encodeFormOrQuery(data) {
    * @param {string} parameterName - Parameter name
    * @return {object} encoded parameter names and values
    */
+  if (typeof data === 'string') {
+    try {
+      data = JSON.parse(data);
+      Object.entries(data).forEach(([key, value]) => {
+        if (typeof value === 'object' && !Array.isArray(value)) {
+          data[key] = JSON.stringify(value);
```
glowcloud, line 413: Stringifying the nested value as otherwise it would be seen as undefined here: https://github.com/swagger-api/swagger-js/blob/ff9631fa905ecfd9fb7a92dad5c18455273404cd/src/http/index.js#L226
And if we did 
```js
data[key] = { value }
```
the value would be empty because of this: https://github.com/swagger-api/swagger-js/blob/ff9631fa905ecfd9fb7a92dad5c18455273404cd/src/http/index.js#L307

I think we could also use
```js
data[key] = { value, encoding: {} }
```
so that it's treated as an OpenAPI 3.x object but in the end it would just be stringified anyway:
https://github.com/swagger-api/swagger-js/blob/ff9631fa905ecfd9fb7a92dad5c18455273404cd/src/http/index.js#L285-L286

Issue #3481, opened 2024-04-23
Consolidate code for HTTP client


glowcloud:
We need to consolidate the code contained inside https://github.com/swagger-api/swagger-js/blob/master/src/http/index.js.

Some parts of it seem like they could be moved to an outside module but first we need to analyse it and understand how it actually works. The refactor will help us with maintaining this code and make it easier to understand in the future.

glowcloud:
Addressed in https://github.com/swagger-api/swagger-js/pull/3482

PR #3482, opened 2024-04-23
refactor(http): refactor http client

glowcloud:
Refs #3481

1. Moved request and response serializing/encoding into separate files 
2. Moved the logic introduced by https://github.com/swagger-api/swagger-js/pull/3476 and https://github.com/swagger-api/swagger-js/pull/3474 to request builder

### DIFF
--- a/src/http/index.js
+++ b/src/http/index.js
@@ -1,14 +1,10 @@
-import qs from 'qs';
-import jsYaml from 'js-yaml';
-
 import '../helpers/fetch-polyfill.node.js';
-import { valueEncoder } from '../execute/oas3/style-serializer.js';
-
-// For testing
-export const self = {
+import { mergeInQueryOrForm, encodeFormOrQuery, isFile } from './serializers/request/index.js';
+import {
   serializeRes,
-  mergeInQueryOrForm,
-};
+  serializeHeaders,
+  shouldDownloadAsText,
+} from './serializers/response/index.js';
 
 // Handles fetch-like syntax and the case where there is only one object passed-in
 // (which will have the URL as a property). Also serializes the response.
@@ -23,7 +19,7 @@ export default async function http(url, request = {}) {
   // Serializes query, for convenience
   // Should be the last thing we do, as its hard to mutate the URL with
   // the search string, but much easier to manipulate the req.query object
-  self.mergeInQueryOrForm(request);
+  mergeInQueryOrForm(request);
 
   // Newlines in header values cause weird error messages from `window.fetch`,
   // so let's message them out.
@@ -59,7 +55,7 @@ export default async function http(url, request = {}) {
   let res;
   try {
     res = await (request.userFetch || fetch)(request.url, request);
-    res = await self.serializeRes(res, url, request);
+    res = await serializeRes(res, url, request);
     if (request.responseInterceptor) {
       res = (await request.responseInterceptor(res)) || res;
     }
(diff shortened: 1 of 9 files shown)


glowcloud:
1. For placement of response and request serializers - `mergeInQueryOrForm` is also used in one place in `execute`, but it's also for building requests, so it still seemed related to `http`
2. I also thought about leaving index in http just for exports - let me know if that would be better and the functions should be moved to an additonal file, ex. `http.js`
3. `file` and `format` functions are only used for request serializing/encoding and in tests, so also looks like it's an appropriate place for them to be

src/http/index.js
```diff
@@ -85,418 +81,26 @@ export default async function http(url, request = {}) {
...
       req = { url: req };
     }
-    self.mergeInQueryOrForm(req);
+    mergeInQueryOrForm(req);
     req = preFetch(req);
     return postFetch(httpFn(req));
   };
 }
+
+export {
```
char0n, line 98: Let's make the exports consistent. Somethings we do this https://github.com/swagger-api/swagger-js/pull/3482/files#diff-26de8ea9a0c9cc66754ca256b3e78709e038552b18547dfb30e99eb6a9143c1cR25  and sometimes we export all symbols using `export { }`.

`export { }` construct is used conventionally for re-exporting symbols.

src/http/index.js
```diff
@@ -85,418 +81,26 @@ export default async function http(url, request = {}) {
...
+  makeHttp,
+  mergeInQueryOrForm,
+  encodeFormOrQuery,
+  isFile,
```
char0n, line 102: We should only export symbols here that actually form the public API of this module. I'd say that only `http` (default export) and `makeHttp` for the publi API of this module.

Other parts of the codebase can look into `http/serializers/request/...`

src/http/serializers/request/index.js
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    const finalStr = joinSearch(newStr, encodeFormOrQuery(query));
+    req.url = baseUrl + finalStr;
+    delete req.query;
+  }
+  return req;
+}
+
+export { mergeInQueryOrForm, encodeFormOrQuery, isFile };
```
char0n, line 112: Do all the symbols need to be exported. Generally, let's address all the exports in this PR and make sure that we only exports symbols that are used elsewhere.

src/http/serializers/request/index.js
```diff
@@ -0,0 +1,112 @@
...
+    }
+    return result;
+  }, {});
+
+  return qs.stringify(encodedQuery, { encode: false, indices: false }) || '';
+}
+
+// If the request has a `query` object, merge it into the request.url, and delete the object
+// If file and/or multipart, also create FormData instance
+function mergeInQueryOrForm(req = {}) {
```
char0n, line 68: Let's give a proper naming consistent with https://github.com/swagger-api/swagger-js/pull/3482/files#r1576329286. The fact that it's serializing query or form is just an implementation detail, right?

```suggestion
function serializeRequest(req = {}) {
```

char0n (changes requested):
Looks cleaner and more maintainable for sure. Nicely done!

Please have a look at my CRCs.

src/http/index.js
```diff
@@ -23,7 +19,7 @@ export default async function http(url, request = {}) {
   // Serializes query, for convenience
   // Should be the last thing we do, as its hard to mutate the URL with
   // the search string, but much easier to manipulate the req.query object
-  self.mergeInQueryOrForm(request);
+  mergeInQueryOrForm(request);
```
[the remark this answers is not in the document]
char0n: the function is returning the modified request, let's use that fact.

src/http/serializers/request/index.js
glowcloud: We have some of these exports because we test them.

src/http/index.js
```diff
@@ -1,14 +1,8 @@
...
-import { valueEncoder } from '../execute/oas3/style-serializer.js';
+import { serializeRequest } from './serializers/request/index.js';
+import { serializeResponse, serializeHeaders } from './serializers/response/index.js';
 
-// For testing
-export const self = {
-  serializeRes,
-  mergeInQueryOrForm,
-};
+export { serializeResponse as serializeRes, serializeHeaders };
```
glowcloud, line 5: Left exporting these here as we do expose them:
https://github.com/swagger-api/swagger-js/blob/df80732d2a1c4f2e82bc3a10e8fd8cb1ffc7be64/src/index.js#L3
https://github.com/swagger-api/swagger-js/blob/df80732d2a1c4f2e82bc3a10e8fd8cb1ffc7be64/src/index.js#L46-L47
and serializeRes is also imported in Swagger UI:
https://github.com/swagger-api/swagger-ui/blob/f7373a0bc11268e266eca2173bb143d0eb263a32/src/core/plugins/swagger-client/index.js#L7
https://github.com/swagger-api/swagger-ui/blob/f7373a0bc11268e266eca2173bb143d0eb263a32/src/core/plugins/swagger-client/index.js#L43

although I don't see any other usage for it there.

char0n: Let's not export `serializeHeaders` from here. If needed by other code, it should import from `http/serializers/response/**.js`.

char0n (approved):
Let's process https://github.com/swagger-api/swagger-js/pull/3482/files#r1577376938 and "squash + merge"

src/http/index.js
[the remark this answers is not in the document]
char0n: this new form allow to handle both mutating and immutable  cases

Issue #3508, opened 2024-05-09
[3.27.4 Regression] Multiple path parameters are not all applied


kevinoid:
### Q&A (please complete the following information)
 - OS: Linux
 - Environment: Node.js v18.19.1
 - Method of installation: npm
 - Swagger-Client version: 3.27.7
 - Swagger/OpenAPI version: Swagger 2.0 or OpenAPI 3.0

### Content & configuration

Swagger/OpenAPI definition:

<details>
<summary>OpenAPI 3.0 example (saved as <code>openapi.yaml</code>)</summary>

```yaml
openapi: 3.0.1
info:
  title: Example OpenAPI 3.0 spec with multiple path parameters
  version: 1.0.0
servers:
- url: https://example.com/
paths:
  /{owner}/{pet}/age:
    get:
      description: Get the age of a given pet for a given owner.
      operationId: getAge
      parameters:
      - name: owner
        in: path
        required: true
        schema:
          type: string
      - name: pet
        in: path
        required: true
        schema:
          type: string
      responses:
        "200":
          description: "Age of the pet, in years"
          content:
            application/json:
              schema:
                type: integer
```
</details>

<details>
<summary>Swagger 2.0 example (saved as <code>swagger.yaml</code>)</summary>

```yaml
swagger: "2.0"
info:
  version: 1.0.0
  title: Example Swagger 2.0 spec with multiple path parameters
host: example.com
produces:
- application/json
consumes:
- application/json
paths:
  /{owner}/{pet}/age:
    parameters:
    - name: owner
      in: path
      type: string
      required: true
    - name: pet
      in: path
      type: string
      required: true
    get:
      description: Get the age of a given pet for a given owner.
      responses:
        "200":
          description: Age of the pet, in years
          schema:
            type: integer
```
</details>

### Describe the bug you're encountering

When an operation contains multiple path parameters, SwaggerClient makes a request to a URL where only one of the path parameters is applied and the other parameter names remain unchanged from the URL template in the request URL.

### To reproduce...

Run the following code
```js
import SwaggerClient from 'swagger-client';

function testFetch(url) {
  console.log(`fetch('${url}')`);
  process.exit(url === 'https://example.com/joe/fido/age' ? 0 : 1);
}

const client = await new SwaggerClient({
  spec: {
    swagger: '2.0',
    info: {
      version: '1.0.0',
      title: 'Example Swagger 2.0 spec with multiple path parameters',
    },
    schemes: ['https'],
[...]

PR #3511, opened 2024-05-10
fix(execute): resolve multiple path parameters with the same name in path templates

glowcloud:
Refs #3508

### DIFF
--- a/src/execute/oas3/parameter-builders.js
+++ b/src/execute/oas3/parameter-builders.js
@@ -3,24 +3,25 @@ import { resolve as resolvePathTemplate } from 'openapi-path-templating';
 import stylize, { encodeCharacters } from './style-serializer.js';
 import serialize from './content-serializer.js';
 
-export function path({ req, value, parameter, pathName }) {
+export function path({ req, value, parameter, baseURL }) {
   const { name, style, explode, content } = parameter;
 
   if (value === undefined) return;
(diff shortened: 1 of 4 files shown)


src/execute/oas3/parameter-builders.js
```diff
@@ -3,24 +3,25 @@ import { resolve as resolvePathTemplate } from 'openapi-path-templating';
 import stylize, { encodeCharacters } from './style-serializer.js';
 import serialize from './content-serializer.js';
 
-export function path({ req, value, parameter, pathName }) {
+export function path({ req, value, parameter, baseURL }) {
   const { name, style, explode, content } = parameter;
 
   if (value === undefined) return;
 
+  const pathname = req.url.replace(baseURL, '');
```
char0n, line 11: Let's make it safe by really striping the baseURL only from the begging of the `req.url`.

```suggestion
  const pathname = req.url.replace(new RegExp(`^${escapeRegExp(baseURL)}`), '');
```

char0n (changes requested):

src/execute/oas3/parameter-builders.js
char0n: Never mind, it's equivalent with `string` replace.

char0n (approved):

Issue #3520, opened 2024-05-17
OpenAPI 3.1: visitors after the dereference visitor don't get mutated elements


glowcloud:
### Content & configuration

Swagger/OpenAPI definition:
```yaml
openapi: 3.1.0
info:
  title: Example generation test
  version: 1.0.0
paths:
  /foo:
    get:
      responses:
        '200':
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Foo'
components:
  schemas:
    Foo:
      type: object
      properties:
        bar:
          allOf:
            - type: object
              properties:
                a:
                  type: integer
```

### Describe the bug you're encountering

The `allOf` properties are not merged into the parent schema if the schema was referenced. The dereferenced schema in the response will still have the `allOf` keyword:

<img width="297" alt="Screenshot 2024-05-16 at 10 01 44" src="https://github.com/swagger-api/swagger-js/assets/44094416/07651f63-9517-4bbf-b8e0-1e74fc9d7c7d">

while the schema itself will be merged:

<img width="221" alt="Screenshot 2024-05-16 at 09 59 45" src="https://github.com/swagger-api/swagger-js/assets/44094416/a34bec39-e9cb-4407-894a-944dd07af9b6">

The issue here is that the parent elements get mutated by the `DereferenceVisitor`:
https://github.com/swagger-api/swagger-js/blob/fa77bfd689e070c2e1fe7f6f58a7cd5b9368ebea/src/resolver/apidom/reference/dereference/strategies/openapi-3-1-swagger-client/visitors/dereference.js#L776-L780
and the subsequent visitors do not see this change. 

We get the correct result if instead of merging the `AllOfVisitor` with the `DereferenceVisitor` and visiting here:
https://github.com/swagger-api/swagger-js/blob/fa77bfd689e070c2e1fe7f6f58a7cd5b9368ebea/src/resolver/apidom/reference/dereference/strategies/openapi-3-1-swagger-client/index.js#L107-L118

we did a second merge and visit with only the `AllOfVisitor` on the dereferenced element:

```js
  if (this.mode !== 'strict') {
    const allOfVisitor = AllOfVisitor({ options });
    secondVisitors.push(allOfVisitor);
   }

   // establish root visitor by visitor merging
   const rootVisitor = mergeAllVisitorsAsync(visitors, { nodeTypeGetter: getNodeType });

   let dereferencedElement = await visitAsync(refSet.rootRef.value, rootVisitor, {
    keyMap,
[...]

PR #3521, opened 2024-05-17
fix(resolver): fix execution of OpenAPI 3.1.0 dereference plugins

glowcloud:
These plugins includes:
- allOf meging
- parameter macro
- modelProperty macro

Refs #3520


test/resolver/strategies/openapi-3-1-apidom/index.js
```diff
@@ -291,6 +291,17 @@ describe('resolve', () => {
           });
         });
       });
+
+      describe('and referenced schema contains allOf keyword', () => {
+        test('should contain schema with merged properties from allOf', async () => {
```
glowcloud, line 296: Not sure if this test will work properly - it only generated the snapshot after I changed it to `test.only(...)`, so not sure if it tests it without it.

test/resolver/apidom/reference/dereference/strategies/openapi-3-1-swagger-client/schema-object/all-of.js
```diff
@@ -184,6 +184,33 @@ describe('dereference', () => {
           });
         });
 
+        test.only('should resolve allOf hidden behind a reference', async () => {
```
glowcloud, line 187: Changing this to 
```js
test('should resolve allOf hidden behind a reference', async () => {
```
should fix the tests.

char0n (approved):

Issue #3537, opened 2024-05-27
Allow passing baseURL to Swagger.exeute  method.


sauravazad:
### Content & configuration

 Swagger-Client usage:
 ```js
 SwaggerClient.execute({
  baseURL: 'someURL'
 })
 ```

### Is your feature request related to a problem?

**baseURL** is computed every time when an execute is invoked . 
https://github.com/swagger-api/swagger-js/blob/ee7fdd67a1369cdb90056ad96751ac97c9c31131/src/execute/index.js#L313

Please have a look at below flame chart . 
Building the URL consumes almost 75% of execution time  for the **executeClient** method .
<img width="2041" alt="image" src="https://github.com/swagger-api/swagger-js/assets/3856934/845a9dcb-baf5-4754-a3aa-27475d30ebf9">

### Describe the solution you'd like

If  SwaggerClient.execute() were to accept a **baseURL**  , and use it if provided instead of building it , It would help in improving the performance and execution time .

Please let me know if  this added support is acceptable . If so, I will raise a PR to add the support.

sauravazad:
@char0n

Any comments or thought on the above proposal ?

char0n:
Hi @sauravazad,

The proposal is reasonable. We're gonna go for it. I'll tend the the PR ASAP. 

Thank you!

char0n:
Addressed in https://github.com/swagger-api/swagger-js/pull/3538

char0n:
Released as https://github.com/swagger-api/swagger-js/releases/tag/v3.31.0

PR #3538, opened 2024-05-27
feat: Adding support to pass down resolved baseURL, to avoid repeated computation . Issue: #3537

sauravazad:
https://github.com/swagger-api/swagger-js/issues/3537
### Description

Add support for `baseURL` in options for performance gains .

### Motivation and Context

Issue: #3537

### How Has This Been Tested?

Added supporting  test case for the scenario. 

### Types of changes

### Checklist:

### DIFF
--- a/src/execute/index.js
+++ b/src/execute/index.js
@@ -125,10 +125,10 @@ export function buildRequest(options) {
     server,
     serverVariables,
     http,
-    signal,
+    signal
   } = options;
 
-  let { parameters, parameterBuilders } = options;
+  let { parameters, parameterBuilders, baseURL } = options;
 
   const specIsOAS3 = isOpenAPI3(spec);
   if (!parameterBuilders) {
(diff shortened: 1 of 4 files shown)


char0n:
I'll provide code review ASAP.

char0n (approved):
Thanks, LGTM! We're missing test  for `execute` as well, I'll probably add it now.

src/execute/index.js
```diff
@@ -177,7 +177,7 @@ export function buildRequest(options) {
 
   const { operation = {}, method, pathName } = operationRaw;
 
-  const baseURL = baseUrl({
+  baseURL = baseURL ?? baseUrl({
```
char0n, line 180: We still need to support Node.js >= 12.20

char0n: This is going to be correctly transpiled.

Issue #1396, opened 2018-10-16
Swagger spec is not normalized if structure is hidden behind a reference


biochimia:
I tracked down a regression in `swagger-ui` to an interaction between swagger normalization and subtree resolution introduced in https://github.com/swagger-api/swagger-js/pull/1274.

With that change, normalization happens only before subtree resolution, and this breaks in cases where the structure of the references hides the structure of the swagger file from normalization. 

I boiled it down to this this test https://github.com/swagger-api/swagger-js/compare/master...biochimia:regression/normalize-before-resolution#diff-aa37864fec325d134d6cf6fc92d7655a

### Q&A (please complete the following information)
 - OS: [e.g. macOS] macOS
 - Environment: [e.g. Chrome 59, Node.js v10.0.0] Node.js v10.12.0
 - Method of installation: [e.g. npm, unpkg] npm
 - Swagger-Client version: [e.g. 3.8.0] >3.6.0, <=3.8.21
 - Swagger/OpenAPI version: [e.g. Swagger 2.0, OpenAPI 3.0] OpenAPI 3.0

### Content & configuration

Test case: https://github.com/swagger-api/swagger-js/compare/master...biochimia:regression/normalize-before-resolution#diff-aa37864fec325d134d6cf6fc92d7655a

Swagger/OpenAPI definition:
```yaml
paths.yaml:
  root:
    parameters:
      - $ref: '#/parameters.yaml/accept-header'
    get:
      parameters:
        - $ref: '#/parameters.yaml/name'
        - $ref: '#/parameters.yaml/status'
parameters.yaml:
  accept-header:
    name: accept
    in: header
    description: content-type for response
    required: true
    type: string
  name:
    name: name
    in: formData
    description: name of the pet
    required: false
    type: string
  status:
    name: status
    in: formData
    description: status of the pet
    required: false
    type: string
paths:
  /:
    $ref: '#/paths.yaml/root'
```

Swagger-Client usage:
```js
SwaggerClient({
  // your config options here
})
```

### Describe the bug you're encountering

Because `normalizeSwagger()` doesn't get to see the structure of the path object before resolution, if doesn't copy path parameters into individual methods. Later when the subtrees are resolved normalization is already disabled and is not performed.

### To reproduce...

1. ...

char0n:
The same bug will be part of normalization that will be introduced for OpenAPI 3.1 in https://github.com/swagger-api/swagger-js/pull/2738. 

[...]

PR #3575, opened 2024-07-08
fix(resolver): don't skip normalization of OpenAPI 3.1.0 specs

glowcloud:
Refs #1396


src/subtree-resolver/index.js
```diff
@@ -53,7 +54,7 @@ const resolveSubtree = async (obj, path, options = {}) => {
     ...resolveOptions,
     spec: normalized,
     allowMetaPatches: true,
-    skipNormalization: true,
+    skipNormalization: !isOpenAPI31(normalized),
```
char0n, line 57: This shouldn't be needed. The OpenAPI 3.1.0 strategy should just ignore this setting as the strategy knows best what to do to achieve the goal.

`skipNormalization` is just going to be used for OpenAPI 3.0.x and OpenAPI 2.0, until they are replaced by new ApiDOM strategies.

char0n: IMHO we need to address overall architecture in the light of the fact, that in order for have correctly normalized documents, normalization must always be performed after the dereference/resolution.

```js
const normalized = strategy.normalize(resolveOptions);
```

This doesn't make sense anymore.

Let me think some more to figure out how architecture needs to be changed to accommodate new realities.

src/resolver/strategies/generic/resolve.js
```diff
@@ -54,7 +55,10 @@ export default async function resolveGenericStrategy(options) {
       parameterMacro,
       modelPropertyMacro,
       useCircularStructures,
-      // eslint-disable-next-line camelcase
-    }).then(skipNormalization ? async (a) => a : normalize);
+    });
+
+    result.spec = skipNormalization ? result.spec : strategy.normalize(result.spec);
```
char0n, line 60:
```suggestion
    if (!skipNormalization) {
      result.spec = strategy.normalize(result.spec)
    }
```

Not sure which one is more readable

src/resolver/strategies/openapi-3-1-apidom/normalize.js
```diff
@@ -46,18 +45,19 @@ const normalize = (element) => {
...
+
   const normalized = normalizeFn(openApiElement);
-  const value = toValue(normalized);
 
-  pojoAdapter.cache.set(spec, value);
+  /**
+   * We're setting the cache here to avoid repeated refracting
+   * in `openapi-3-1-apidom` strategy resolve method.
+   */
+  resolveOpenAPI31Strategy.cache.set(spec, normalized);
```
char0n, line 57:
```suggestion
  const value = toValue(normalized)

  resolveOpenAPI31Strategy.cache.set(value, normalized);
  
  return value;
```

src/subtree-resolver/index.js
char0n: Both of these comments no longer apply.

We need to always perform normalization on original definition before we do lazy resolution (pre-normalization) . And then after lazy resolution perform the normalization again (post-normalization).
When lazy resolution is requested again on the same definition only post-normalization is performed.
It needs to work in this way to properly handle following case -if we perform lazy resolution without pre-normalization on a certain path e.g. `[paths, /, get]` - then the post-normalization can introduce an unresolved reference. (because it can copy e.g. parameters from unresolved/undereferenced part of the definition)

src/resolver/strategies/openapi-3-1-apidom/index.js
```diff
@@ -1,14 +1,28 @@
...
   name: 'openapi-3-1-apidom',
-  match({ spec }) {
+  match(spec) {
     return isOpenAPI31(spec);
   },
-  normalize({ spec }) {
-    return pojoAdapter(normalize)(spec);
+  normalize(spec) {
+    // pre-normalization - happens only once before the first lazy dereferencing and in JavaScript context
+    if (!isElement(spec) && isPlainObject(spec) && !spec.$$normalized) {
```
glowcloud, line 15: Didn't we remove setting the `$$normalized` field so `!spec.$$normalized` is unnecessary here?

char0n (approved):

Issue #3585, opened 2024-07-10
Consider replacing `qs` with `URLSearchParams`


wojtekmaj:
### Is your feature request related to a problem?

`qs` is pulling in 14 other dependencies with it: https://npmgraph.js.org/?q=qs@6.12.3, making swagger-js installing, evaluating and running slower than it could. On top of that, since qs is being maintained by you-know-who, changes are that even more needless dependencies will pop up in the dependency tree.

### Describe the solution you'd like

Consider replacing `qs` with `URLSearchParams`.

### Describe alternatives you've considered

Custom function like:

```js
function stringify(obj) {
    function stringifyInner(obj, parentKey) {
        return Object.entries(obj).flatMap(([key, value]) => {
            const fullKey = `${parentKey}[${key}]`;
            if (typeof value === 'object') return stringifyInner(value, fullKey)
            return [[fullKey, value]]
        })
    }
    return new URLSearchParams(Object.entries(obj).flatMap(([k, v]) => {
        if (typeof v === 'object') return stringifyInner(v, k)
        return [[k, v]]
    }));
}
```

if nested value support would be needed.

char0n:
Started the effort in https://github.com/swagger-api/swagger-js/pull/3633

char0n:
@wojtekmaj addressed in https://github.com/swagger-api/swagger-js/pull/3634. I would appreciated your review.

PR #3606, opened 2024-07-28
feat: migrate to picoquery

43081j:
Migrates to using picoquery for query string parsing, a much faster and lighter library.

Note that we have to decode the stringified query in some cases since we were previously double encoding.

### Motivation and Context

As part of the ongoing [ecosystem cleanup](https://github.com/es-tooling/ecosystem-cleanup), we are migrating various projects to use lighter/faster packages.

In this case, moving from qs to picoquery - primarily because it is much faster.

### How Has This Been Tested?

Existing tests should cover it. Though if there are any gaps or places you think we should increase test coverage, please do shout up.

### Types of changes

### Checklist:


43081j:
@char0n any chance you or another maintainer can take a look at this?

the performance gain is around 6-7x the speed of qs

char0n:
Hi,

It's being considered. There is also a proposal to just use native URLSearchParams here - https://github.com/swagger-api/swagger-js/issues/3585.

char0n:
I've issued https://github.com/swagger-api/swagger-js/pull/3633 where we eliminate code that don't really need `qs` library at all.

43081j:
that works for me!

i would push for using `URLSearchParams` too but assumed you needed nesting support

note your other PR still has the dependency. we should be able to drop it still

char0n:
> i would push for using `URLSearchParams` too but assumed you needed nesting support

I'm still researching how exactly we use `qs` and what we really need from it. Ideal situation is to just use `URLSearchParams`.

> note your other PR still has the dependency. we should be able to drop it still

There are two location where `qs` is used. My [PR](https://github.com/swagger-api/swagger-js/pull/3633) addressed one of those locations, where `qs` is not needed  at all, and pure URLSearchParams just do the job. 

---

Related to `picoquery` - just by looking at the library, it is very young which means, not really battle tested and have no downloads. It's a bit risky integrating it, given it has no history. Please don't take this the wrong way, I'm just stating objective facts.

43081j:
> Related to picoquery - just by looking at the library, it is very young which means, not really battle tested and have no downloads

temporarily true while we land the various PRs across the ecosystem to introduce it (starting with storybook for example)

you are right though, it is fairly new (though it is a fork on fast-querystring which certainly is battle tested)

we have to start somewhere. this is the start of a larger migration (part of the [ecosystem-cleanup](https://github.com/43081j/ecosystem-cleanup))

you have these options really:

- do nothing (not great, qs is slow and bloated
- switch to picoquery (happening elsewhere across the ecosystem but is a fairly new library. very actively developed so even if you do find problems, they will be fixed)
- switch to URLSearchParams and stop relying on nested query strings (ideal, but depends what you're using it for currently)
- switch to fast-querystring (same as picoquery but without nesting support, only supports arrays like `foo=1&foo=2`)

char0n:
@43081j, thanks for explaining the full context in https://github.com/swagger-api/swagger-js/pull/3606#issuecomment-2282780126 and I fully support your effort. 

I've managed to construct a simple algorithm that just replicates what we originally needed from `qs` in https://github.com/swagger-api/swagger-js/pull/3634 just using URLSearchParams. Please have a look if you can.

char0n:
Superseded by https://github.com/swagger-api/swagger-js/pull/3634. Thank you again for opening this topic and all the ecosystem-cleanup effort

PR #3630, opened 2024-08-09
feat(resolver): loosen requirement for file extensions in ApiDOM parser plugins

char0n:
Refs https://github.com/swagger-api/apidom/issues/4298


glowcloud (approved):
LGTM!

PR #3634, opened 2024-08-11
refactor(http): replace qs usage by URLSearchParams

char0n:
Refs #3606
Refs #3585


43081j (approved):
looks good to me 🎉 

do you ever deal with dates in these params? might be worth having an extra `else if` specifically for dates, to `toISOString()` them (i seem to remember qs doing something like this)

char0n:
@43081j great tip. Adde the ifelse branch for Date. Thanks!

43081j (approved):

glowcloud (approved):
LGTM!

Issue #3656, opened 2024-09-09
No way to prevent encoding in serverVariables injection


paztis:
Since this PR: https://github.com/swagger-api/swagger-js/pull/3557/files all the serverVariables are automatically encoded.
We use serverVariable to manage the protocol://host:port, but :// are encoded

https://github.com/swagger-api/swagger-js/blame/b52871ebcc396c873c5616724cd191547c730e4e/src/execute/index.js#L362

openapi-server-url-templating is allowing to provide customEncoder. Is it possible to add a topLevel optional  serverVariablesEncoder to be able to cover our case ? Or to disable the encoding by default ?

the Saas and On-Premise specification provide this example of serverVariables  https://swagger.io/docs/specification/api-host-and-base-path/
![image](https://github.com/user-attachments/assets/3065a8b0-498a-443b-8da2-b570c0818dd0)

and it no more works

ponelat:
Thank you @paztis, and good catch. We don't have the resources to look into this right now, but would welcome a PR :)

ponelat:
@char0n are you able to take a look though?

char0n:
Hi @paztis,

Thanks for the issue.

If we want to support “SaaS on On-Premise” usecase, which we do - then we have to turn off the default encoding. 
We can also introduce a config option to SwaggerJS to turn the encoding on/off, with the default behavior turned off.

I'm on it.

char0n:
Addressed in https://github.com/swagger-api/swagger-js/pull/3660

PR #3660, opened 2024-09-13
fix(execute): do not encode server variables by default

char0n:
This will allow Saas and On-Premise usecase to be handled correctly.
Encoding can be turned on again by providing serverVariableEncoder option.

Refs #3656

---

**WARNING**: needs testing with SwaggerUI before merging.


glowcloud (approved):
LGTM.

PR #3677, opened 2024-10-05
Allowing user to  set "AUTH_TOKEN" in env to download privately hosted schema as well

manzarul:
In certain use cases where OpenAPI specifications reference schemas hosted on private URLs, it's important to support authentication to allow access to these resources. I identified this section of code that resolves schema references. To enable seamless access to private URLs, I propose extending the functionality to include an Authorization header if a user sets an authentication token via an environment variable. This way, when the schema reference is hosted on a private server, users can authenticate by setting the auth token in their environment, allowing the private URL to be accessed.

### Description

Allowing user to "**AUTH_TOKEN**" in env to download privately hosted schema as well.

### Motivation and Context

Currently Editor can resolve only ref which is publicly available , but if we add this then it can allow to resolved authenticated url as well.

Not sure!

### Types of changes

### Checklist:

### DIFF
--- a/src/resolver/utils/index.js
+++ b/src/resolver/utils/index.js
@@ -3,8 +3,13 @@ import { ACCEPT_HEADER_VALUE_FOR_DOCUMENTS } from '../../constants.js';
 // eslint-disable-next-line import/prefer-default-export
 export function makeFetchJSON(http, opts = {}) {
   const { requestInterceptor, responseInterceptor } = opts;
+
   // Set credentials with 'http.withCredentials' value
   const credentials = http.withCredentials ? 'include' : 'same-origin';
+
+  // Get Authorization token from environment variables if set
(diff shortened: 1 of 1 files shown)


char0n:
Hi @manzarul,

`swagger-client` is an isomorphic library - it must work in browser and Node.js like environment. Reading environment variables is platform specific thing. 

I would say, that what you're trying to achieve if fully possible by using `requestInterceptor`. Please see the following docs section, and let me know if it resolves your requirement:

- https://github.com/swagger-api/swagger-js/blob/master/docs/usage/http-client.md#loading-specification
- https://github.com/swagger-api/swagger-js/blob/master/docs/usage/http-client.md#request-interceptor

char0n:
I'm going to close this. @manzarul please reopen if `requestInceptor` won't handle your usecase.

PR #3726, opened 2024-12-02
fix(execute): fix encoding of object properties with null value

glowcloud:
Refs https://github.com/swagger-api/swagger-ui/issues/10226

Fixes the issue with `null` and `undefined` values causing errors due to trying to check their length.

### DIFF
--- a/test/oas3/execute/main.js
+++ b/test/oas3/execute/main.js
@@ -664,6 +664,78 @@ describe('buildRequest - OpenAPI Specification 3.0', () => {
       });
     });
 
+    it('should encode objects with a property with `null` value', () => {
+      const req = buildRequest({
+        spec: {
+          openapi: '3.0.0',
+          paths: {
+            '/{pathPartial}': {
+              post: {
+                operationId: 'myOp',
+                parameters: [
+                  {
+                    name: 'query',
+                    in: 'query',
+                    schema: {
+                      type: 'object',
+                    },
+                  },
+                  {
+                    name: 'FooHeader',
+                    in: 'header',
+                    schema: {
+                      type: 'object',
+                    },
+                    explode: true,
+                  },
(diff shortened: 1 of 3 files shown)


test/oas3/execute/main.js
```diff
@@ -664,6 +664,142 @@ describe('buildRequest - OpenAPI Specification 3.0', () => {
...
+          },
+          myCookie: {
+            d: null,
+          },
+        },
+      });
+
+      expect(req).toEqual({
+        method: 'POST',
+        url: `/a=null?b=null`,
```
char0n, line 730: This look weird. Why would anybody want that? Le't think about this, if it make sense

test/oas3/execute/main.js
```diff
@@ -664,6 +664,142 @@ describe('buildRequest - OpenAPI Specification 3.0', () => {
...
+                  },
+                ],
+              },
+            },
+          },
+        },
+        operationId: 'myOp',
+        parameters: {
+          pathPartial: {
+            a: null,
```
char0n, line 714:
```suggestion
            a: {a: null},
            
            a: '{"a": null}'
```

test/oas3/execute/main.js
```diff
@@ -664,6 +664,142 @@ describe('buildRequest - OpenAPI Specification 3.0', () => {
...
+              },
+            },
+          },
+        },
+        operationId: 'myOp',
+        parameters: {
+          pathPartial: [undefined],
+          query: [undefined],
+          FooHeader: [undefined],
+          myCookie: [undefined],
```
char0n, line 788: Please test for this as well `[undefined, undefined, undefined]`

char0n (approved):
LGTM + please do the change for `stringyfiQuery`

PR #3762, opened 2025-01-06
fix(execute): serialize req.cookies into valid cookie-string

char0n:
This is part of aligning with OpenAPI 3.0.4 clarifications.

Specifically this PR is aligning implementations around cookies with https://spec.openapis.org/oas/v3.0.4.html#appendix-d-serializing-headers-and-cookies.

`swagger-client` will now product valid cookies, which can be processed by [RFC 6265](https://tools.ietf.org/html/rfc6265) parser, `cookie-value` is percent decoded and can be further processed by [RFC 6570](https://www.rfc-editor.org/rfc/rfc6570) implementation.

### DIFF
--- a/src/execute/oas3/parameter-builders.js
+++ b/src/execute/oas3/parameter-builders.js
@@ -1,7 +1,9 @@
 import { resolve as resolvePathTemplate } from 'openapi-path-templating';
+import { serializeCookie } from '@swaggerexpert/cookie';
 
 import stylize, { encodeCharacters } from './style-serializer.js';
 import serialize from './content-serializer.js';
+import cookieValueEncoder from '../../helpers/cookie-value-encoder.js';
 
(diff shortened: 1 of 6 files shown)


src/execute/oas3/parameter-builders.js
```diff
@@ -106,28 +108,41 @@ export function header({ req, parameter, value }) {
...
+    const serializedValue = stylize({
+      key: parameter.name,
+      value,
+      escape: false,
+      style: parameter.style || 'form',
+      explode: parameter.explode ?? false,
+    });
+    const cookieValue =
+      Array.isArray(value) && parameter.explode
+        ? `${parameter.name}=${serializedValue}`
```
char0n, line 138:
```suggestion
        ? `${cookieName}=${serializedValue}`
```

glowcloud (approved):
LGTM.

PR #3782, opened 2025-01-24
fix(execute): explode object parameters for combined schemas

robert-hebel-sb:
Refs https://github.com/swagger-api/swagger-ui/issues/10002


src/execute/index.js
```diff
@@ -25,6 +25,14 @@ import { isOpenAPI3 } from '../helpers/openapi-predicates.js';
 
 const arrayOrEmpty = (ar) => (Array.isArray(ar) ? ar : []);
 
+const isObjectSchema = (schema) => schema?.type === 'object';
+
+const isAnySchemaObjectType = (schema) =>
```
char0n, line 30: ### allOf

We don't actually need to handle all of as the OpenAPI description is normalized before it hits this code. For OpenAPI 3.0.x in [here](https://github.com/swagger-api/swagger-js/blob/6e1c10e396fbcb40185a96485ec82e925e5bf1fa/src/resolver/specmap/lib/all-of.js) and for OpenAPI 3.1.x [here](https://github.com/swagger-api/swagger-js/blob/56df8f613b6e73ded8f72e79c336f817dcf694c5/src/resolver/apidom/reference/dereference/strategies/openapi-3-1-swagger-client/visitors/all-of.js#L52). Having said that, the resolved/derefenced OpenAPI Description doesn't have any `allOf` keyword anymore.

### anyOf & oneOf

Until now we haven't really supported `anyOf` or `oneOf` keywords in SwaggerJS. This is the first time we're handling them. For this particular case I'm suggesting the following:

We'll traverse recursively first `oneOf` and then `anyOf` (in that order). The order is defined like that because with `oneOf` keyword, exactly one must be valid, apart to `anyOf` which is the most permissive and uncertain.  We will look for the first `Schema Object` with `type="object"`. The traversal can be done with `neotraverse/legacy` library which we already use inside SwaggerJS. We only perform the traversal if the variable `value` can be parsed by `JSON.parse`, otherwise we assume the lookup for JSONSchema(type="object") is unnecessary.

### type="object"

As we now support OpenAPI 3.0 and 3.1, the `type` keyword can be `string` or an array of `string` (type union) represting types.

OpenAPI 3.0: { type: "object" }
OpenAPI 3.1: { type: ["object"] } 

The code needs to compensate for that and compare the stings or the do the lookup within the array.

src/execute/index.js
```diff
@@ -25,6 +26,26 @@ import { isOpenAPI3 } from '../helpers/openapi-predicates.js';
 
 const arrayOrEmpty = (ar) => (Array.isArray(ar) ? ar : []);
 
+const hasCombinedSchema = (schema) => !!schema?.oneOf || !!schema?.anyOf;
```
char0n, line 29: @robert-hebel-sb it looks a bit complex and we have a couple of assertion issues:

- we don't checked if `oneOf`, `anyOf` are arrays
- we use `includes('object')` which will return true for "testobjecttest" as well
- we can refine the traversal to only take account JSON Schemas that are directly (in not nested way) attached to `anyOf`, `oneOf`

What would you say to something like this?

```js
const findObjectSchema = (schema) => {
    if (!isPlainObject(schema)) return undefined;
    
    // check if the schema is an object type
    if (schema.type === "object" || (Array.isArray(schema.type) && schema.type.includes("object"))) {
        return schema;
    }
    
    // traverse oneOf keyword first
    if (Array.isArray(schema.oneOf)) {
        for (const subschema of schema.oneOf) {
            const result = findObjectSchema(subschema);
            if (result) return result;
        }
    }
    
    // traverse anyOf keyword second
    if (Array.isArray(schema.anyOf)) {
        for (const subschema of schema.anyOf) {
            const result = findObjectSchema(subschema);
            if (result) return result;
        }
    }
    
    return undefined;
}

if (specIsOAS3 && findObjectSchema(parameter.schema) && typeof value === 'string') {
  try {
    value = JSON.parse(value);
  } catch (e) {
    throw new Error('Could not parse object parameter value string as JSON');
  }
}

Comment (char0n) [reply src/execute/index.js:29]: @glowcloud will take over from here.

Comment (char0n) [reply src/execute/index.js:29]: Please also make sure that traversal cannot be DDOSed (by stopping after the same number of recursions as we do during specpam dereferencing)

Comment (robert-hebel-sb) [reply src/execute/index.js:29]: @char0n @glowcloud 
1. This solution performs traversal even in cases when the value can't be parsed by `JSON.parse` which was described in the previous comment as a requirement. Let me know which approach should be applied here. 
2. Do you have an example of traversal protection from DDOS that can be reused here?

Comment (glowcloud) [reply src/execute/index.js:29]: > @char0n @glowcloud
> 
> 1. This solution performs traversal even in cases when the value can't be parsed by `JSON.parse` which was described in the previous comment as a requirement. Let me know which approach should be applied here.
> 2. Do you have an example of traversal protection from DDOS that can be reused here?

1. Yes, we still want to check if it can be parsed first
2. I think what @char0n meant is the [traverse limit](https://github.com/swagger-api/swagger-js/blob/7a471c3d84177e5410074093f679fddc9a298f5f/src/resolver/specmap/index.js#L9)

Comment (char0n) [reply src/execute/index.js:28]: Hi all,

> This solution performs traversal even in cases when the value can't be parsed by JSON.parse which was described in the previous comment as a requirement. Let me know which approach should be applied here.

Yes, but `findObjectSchema` has small distinction - it doesn't traverse nested arbitrary subschemas, it only traverse root schema and all direct subschemas of `anyOf` and `oneOf` applicators and only returns match if there is an Object JSON Schema there. Therefore the logic of throwing an error applies here as well now.

> Yes, we still want to check if it can be parsed first

Not really anymore. See the explanation above.

> Do you have an example of traversal protection from DDOS that can be reused here?

We can track the depth of traversal via function argument:

```js
const findObjectSchema = (schema, { depth = 1 } = {}) => {
  if (!isPlainObject(schema)) return undefined;

  // check if the schema is an object type
  if (schema.type === 'object' || (Array.isArray(schema.type) && schema.type.includes('object'))) {
    return schema;
  }

  if (depth > TRAVERSE_LIMIT) return undefined;

  if (Array.isArray(schema.oneOf)) {
    // traverse oneOf keyword first
    for (const subschema of schema.oneOf) {
      const result = findObjectSchema(subschema, { depth: depth + 1 });
      if (result) return result;
    }
  }

  // traverse anyOf keyword second
  if (Array.isArray(schema.anyOf)) {
    for (const subschema of schema.anyOf) {
      const result = findObjectSchema(subschema, { depth: depth + 1 });
      if (result) return result;
    }
  }

  return undefined;
};
```

---

We can further extend the `findObjectSchema` for `recurse` option to allow for more flexibility:


```js
const findObjectSchema = (schema, { recurse = true, depth = 1 } = {}) => {
  if (!isPlainObject(schema)) return undefined;

  // check if the schema is an object type
  if (schema.type === 'object' || (Array.isArray(schema.type) && schema.type.includes('object'))) {
    return schema;
  }

  if (depth > TRAVERSE_LIMIT) return undefined;

  if (recurse && Array.isArray(schema.oneOf)) {
    // traverse oneOf keyword first
    for (const subschema of schema.oneOf) {
      const result = findObjectSchema(subschema, { recurse, depth: depth + 1 });
      if (result) return result;
    }
  }

  // traverse anyOf keyword second
  if (recurse && Array.isArray(schema.anyOf)) {
    for (const subschema of schema.anyOf) {
      const result = findObjectSchema(subschema, { recurse, depth: depth + 1 });
      if (result) return result;
    }
  }

  return undefined;
};
```

That way we can simplify our implementation to:

```js
    if (
      specIsOAS3 &&
      typeof value === 'string' &&
      (findObjectSchema(parameter.schema, { recurse: false }) ||
        (!('type' in parameter.schema) && findObjectSchema(parameters.schema)))
    ) {
      try {
        value = JSON.parse(value);
      } catch (e) {
        throw new Error('Could not parse object parameter value string as JSON');
      }
    }
```

Explanation: 

` (findObjectSchema(parameter.schema, { recurse: false }) ||` - just tells us if root schema is of type `object`. If yes we dive into success block of the if statement.

` (!('type' in parameter.schema) && findObjectSchema(parameters.schema)))` - now we need to make sure root schema doesn't have a `type` keyword. If it would have it, we would get into logical dilemma with `anyOf` and `oneOf` applicators. Here is an example where the schema instance can neither be string nor object (it always validates to false). Our boolean logic avoids this situation.

```json
{ 
  "type": "string",
  "oneOf": [
    { "type": "object" }
  ]
}
```

Pleas let me know your thoughts.

char0n: After further discussing with @glowcloud, she uncovered flaw in my logic. The logic conditionality includes assertion on parsed value expectation.

```
    if (
      specIsOAS3 &&
      typeof value === 'string' &&
      findObjectSchema(parameter.schema, { recurse: false })    
    ) {
      try {
        value = JSON.parse(value);
        if (typeof value !== 'object') throw new Error('Expected JSON serialized object');
      } catch (e) {
        throw new Error('Could not parse object parameter value string as JSON');
      }
    } else if (
      specIsOAS3 &&
      typeof value === 'string' &&
      !('type' in parameter.schema) && 
      findObjectSchema(parameters.schema)
    ) {
       try {
         const parsedValue = JSON.parse(value);
         if (typeof parsedValue === 'object') value = parsedValue
       catch { /* noop */ }
   }
```

char0n: Feel free to simplify. This is a mere pseudo-code that was created to capture the business logic.

char0n: And thank you for having patience for multiple interactions on this. Spec related things are hard to get right, and usually require many iterations and probing.

src/execute/index.js
```diff
@@ -261,16 +300,12 @@ export function buildRequest(options) {
...
-      specIsOAS3 &&
-      parameter.schema &&
-      parameter.schema.type === 'object' &&
-      typeof value === 'string'
-    ) {
-      try {
-        value = JSON.parse(value);
-      } catch (e) {
-        throw new Error('Could not parse object parameter value string as JSON');
+    if (specIsOAS3 && typeof value === 'string') {
```
char0n, line 303: @robert-hebel-sb the traversal looks cool, but the further business logic seems to be not-aligned with the pseudocode provided before. Your new code with throw when object schema was found within the `anyOf` or `oneOf` and the value is not a parse-able JSON Object. The code shouldn't throw in that case.

**Pseudocode**:
```js
 if (
      specIsOAS3 &&
      typeof value === 'string' &&
      findObjectSchema(parameter.schema, { recurse: false })    
    ) {
      try {
        value = JSON.parse(value);
        if (typeof value !== 'object') throw new Error('Expected JSON serialized object');
      } catch (e) {
        throw new Error('Could not parse object parameter value string as JSON Object');
      }
    } else if (
      specIsOAS3 &&
      typeof value === 'string' &&
      !('type' in parameter.schema) && 
      findObjectSchema(parameters.schema, { recurse: true })
    ) {
       try {
         const parsedValue = JSON.parse(value);
         if (typeof parsedValue === 'object') value = parsedValue
       } catch { /* noop */ }
   }
```

src/execute/index.js
```diff
@@ -25,6 +25,52 @@ import { isOpenAPI3 } from '../helpers/openapi-predicates.js';
...
+      ? schema.anyOf.find((subschema) => findObjectSchema(subschema, { recurse, depth: depth + 1 }))
+      : undefined;
+
+    if (anyOfResult) return anyOfResult;
+  }
+
+  return undefined;
+};
+
+const parseJsonValue = ({ value, shouldFailSafe = false }) => {
```
char0n, line 57: NITPICK: Can I suggest naming this `parseJsonObject` ? That's what this function basically does. And maybe consider naming `shouldFailSafe` as `shouldFailSilently` or just `silentFail` ?

```suggestion
const parseJsonObject = ({ value, silentFail = false }) => {
```

src/execute/index.js
```diff
@@ -25,6 +25,52 @@ import { isOpenAPI3 } from '../helpers/openapi-predicates.js';
...
+  try {
+    const parsedValue = JSON.parse(value);
+    if (typeof parsedValue === 'object') {
+      return parsedValue;
+    }
+    if (!shouldFailSafe) {
+      throw new Error('Expected JSON serialized object');
+    }
+  } catch (e) {
```
char0n, line 66:
```suggestion
  } catch {
```

src/execute/index.js
```diff
@@ -261,16 +307,14 @@ export function buildRequest(options) {
...
-    ) {
-      try {
-        value = JSON.parse(value);
-      } catch (e) {
-        throw new Error('Could not parse object parameter value string as JSON');
+    if (specIsOAS3 && typeof value === 'string') {
+      if (findObjectSchema(parameter.schema, { recurse: false })) {
+        value = parseJsonValue({ value, shouldFailSafe: false });
+      } else if (
+        isNil(parameter.schema?.type) &&
```
char0n, line 314: NITPICK: can we use native JS constructs if they can be expressed in simple way? In this particular case:

```suggestion
        parameter.schema?.type == null &&
```

We can further refine logic here and skip the execution of recursive search when type is defined with whatever value. User can be define it as `type: null` or `type: undefined` and we SHOULD skip the recursive traversal in this case as well. We cannot use `Object.hasOwn()` as this library still supports Node.js >= 12.20.0. But we can use `R.has` from ramda.

char0n (approved):
Expected business logic is now in place. Thank you!

PR #3826, opened 2025-03-10
fix(execute): explode array parameters for combined schemas

glowcloud:
This PR fixes encoding of parameters with:
- union type which includes `array` (`type: [array, ...]`)
- complex schemas with `anyOf/oneOf` and nested `type: array`

---

When an OpenAPI 3.1.0 parameter contains a complex union type, the formatting of requests is not applied properly.


src/execute/index.js
```diff
@@ -303,10 +312,14 @@ export function buildRequest(options) {
     }
 
     if (specIsOAS3 && typeof value === 'string') {
-      if (has('type', parameter.schema) && findObjectSchema(parameter.schema, { recurse: false })) {
-        value = parseJsonObject({ value, silentFail: false });
-      } else if (findObjectSchema(parameter.schema, { recurse: true })) {
-        value = parseJsonObject({ value, silentFail: true });
+      if (
+        has('type', parameter.schema) &&
+        !Array.isArray(parameter.schema.type) &&
```
glowcloud, line 317: This check should fix a case where we have e.g.
```yaml
type: [object, string]
```
and we try to send a simple value, e.g. `"test"` - previously we would throw an error.

char0n, line 317:
```suggestion
        typeof parameter.schema.type === 'string' &&
```

src/execute/index.js
```diff
@@ -303,10 +312,14 @@ export function buildRequest(options) {
...
+        findObjectOrArraySchema(parameter.schema, { recurse: false })
+      ) {
+        value = parseJsonObjectOrArray({ value, silentFail: false });
+      } else if (findObjectOrArraySchema(parameter.schema, { recurse: true })) {
```
char0n, line 321:
```suggestion
      } else if (
      has('type', parameter.schema) &&
      Array.isArray(parameter.schema.type) &&
      findObjectOrArraySchema(parameter.schema, { recurse: false })) {
         
         value = parseJsonObjectOrArray({ value, silentFail: true });
```

src/execute/index.js
```diff
@@ -303,10 +312,14 @@ export function buildRequest(options) {
...
+        value = parseJsonObjectOrArray({ value, silentFail: true });
```
char0n, line 322: 3rd case is: checking for not having a type:

```js
} else if (
!has('type', parameter.schema) &&
findObjectOrArraySchema(parameter.schema, { recurse: true })) {
  value = parseJsonObjectOrArray({ value, silentFail: true });
}
```

src/execute/index.js
```diff
@@ -49,18 +58,18 @@ const findObjectSchema = (schema, { recurse = true, depth = 1 } = {}) => {
   return undefined;
 };
 
-const parseJsonObject = ({ value, silentFail = false }) => {
+const parseJsonObjectOrArray = ({ value, silentFail = false }) => {
   try {
     const parsedValue = JSON.parse(value);
     if (typeof parsedValue === 'object') {
```
char0n, line 64: isPlainObject(parsedValue) || Array.isArray(parsedValue)

char0n: handles the `null` parsing, where `typeof null === 'object'` which is wrong (attribution @glowcloud)

char0n (approved):
LGTM now

PR #3857, opened 2025-03-27
fix(specmap): fix resolver being affected by the order of schemas

glowcloud:
Refs https://github.com/swagger-api/swagger-ui/issues/10096


char0n (approved):
The change has been discussed in depth. The only risk is that we'll introduce a possible inefficiency in the specmap traversal (which is acceptable).

PR #3927, opened 2025-05-19
fix(resolver): avoid accessing properties of undefined or null parameters

glowcloud:
This PR fixes acessing properties of empty parameters during normalization, e.g.:

```yaml
openapi: 3.0.4
paths:
  /test:
    parameters:
      - 
    get:
      parameters:
        - 
```

### DIFF
--- a/src/resolver/strategies/generic/normalize.js
+++ b/src/resolver/strategies/generic/normalize.js
@@ -85,10 +85,12 @@ export default function normalize(parsedSpec) {
                 for (const param of inherits[inheritName]) {
                   const exists = operation[inheritName].some(
                     (opParam) =>
-                      (opParam.name && opParam.name === param.name) ||
-                      (opParam.$ref && opParam.$ref === param.$ref) ||
-                      (opParam.$$ref && opParam.$$ref === param.$$ref) ||
-                      opParam === param
+                      opParam &&
+                      param &&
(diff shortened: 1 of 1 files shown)


src/resolver/strategies/generic/normalize.js
```diff
@@ -85,10 +85,12 @@ export default function normalize(parsedSpec) {
                 for (const param of inherits[inheritName]) {
                   const exists = operation[inheritName].some(
                     (opParam) =>
```
char0n, line 87: The business logic checks out. Maybe a more safer alternative of the code?

```js
              if (!Array.isArray(operation[inheritName])) {
                operation[inheritName] = inherits[inheritName];
              } else if (inheritName === 'parameters') {
                // eslint-disable-next-line no-restricted-syntax
                for (const param of inherits[inheritName]) {
                  const exists = operation[inheritName].some((opParam) => {
                    if (!isPlainObject(opParam) && !isPlainObject(param)) return false;
                    if (opParam === param) return true;

                    return ['name', '$ref', '$$ref'].some(
                      (key) =>
                        typeof opParam[key] === 'string' &&
                        typeof param[key] === 'string' &&
                        opParam[key] === param[key]
                    );
                  });

                  if (!exists) {
                    operation[inheritName].push(param);
                  }
                }
              }
```

char0n (approved):
Approved with suggestion

PR #3939, opened 2025-05-30
fix(resolver): avoid duplicating enum values when nested in allOf

glowcloud:


src/resolver/apidom/reference/dereference/strategies/openapi-3-1-swagger-client/visitors/all-of.js
```diff
@@ -37,7 +43,32 @@ class AllOfVisitor {
...
         const { allOf } = schemaElement;
         schemaElement.remove('allOf');
-        const allOfMerged = deepmerge.all([...allOf.content, schemaElement]);
+        const allOfMerged = deepmerge.all([...allOf.content, schemaElement], {
+          customMerge: (keyElement) => {
+            if (toValue(keyElement) === 'enum') {
+              return (targetElement, sourceElement) => {
+                if (isArrayElement(targetElement) && isArrayElement(sourceElement)) {
+                  const primitiveElements = new ArrayElement([
+                    ...targetElement.findElements(isPrimitiveElement),
```
char0n, line 52: `isPrimitiveElement` is more about distinction between generic APiDOM elements and Semantic ones. `info`  (from OpenAPI spec) vs `object`.  So this doesn't really do what is was intended to do. 

Let's try to put all element inside one array, filter out nonsense (object/array), and use R.uniqBy with `toValue` as first argument.

src/resolver/apidom/reference/dereference/strategies/openapi-3-1-swagger-client/visitors/all-of.js
```diff
@@ -37,7 +38,35 @@ class AllOfVisitor {
       while (schemaElement.hasKey('allOf')) {
         const { allOf } = schemaElement;
         schemaElement.remove('allOf');
-        const allOfMerged = deepmerge.all([...allOf.content, schemaElement]);
+        const allOfMerged = deepmerge.all([...allOf.content, schemaElement], {
+          customMerge: (keyElement) => {
+            if (toValue(keyElement) === 'enum') {
+              return (targetElement, sourceElement) => {
+                if (isArrayElement(targetElement) && isArrayElement(sourceElement)) {
```
char0n, line 45:
```suggestion
                if (includesClasses(['json-schema-enum'], targetElement) && includesClasses(['json-schema-enum'], sourceElement)) {
```

char0n (approved):
One comment.

PR #4281, opened 2026-09-18
fix(resolver): generate allOf $ref patches before merging members

### LINKED ISSUE #11018: [Bug]: Resolver error for array item refs in nested external OpenAPI documents
## Describe the bug you're encountering

Swagger UI reports a resolver error when an operation is reached through an external Path Item and its request schema is in another external YAML file. The schema contains `allOf`/`oneOf` branches and an array whose `items` is a local `$ref` to an enum in that schema file.

All referenced YAML files are served successfully. Replacing the two `items.$ref` nodes with equivalent inline string enums removes the error, so this appears to be a resolver issue rather than an unavailable-file or invalid-schema issue.

This looks related to #5820, but this reproduction specifically requires the external Path Item -> external Schema reference chain.

## Environment

- Swagger UI: `5.32.14` (latest at the time of reporting)
- OpenAPI: `3.0.3`
- Browser: Chromium
- Installation: CDN assets from `swagger-ui-dist@5.32.14`

## To reproduce

Serve the following three files from one directory, for example:

```sh
python3 -m http.server 8080
```

Then open `http://localhost:8080/index.html`.

`index.html`

```html
<!doctype html>
<html lang="en">
<head>
  <meta charset="utf-8">
  <title>Swagger UI resolver reproduction</title>
  <link rel="stylesheet" href="https://unpkg.com/swagger-ui-dist@5.32.14/swagger-ui.css">
</head>
<body>
<div id="swagger-ui"></div>
<script src="https://unpkg.com/swagger-ui-dist@5.32.14/swagger-ui-bundle.js"></script>
<script src="https://unpkg.com/swagger-ui-dist@5.32.14/swagger-ui-standalone-preset.js"></script>
<script>
  window.ui = SwaggerUIBundle({
    url: 'openapi.yml',
    dom_id: '#swagger-ui',
    presets: [SwaggerUIBundle.presets.apis, SwaggerUIStandalonePreset],
    layout: 'StandaloneLayout'
  })
</script>
</body>
</html>
```

`openapi.yml`

```yaml
openapi: 3.0.3
info:
  title: Chart configuration API
  version: 1.0.0
paths:
  /charts:
    $ref: './paths/charts.yml#/charts'
```

`paths/charts.yml`

```yaml
charts:
  post:
    operationId: createChart
    requestBody:
      required: true
      content:
        application/json:
          schema:
            $ref: '../schemas/chart.yml#/ChartConfiguration'
    responses:
      '204':
        description: Chart created
```

`schemas/chart.yml`

```yaml
ChartConfiguration:
  type: object
  additionalProperties: false
  required: [view]
  properties:
    view:
      type: string
    chartTypes:
      type: array
      items:
        $ref: '#/ChartType'
  allOf:
    - oneOf:
        - properties:
            view:
              type: string
              enum: [summary]
        - properties:
            view:
              type: string
              enum: [detail]
    - oneOf:
        - additionalProperties: true
          required: [chartTypes]
          properties:
            view:
              type: string
              enum: [summary]
            chartTypes:
              type: array
              minItems: 1
              items:
                $ref: '#/ChartType'
          allOf:
            - oneOf:
                - additionalProperties: true
                  properties:
                    includeLegend:
                      type: boolean
                      enum: [true]
                - additionalProperties: true
                  properties:
                    includeLegend:
                      type: boolean
                      enum: [false]
        - additionalProperties: true
          properties:
            view:
              type: string
              enum: [detail]

ChartType:
  type: string
  enum: [line, bar, pie]
```

## Actual behavior

Swagger UI renders the operation but displays:

```text
Resolver error
Cannot read properties of undefined (reading 'items')
```

The browser console points to `applyOperation` / `applyPatch` in `swagger-ui-bundle.js`.

## Expected behavior

The local `ChartType` references in `items` should resolve without an error. The schema is valid OpenAPI 3.0.3, and every YAML file is accessible.

## Workaround

Replacing both occurrences below with the equivalent inline schema removes all resolver errors:

```yaml
items:
  type: string
  enum: [line, bar, pie]
```

htgylzhq:
### Description

The `allOf` plugin absolutifies `$ref`/`$$ref` values via `generateAbsoluteRefPatches`. Those patches were generated *after* the `mergeDeep` patch for each member, using paths relative to that member, with `allOf` dropped from the path.

`deepmerge` concatenates arrays, so a member's array keyword does not stay where it was: once an earlier member has already contributed to `oneOf`, the second member's `oneOf/0` lives at `oneOf/2` in the merged parent. The patch then targets a branch that has no such node, and applying it throws:

```
Resolver error
Cannot read properties of undefined (reading 'items')
```

This PR generates those patches up front instead, while every member is still in place at `allOf/<i>`, so their paths cannot go stale. Nothing about the merge itself changes.

Two notes for reviewers:

- **The patches cannot be replaced by resolving the values inline before merging.** I tried that first: every `allOf` test still passed, but `test/resolver/specmap/complex.js` then ran without finishing (~12s before, killed at 180s). Applying these patches also re-runs the refs plugin over each `$ref`, which is what keeps the cycle detection in `refs.js` fed; without it, recursive schemas expand without bound. There is a comment in the code about this.
- **`String(i)` is deliberate.** A numeric path token throws `token.replace is not a function` in `escapeJsonPointerToken()`, since specmap paths are strings everywhere else.

`helpers.js` is unchanged: the default `getBaseUrlForNodePath` already resolves the member-relative base URL correctly.

### Motivation and Context

Fixes swagger-api/swagger-ui#11018

The reproduction in that issue uses an external Path Item pointing at an external schema, but the external files are not required — a single-document spec with local `$ref`s inside `allOf` members reproduces it just as well. The added unit tests cover both.

### How Has This Been Tested?

- `npm run test:unit`: 91 suites / 1072 tests pass in ~27s (1070 before, +2 new tests). `npm run lint` clean.
- Two regression tests added to `test/resolver/specmap/all-of.js`: a minimal local-`$ref` case, and one mirroring the issue's external Path Item to external schema chain, mocked with `undici.MockAgent`.
- Confirmed both new tests fail on `main` with exactly the error from the issue.
- End to end against Swagger UI: built `es/`, dropped it into a Swagger UI checkout, and ran a new Cypress spec with the issue's three documents. It fails on swagger-client 3.38.2 and passes with this change; 14 other specs covering `allOf`, circular refs and external file refs stay green.

n/a

### Types of changes

### Checklist:

🤖 Generated with [Claude Code](https://claude.com/claude-code)


htgylzhq:
One note on the contribution guide: it asks for a `bug/` branch prefix, and this branch is `fix/all-of-ref-patches-array-merge`. I went with `fix/` to match what the repository actually uses — recent branches here are `fix/`, `chore/` and `ci/`. Happy to re-open from a `bug/` branch if you'd rather have the guide followed to the letter.

The rest of the guide is covered: single atomic commit, Conventional Commits, 65-character imperative subject, no closing keywords in the commit message (only in this PR's body), and tests for the change.

